PR #1, opened 2013-12-10
Add on_failure handler (API)

sethvargo:
[Pretty render](https://github.com/opscode/chef-rfc/blob/sethvargo/on_failure/on_failure.md)


athingisathing:
When will this be in prod?

cnmcavoy:
Whatever happened to this RFC? It looks like it was accepted, but never moved past the proof-of-concept stage.

PR #6, opened 2014-02-18
WinRM listener configuration RFC

muktaa:


thommay:
@adamedx @muktaa what's the status of this one?

btm:
So this is mostly implemented with the exception of ```knife windows listener [list|delete]``` and should go out in knife-windows 1.0.0 soon.

https://github.com/chef/knife-windows/blob/master/DOC_CHANGES.md#new-subcommands-to-automate-winrm-ssl-listener-configuration

I wonder if this slips out of the scope of something we need an RFC for. Do we need RFC's for all features? I don't think so, more for designing features that cross different parts of the project or multiple projects, so we have a standard as to how things should act.

thommay:
@btm agree entirely. I've marked a couple more PRs as invalid for similar reasons.

btm:
we probably should add some words to the top of rfc000 about that. I guess process-wise we technically need @adamhjk to close / "reject" RFCs, but I would think that should be something editors could do in cases of maintenance. /cc editors: @jonlives @coderanger @adamleff

muktaa:
^ @nimishas - close the pr if no one is working on this.

nathenharvey:
This PR is currently on the [agenda for our next IRC developers' meeting](https://github.com/chef/chef-community-irc-meetings/blob/master/2015-03-19-agenda.md).

Please let me know if it gets merged or otherwise closed before then so that the agenda can be updated.

Thanks!

adamedx:
Closing since this pre-dated the RFC process and also concerns a separate open source project, not Chef.

Thanks @nathenharvey.

PR #31, opened 2014-07-25
File Specificity Overhaul

coderanger:


coderanger:
For anyone that finds this via Google, this feature is not usable in Chef Server pending https://github.com/chef/chef-server/pull/318

PR #44, opened 2014-08-15
jjasghar/ubuntu init provider

jjasghar:
This is a RFC to move the way init providers are found and used for ubuntu forward. This can eventually help every major distro if we take the suggestion.


thommay:
I think this is superseded by history; closing. If anyone wants to revisit this, please feel free to reopen.

PR #71, opened 2014-11-21
Dialects in Chef

coderanger:
Have at it. There was a lot of discussion about this last time, and the agreement was that while the framework would be added to core, only Ruby and JSON dialects would be included to match existing functionality. All new dialects will be done in external gems/cookbooks until such time as they can prove their worth in battle, as it were.

I think adapting the old patches would be a good step, which covers recipes, attributes, and non-chef_fs loading of roles/data bags/environments. The chef_fs support can build on that, along with template dialects to support things other than Erb.

Ping @sersut @danielsdeleo @sdelano @lamont-granquist

### DIFF
--- a/new/dialects.md
+++ b/new/dialects.md
@@ -1,0 +1,120 @@
+---
+RFC: unassigned
+Author: Noah Kantrowitz <noah@coderanger.net>
+Status: Draft
+Type: Standards Track
+---
+
+# Dialects in Chef
+
+Currently Chef supports a mix of Ruby DSLs and JSON for it's data. This RFC
+proposes to add hooks in to Chef to allow cookbooks and knife plugins to support
+additional formats.
+
+## Motivation
+
+The overall motivation can be summed up as:
+
+    As a Chef user,
+    I want to write in a variety of formats,
+    so that cookbook maintenance is easier.
+
+A specific version of that which deserves extra mention is:
+
+    As a new Chef user,
+    I want to write in a variety of formats,
+    so that I don't have to learn Ruby.
+
+## Specification
+
+At heart, the dialects system is an additional layer of indirection around
+all forms of file loading in Chef. Currently most code either calls
+`FFI_Yajl::Parser.parse` or `ClassName.from_file`, sometimes with a simple
+dispatcher like in `Chef::Knife::Core::ObjectLoader`. The dialects registry
+provides a central location to register code to handle loading a given file
+suffix or MIME type and then utility functions to return an object of a given
+type and filename.
+
+### Dialect Types
+
+Dialects are used in two distinct places: cookbooks and knife plugins.
+Dialect cookbooks are used to provide dialects for things like attribute
+and recipe files, while dialect knife plugins provide support for files like
+roles and data bags.
+
+While both types of dialects use the same implementation, the ways they are used
+require different packaging and support code (cookbook vs gem).
+
+### Dialect Classes
+
+The heart of any dialect implementation is a dialect class. This inherits from
+`Chef::Dialect` and declares which file extensions and MIME types this dialect
+will handle. For each object type the dialect will support, it defines a
+method taking a blank object and a filesystem path which will process any
+needed data and insert it in to the provided object.
+
+### Dialect Selection
+
+`Chef::Dialect` provides two class methods to load a dialect from either a
+file extension or MIME type. These both return an instance of the dialect
+class.
+
+### Dialect Plugins
+
+A dialect plugin for knife is simply a gem that defines one or more dialect
+classes and ensures they are loaded by knife's plugin framework. The easiest
+way to accomplish this is to place the dialect class in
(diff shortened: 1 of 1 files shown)


adamhjk:
Ok - @jaymzh @lamont-granquist @mcquin @tyler-ball @danielsdeleo @jonlives - unless one of you thinks we shouldn't have this extensible and puts up a convincing argument, I'm going to :ship: this RFC.

jaymzh:
Wait - the RFC still allows for arbitrary languages of recipes, which, based on your summary of the IRC meeting seems like a no-go with everyone. Did I miss something? I think having recipes being able to be written in arbitrary languages is a really bad idea for all of the reasons @lamont-granquist and @coderanger and @danielsdeleo mentioned.
:-1:

coderanger:
@jaymzh Only via external support, core chef will only support Ruby.

jaymzh:
@coderanger I think this will be a nightmare. People will be posting recipes written in java and ocaml and whatever else. I'm all for adding rb support for the missing ones like databags, but I think allowing people to arbitrarily add their language-of-the-day cookbooks is going to be crazysauce. As @lamont-granquist  pointed out, even if it's not "core chef" or fully supported, it'll still make it to the mailing list and make every newbie's life harder when they download 4 community cookbooks in 4 different languages. Someone will copy-paste code from the list archives, but it'll be python which won't work in core ruby, etc.

coderanger:
@jaymzh I think you are way overestimating things. It seems really unlikely anyone beyond me is going to invest the time to write dialects that are anything but a mild change. Writing them is Very Hard. Yes I'll probably release my JS recipe dialect (with some pretty strong warnings that it has a habit of segfaulting if you call stuff incorrectly) but I think its just way over stating things to assume instant and massive confusion. I would be honestly surprised if many people know about the feature in a years time.

jaymzh:
Then why have it in the first place? Your assertion in the RFC is that people *want* to write in other languages, which mean certainly someone will write a python language and then now newbs have to make more difficult and unclear choices. IF you don't think anyone will use it, why add the complexity?

coderanger:
@jaymzh Because it lets me support other people better? I would like to think more people will write them, but its going to be vastly more consuming than contributing. And if the question is "Do you think you (Noah) can write a recipe dialect plugin safely which preserves the semantics and intent of Chef" I would say the answer is yes. That said, I think uptake will be much higher on the other data types. There is already another RFC that is literally wanting to add a Ruby dialect for data bags, that on its own shows interest.

coderanger:
If your argument is "we should add this to every data type except recipes because people might misuse it or get confused", then I'm not really sure that's hugely defensible. Chef recipes already give enough rope to hang yourself (and then light yourself on fire, summon a plague of locusts, etc). I spend a large part of my day answering questions from people that copy-pasta bits of recipe code in ways that won't work, I can't imagine that changing much regardless.

lamont-granquist:
I'm still not so wild about writing roles in YAML and templates in Haml.

someara:
This thread makes my head hurt.

martinb3:
This thread is full of awesome comments I fully intend to quote later :) It seems to me like the proposal from @coderanger will result in cleaned up, more extensible code. And the disagreement is "What if people extend it in a way that makes Chef less approachable and supportable?"

I don't see this as any different than someone coming to the software or the mailing list with any other odd duck that we can't help with. And I think there is *tons* of prior art for features that aren't exposed to vanilla newcomer users of Chef (chef server ACLs feel like a good example of this).

While my opinion doesn't carry any official weight, :+1: from me. I also see it potentially being a *major, major* boon for Windows (and other platforms) support down the road, as allowing for other dialects could lead to experimentation and/or adoption on more platforms.

jonlives:
I still have some reservations about this opening a can of worms but with that said I have confidence in @coderanger's ability to implement and support the initial dialect he's proposing, so I'll :+1:

someara:
Dialects imply a language... and as far as I know, there is no formal
specification of Chef As A Language.

Where does Chef's "public API" begin and end?

Is it just the list of core resources on the docs page?

What about run_contexts, notification / subscriptions?
use_inline_resources? why_run?, the LWRP DSL, etc.

Are we just talking about the pre-compiled bits like metadata.rb, roles,
and data_bags?

What does it even *mean* to use multiple languages in a chef-client run?

Should resource_collections serialized into a neutral format? Right now its
Ruby objects.

We've already seen things like POSHChef pop up that implement a parts of
Chef, how do they interact?

-s

adamhjk:
Let's focus first on the obvious: it's healthy to have a cleaner, extensible core chef. We can then experiment outside the mainline with crazy ideas, and see what sticks. I agree that my assumption is that, in this case, more isn't better. But! I could
Be persuaded by compelling working code.

adamhjk:
Merge it, RFC editors. Make sure its clear we're not accepting new dialects in core with this RFC.

PR #77, opened 2014-12-15
Attribute API 2

danielsdeleo:
**NOTE:** This RFC is incomplete, the specification of the new feature is TBD.

This RFC describes a new attributes API to be added alongside the current one. The goals of this new API are:
- Separate desired and observed state.
- Structure the syntax in a way that "trainwreck" errors
  (`NoMethodError` for `[]` on `nil`) can be avoided.
- Structure the syntax in a way that attribute tracing can be
  implemented simply.
- Separate node-specific persistent storage from
  cookbook/role/environment attributes.
- Structure the syntax in a way that can be implemented efficiently.

This RFC is adapted from https://github.com/opscode/chef-rfc/pull/9 which is retracted.


martinb3:
:+1: for `Block with instance eval, method chaining`. I definitely don't like seeing dots that have syntactic meaning inside strings. I also think there's a lot of cognitive dissonance with node data and the Hash data structure, and any compatibility layer will likely mean people will have to learn (a) the new data structures and (b) the quirks of any compat or translation later between the new data structures and regular hashes.

I'd also like to see arrays come back to wider use, and I think distancing the node data structures from Ruby hashes is an opportunity to make arrays as viable data structure again for node data, as @jaymz mentioned.

lamont-granquist:
@danielsdeleo there needs to be some kind of statement added that RFC18 is rejected for this implementation and that RFC18 needs to be amended so that we can do this.

@adamhjk also feels very strongly from a UX perspective that square brackets should be entirely dropped because attributes aren't hashes.  from a code perspective I also agree that the method chaining required by square brackets and the you-dont-know-when-it-ends problem that it creates when trying to solve problems that we want to solve with attributes means that square brackets are hard to implement consistently.  so the square brackets options should be dropped.  the single-square-bracket-operator-with-a-splat options are clever but not very ruby and should also be dropped.

the chef-sugar method_missing alternative also needs to get dropped, nobody wants that one either and it conflicts with one of the stated goals that we're trying to avoid.

lamont-granquist:
I'd also like to state that symbols will not be supported in the new API and only strings will be supported.

adamhjk:
General consensus is strings only.

jaymzh:
Any thoughts here on handling arrays, as I mentioned above? It seems an entire usecase dropped by this API at this point, which makes it a no-go for lots of people...

lamont-granquist:
In a way i kind of like the idea of just making this work:

```
node.default['foo']['bar'] = []
node.override['foo']['bar'] << 'thing'
```

So that when you have to autovivify an attribute it walks back up the precedence tree and creates a new hash or array if it finds one at a lower precedence level.  Helps to keep the schema consistent.

Obviously, insert whatever the new syntax turns out to be...

jaymzh:
I agree 100%, @lamont-granquist I'd love something like that. I just don't want to ratify a new syntax that doesn't take this into account, and the current syntax assumes key-values at every level, and nothing array-like.

adamhjk:
Lets get @kallistec to weigh in on arrays

danielsdeleo:
In the method chaining syntax it should be pretty straightforward to make array append do what you mean; e.g., in the given example `default_attr('foo', 'bar').append('thing')` the DSL owns the object that's getting `#append` called on it, so it can infer you want an Array. We could alias that as `#<<`, as well. One question though, is how these should handle conflicts; if I already have a non-array value, does an append operation nuke it, or raise an error? I'm partial to the latter as I think 99% of the time it would be a user mistake, but could be persuaded otherwise.

As for the array of hashes thing, I'll have to think about that. I will note that every example of it that I've seen can be implemented as a map; in this example, for instance:

```ruby
[
  {
    'name' => 'bigbla',
    'val' => 'doof',
  },
  {
    'name' => 'smallbla',
    'val' => 'book',
  }
]
```

The "name" attribute is probably unique so you could do this instead:

```ruby
{
  'bigbla' => {
    'name' => 'bigbla',
    'val' => 'doof',
  },
  'smallbla' => {
    'name' => 'smallbla',
    'val' => 'book',
  }
}
```

jaymzh:
I think it should raise an error for sure. Data-type conflicts are totally user error.

To your other point, there's no reason to limit what people put in an array. The reason I brought up this example, is because if it's inside an array, given this new syntax, does the hash inside that get treated like a hash or like an attribute? For example, can I do:

```
attr('foo', 0, 'bar') = 'thing'
```

Because that's keeping parity with what we have now (and I need). Having to do something like this would suck:

```
elmt = attr('foo')[0]
elmt['bar'] = 'thing'
attr('foo')[0] = elmt
```

timurb:
Setting is really great (± rspec-like syntax) but dotted notation for getting is what I’d really miss.

`attr('nginx', 'upstart', 'foreground')` is much heavier to read and type than `attr('nginx.upstart.foreground')` especially when you use 3+ components in the attribute name.

By the way what is the point in differentiating keys containing dots to keys not containing them?
Why would you want to have `attr('foo', 'bar')` meaning something different to `attr('foo.bar')`?
Is there any reason not to treat them the same?

danielsdeleo:
Whenever the attribute is a hostname, IPv4 address, or used by another system that uses dots to separate keys. For example, on my Mac, everything reported by `ohai network/settings` is a key with dots.

tyler-ball:
An alternative is to write `attr(%w{nginx upstart foreground})` - I think this is as easy to read as the dotted notation

adamhjk:
and in case it didn't make it into the thread - I've vetoed the dot
notation. We're removing ambiguity, not keeping it.

Love,
Adam

timurb:
@danielsdeleo you are right, we'll have a problem with iterating over dotted keys when you need to extract a subkey of that, like `attr('my_hostnames').each { |h| h['someattr'] }` -- this would be hard to implement when attributes are directly mapped to a dotted notation.

By the way, what will be a way to handle such cases with the new syntax? Will it be like I've written here or is it going to be some different syntax?

timurb:
... or `attr(h, 'someattr')` to avoid referring to them as hashes?

new/rfcXXX-new-attributes-api.md
```diff
@@ -0,0 +1,151 @@
...
+attribute structure is missing, as this error will occur in an identical
+way if either `node[:foo]` or `node[:foo][:bar]` are `nil`.
+
+The UX of this error case cannot be significantly improved without
+abandoning the nested Hash model. If an object other than `nil` was
+returned for accessing a missing key, then code like
+`do_something if node[:some_attribute_present]` would break (Ruby does
+not allow custom false-y types), and monkey patching `nil` is
+error-prone because `nil` is an immediate object (there is only one
+instance).
```
lamont-granquist: We might want to call this out as the reason why method chaining syntax MUST NOT be used.  I actually misunderstood the problem on first reading, and I don't think its obvious to the casual reader exactly what the implications are to the API design.  Any kind of `node['a']['b']['c']` or `node.a.b.c` kind of method chaining will recreate this problem and therefore CANNOT be part of the new design.

danielsdeleo: I didn't want to clutter this section too much, but I guess that kind of thing is what the rationale section is for.

danielsdeleo: Ok, added a discussion in the rationale section.

lamont-granquist:
Another interesting bug with node attributes is that they get round-tripped to JSON which can be problematic.  A foodcritic user has submitted a bug report to add checking for using symbol values in node data and comparing node data against symbol values:

https://github.com/acrmp/foodcritic/issues/286

Catching that is actually easier to do in code and not static analysis by not allowing the API to set a value that wasn't valid JSON and couldn't be roundtripped through JSON and return as itself.

PR #79, opened 2015-01-06
RFC for Ruby data bag items

ekrupnik:


coderanger:
This is somewhat linked to #71 as that would provide a basis for the implementation.

new/rfcXXX-ruby_data_bag_items.md
```diff
@@ -0,0 +1,131 @@
...
+RFC: unassigned
+Author: Eric Krupnik <ekrupnik@copyright.com>
+Status: Draft
+Type: Standards Track
+---
+# Ruby data bag items
+
+Currently, Chef only supports JSON data bag items, and not Ruby. Chef **does**
+support having other configuration-based files (such as `environments` and `roles`)
+be written in either JSON or Ruby. JSON does not allow comments, so it becomes
```
lamont-granquist: the JSON implementation in Chef allows comments and yajl-ruby allows comments.

ekrupnik: @lamont-granquist Can you please point me at the documentation (or an example) of a Chef data bag item (JSON) which has comments. I have not been able to find anything online regarding data bag items supporting comments. Because by default, I do not believe that JSON supports comments.

lamont-granquist: data bags are missing tests, but environments, nodes and roles have tests:

https://github.com/opscode/chef/blob/master/spec/data/object_loader/environments/test.json
https://github.com/opscode/chef/blob/master/spec/data/object_loader/nodes/test.json
https://github.com/opscode/chef/blob/master/spec/data/object_loader/roles/test.json

ekrupnik: Thank you @lamont-granquist. It would, however, be very helpful to have this comment support noted somewhere, for example in the [data bag documentation](http://docs.chef.io/data_bags.html).

lamont-granquist: You can open an issue or PR against https://github.com/opscode/chef-docs

ekrupnik: @lamont-granquist I've submitted [this pull request (#432)](https://github.com/opscode/chef-docs/pull/432) against the [chef-docs](https://github.com/opscode/chef-docs) repository to document comment support in data bag items.

jonlives:
I'm not super in favour of this RFC given that comments are supported in JSON data bag files, and that was the main motivation for adding Ruby data bag items.

IMO it makes sense to have data bag items as "dumb" pieces of data rather than yet another place where arbitrary ruby code can be added - worth keeping it simple, I feel.

lamont-granquist:
The .rb files are also not supported by ChefFS and 'knife diff' in particular and arbitrary code makes that whole problem somewhat annoying.  I've generally moved towards .json now for everything exclusively.  It'd be nice if the upstream yajl repo would accept some patches for allowing trailing commas and single and double quoted strings and other JSON5 extensions.

coderanger:
My biggest concern is the proposed format isn't like the other Ruby DSLs. It is a single Ruby object expression at the top level of a file, instead of a bunch of method calls on an object. This is technically doable, but definitely a different kind of beast.

adamhjk:
This PR is closed in preference to this work being done as part of the dialects RFC.

PR #81, opened 2015-01-15
XML file editing as a first class resource

ranjib:
Enables user to inject values in an XML file based on XSLT mappings

### DIFF
--- a/new/rfcXXX-xml-fie-resource.md
+++ b/new/rfcXXX-xml-fie-resource.md
@@ -1,0 +1,56 @@
+---
+RFC: unassigned
+Author: Ranjib Dey (ranjib@linux.com)
+Status: Draft
+Type: Standards Track
+---
+
+# XML file resource
+
+XML file resource gives the ability to control parts of an XML file
+idempotently.
+
+# Motivation
+
+A vast majority of JVM based applications rely on XML as their configurations
+(e.g Jenkins, GoCD, Tomcat, JBoss etc). Sometime these files are augmented
+into multiple xml files, sometimes they are just a single giant config file.
+As a chef user, when I automate these applications I want to inject only bits
+and pieces of values that are known, in a larger configuration file.
+For example the `pipelines` in  GoCD config XML, or the `antiResourceLocking`
+attribute in Tomcat configuration xml can be managed by Chef without controlling
+the content of the entire XML file ( which often contain temporal or version
+specific data).
+
+For reference, augeus XML lense provides similar features.
+
+# Specification
+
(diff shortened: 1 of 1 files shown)


coderanger:
Why should this be in core instead of a cookbook/gem?

ranjib:
because we encounter this quiet frequently, currently we use it as an lwrp only. But i think this has wide enough use case that we can consider having this in core.

coderanger:
If it is mostly with Java-y things, maybe include it in the java cookbook?

coderanger:
Or more likely, in its own cookbook that is a dependency of java.

ranjib:
there are several non-java apps (ossec for example). We can write those cookbooks or those cookbooks can offer much better lwrps if this resource is in core.

coderanger:
I don't see why it would be a better LWRP if it is in core, the code is the same regardless of how it is packaged.

lamont-granquist:
Depends on nokogiri, presumably?  So this would effectively write a check on our time for getting nokogiri installing on EL5 and AIX and Solaris 10?

ranjib:
@lamont-granquist yeah, i was thinking of implementing it using nokogiri. :-( . i guess it will be difficult then

lamont-granquist:
Yeah, I suspect there is zero time to port nokogiri to AIX in the next 6 months at least.

erichelgeson:
There is a cookbook that does something similar - https://supermarket.chef.io/cookbooks/xmledit

jonlives:
I'm not super in favour of this being in Core, I agree with @coderanger that it'd probably be better off in a cookbook as an *WRP.

lamont-granquist:
To reiterate the IRC discussion, would like to see this as a cookbook first, and using rexml.

I'd also like to see a bit of higher level design to as we head down this road as to how json and yaml and ini all fit into this in addition to xml.   I really don't want to see the organic growth which led to the confusion over all the different file/template/cookbook_file/remote_file resources and their confusing names.

coderanger:
Also a minor request, if this is going to be built for delta editing, can we call it `xml_edit`?

ranjib:
Following is the first cut for an REXML based HWRP
https://github.com/GoatOS/xml_file

ranjib:
I'm closing this. I have implemented a REXML based HWRP prototype as discussed in the IRC, and i'll be using it. But given no one else has upvoted this, and consuming the HWRP from cookbook is a decent workaround, im closing this. I'll be happy to reopen this if others show any interest.

juliandunn:
I think that's sufficient for now; thanks for the cookbook/HWRP @ranjib

PR #82, opened 2015-01-15
Allow specifying partial templates during bootstrap

ranjib:


lamont-granquist:
hmmm, i'd kind of like to go more in the direction of dan's bootstrapper code and less doubling down on the erb templates.

ranjib:
@lamont-granquist you have any reference link for dan's bootstrapper code? I'd love to see it :-)

lamont-granquist:
have to ask @danielsdeleo for it

danielsdeleo:
On my personal github account, here: https://github.com/opscode/bootstrapper

I'd stopped working on it b/c @jkeiser was making a lot of progress with Chef Provisioning (née chef-metal) and he incorporated a lot of the features I'd planned on adding (in particular, caching chef-client locally to install on the remote node), but maybe there's always gonna be a desire for a command line imperative bootstrapping functionality? If so, maybe we can get shared code from either my project or provisioning into core chef/knife, so that a new bootstrap command (and related cloud plugins) could be built from it?

lamont-granquist:
I'm going to pull the client key generation into knife bootstrap shortly, merged with some chef-vault patches.  Sort of planning on taking pieces of that code and incrementally merging them into knife bootstrap...

adamhjk:
We are, in general, +1 on this RFC. What we would like to see is some more explanation of how it would be implemented, and what would be possible with it.

jtimberman:
Use cases I can think of that would make this super handy:

1. Be able to inject commands, like `apt-get update` that aren't present in the current chef-full.erb, because that template doesn't assume Debianish systems.
2. Ability to install e.g., `curl` or `wget` that may not be present on base OS installations, especially minimal system images like containers.
3. Inject additional default configuration in the `/etc/chef/client.rb` file, because you want other settings present, such as HTTP proxy, disabled Ohai plugins, etc.

These things are all possible with custom bootstrap templates but that can vary in unwieldy-ness and can add quite a lot of burden to the user.

new/rfcXXX-bootstrap-partial.md
```diff
@@ -0,0 +1,49 @@
...
+Chef allows knife based bootstrapping via standard templates as well
+as custom templates. But theres no way to reuse the bootstrapping logic
+of standard templates. For minimal distributions where tools like
+wget and curl is absent, one has to roll his/her own bootstrap template,
+which most often is a copy of the standard template with extra commands
+to install those tools. Over the time, this becomes tech debt, as chef evolves
+and with it the standard template changes, incorporating new bootstrap options.
+
+Scope of this RFC is to offer users a mechnism to execute custom logic before
+the standard bootstrapping process, instead of specifying the entire bootstrapping
```
btm: This sounds like a separate template that runs before the bootstrap template. An example was given that you could ```apt-get update && apt-get install wget```. In that case, I think this should be described as a pre-template.

If we want true partials, I'd want a way we can describe the chef-full template into steps, and it would be best if we could agree on those steps in the RFC. For instance, right now the templates call the "start_chef" step which is defined in core/bootstrap_context.rb instead of the template. The partial should be able to replace a step. There could be a pre step that does nothing in the base template that a partial could hook. There could be a start_chef step. Maybe a config_files step and a install step.

For discussion:

* preinstall (empty)
* install
* postinstall (empty)
* config_files
* start
* finish (empty)

adamedx:
Yes, this would be great for grabbing cookbooks first and then running local mode in the bootstrap.

adamedx:
Also great for setting environment variables, configuring ssl certs, or even obtaining databag secrets in some out of band fashion. Lots of use cases I think.

thommay:
I'd like to see this updated to use @btm's suggestion to provide a set of steps that could be replaced by partials.

miketheman:
I had previously committed to providing an example of how one might use partials during bootstraps with no modification to Chef/Knife at all.

Here's one way we've done it, and continue to do it successfully every day: https://gist.github.com/miketheman/4bb1b16618bee1365893

ranjib:
im closing this one. along with the github release RFC. The current bootstrapping implementation forces us to swell the erb rendering logic, as well as imposes major UX challenge for new features. But we are likely to need more and more complex bootstrapping options . We need a broader discussion around this, before we start adding individual features.

PR #83, opened 2015-01-15
Recipe DSL method for event handlers

ranjib:


lamont-granquist:
:+1:

coderanger:
:+1:

This might have some level of overlap with the audit mode work, since that involves some extra event dispatch stuff for test events.

jonlives:
The language here could do with a little clarification I think to distinguish between report and exception handlers, and formatters / event dispatcher subscribers - it looks like you're talking about the latter here.

coderanger:
Request for IRC, use something slightly less generic than `on` if there is going to be a DSL. Counter-proposal was `on_event`.

danielsdeleo:
I think you lose a lot of value if this is in a recipe. For example, 

```ruby
on :run_start do
  Etcd.lock_acquire(lock_key)
end
```

cannot possibly work because `run_start` occurs way before cookbooks are evaluated.

So I think this really needs to exist outside of recipes. What I'd like to see is:

* Add a `conf.d` convention for Chef's config.
* Tweak the DSL so it doesn't require contextual evaluation. For example:

```ruby
Chef.event_handler do
  # DSL
end
```

Doing it this way also means that you could define handlers in cookbooks libraries, if you don't need access to events that have previously occurred. Bonus points if the DSL can detect events that have already passed and warn you, but that could be hard to implement. Anyway, if you do need run start and such, then you can `cookbook_file` your code instead. I could see adding a syntax-sugar-y resource like `chef_config_file "foo.rb"` which handles the details of installing a cookbook_file to Chef's con.d being handy as well.

martinisoft:
I would absolutely love this. Especially if you have a cookbook that could provide handler mechanisms or you need to incorporate reporting tasks in a cookbook. :+1:

coderanger:
:+1: for @danielsdeleo's idea.

martinisoft:
If this gets treated as a resource, would or could you also trigger resource notifications via these handler calls?

lamont-granquist:
If we add it to the Recipe DSL we should catch the case where a user attempts to `on event :run_start` and throw a useful error message rather than not doing anything.

jtimberman:
For what it is worth, the chef-client cookbook's config recipe supports a [conf.d style configuration](https://github.com/opscode-cookbooks/chef-client/blob/e0efda8cd454d79caad19d88f9691d006d724551/templates/default/client.rb.erb#L66-L68) currently.

lamont-granquist:
it might be nice to bake that directly into core chef @jtimberman and have that behavior be the default

jtimberman:
I can dig it. Separate rfc?

lamont-granquist:
PR?  its a pretty minor feature

ranjib:
@danielsdeleo thanks for catching that :-) . I'll fix the example. We currently use the conf.d style handler and in some scenarios they can be painful. For example, conf.d style allows us to drop an .rb file to load report handlers, which often involve requiring some libraries. This has two implications:
1) During omnibus upgrade (when ruby and the entire system is being updated), initial chef-client runs fail as those configs point to now missing libraries.
2) The chef_handler resource itself is not idempotent. the way it tries to launch an load the rb files, involves chef reconfigure. (i think this can also be fixed though)

Building an authoritative conf.d style pattern in core will be helpful. Doing conf.d style as a resource/provider is hard, infact im not sure if you can build it reliably. Because deleting un-managed files logic needs to be run at the end of the file.

What is the best way to package the handler inside the cookbook. without writing additional config file on disk.
What about adding a separate segment as /handlers and include_handler directive in recipe DSL ?

The recipe dsl still feels a lot more helpful for other use cases like i want to drop a hipchat notification if a key config file has changed.. etc

danielsdeleo:
If you attach the DSL to a top-level object (the `Chef` class in my example), then you can actually use it anywhere. It'd be valid in the context of eval-ing config files, so you could put it in a conf.d type structure and you'd have access to all the events that occur after `config_load_start` (or whatever that event is called), but you could also put it in a library, which would mean you could wrap the recipe compile and converge portion of a Chef run with a lock/unlock if you needed to (of course in this case you'll still have to deal with the problem of installing 3rd party libraries which would normally occur in recipes). You could also put `Chef.event_handler(args) do; end;` in a recipe if you didn't want Chef to see it until recipe compile time.

As for the chicken/egg problems of extending Chef within a cookbook and then using that extension, I think that's a more general issue, and we should come to consensus on a solution to that in a different venue (dev ML, separate RFC, etc.). My quick two cents is that installing libraries for use in a Chef run is part of the process of bootstrapping Chef itself and we'd get a lot of benefit if we stopped trying to do it inside the main Chef run.

ranjib:
@danielsdeleo i have updated the spec. Can you review that i have captured it correctly ? i.e. the `on` DSL method is evaluated inside Chef.event_handler context.

new/rfcXXX-event-handler-dsl.md
```diff
@@ -0,0 +1,81 @@
...
+## Motivation
+
+Chef has an extensive event handler mechanism, but incorporating
+some custom logic with any of the events is onerous process which involves
+subclassing the based evant handler and adding it via the config. This RFC
+proposes a recipe DSL method to ease this. For new chef users this increases
+the entry barrier, while for experienced chef users this amounts to writing
+boilerplate code.
+
+## Specificatiin
```
danielsdeleo: typo

new/rfcXXX-event-handler-dsl.md
```diff
@@ -0,0 +1,81 @@
...
+
+Allow cookbook authors easily add custom logic on Chef events.
+
+
+## Motivation
+
+Chef has an extensive event handler mechanism, but incorporating
+some custom logic with any of the events is onerous process which involves
+subclassing the based evant handler and adding it via the config. This RFC
+proposes a recipe DSL method to ease this. For new chef users this increases
```
danielsdeleo: Might want to rearrange these sentences. When I first read this it seemed like it was saying the recipe DSL method increases the entry barrier, etc.

danielsdeleo:
@ranjib What you specified is exactly what I had in mind. Have a few wording/typo comments above, but otherwise am :+1: on this.

new/rfcXXX-event-handler-dsl.md
```diff
@@ -0,0 +1,81 @@
...
+RFC: unassigned
+Author: Ranjib Dey (ranjib@linux.com)
+Status: Draft
+Type: Standards Track
+---
+
+
+# Recipe DSL method for event handler hooks
+
+Allow cookbook authors easily add custom logic on Chef events.
```
juliandunn: "authors to easily add"

new/rfcXXX-event-handler-dsl.md
```diff
@@ -0,0 +1,81 @@
...
+
+
+## Motivation
+
+Chef has an extensive event handler mechanism, but incorporating
+some custom logic with any of the events is onerous process which involves
```
juliandunn: "is an onerous process"

new/rfcXXX-event-handler-dsl.md
```diff
@@ -0,0 +1,81 @@
...
+subclassing the based evant handler and adding it via the config. This RFC
```
juliandunn: "event handler"

new/rfcXXX-event-handler-dsl.md
```diff
@@ -0,0 +1,81 @@
...
+proposes a recipe DSL method to ease this. For new chef users this increases
+the entry barrier, while for experienced chef users this amounts to writing
+boilerplate code.
+
+## Specificatiin
+
+Currently chef client sets up couple of default handlers (doc, base) during
+initialization. An additional empty event handler (i.e. just subclass
+of base handler without any custom logic) can be added along side the
+existing hanlder which will used as a space holder for user specific hooks.
```
juliandunn: "handler"

also I think you mean "placeholder"

new/rfcXXX-event-handler-dsl.md
```diff
@@ -0,0 +1,81 @@
...
+
+A top level (::Chef) method will be introduced (`event_handler`) along side the
```
juliandunn: "alongside"

ranjib:
@juliandunn thank you for proof reading :-) . I have addressed all your comments.

new/rfcXXX-event-handler-dsl.md
```diff
@@ -0,0 +1,80 @@
...
+
+
+# Recipe DSL method for event handler hooks
+
+Allow cookbook authors to easily add custom logic on Chef events.
+
+
+## Motivation
+
+Chef has an extensive event [disptach mechanism](https://github.com/chef/chef/blob/master/lib/chef/event_dispatch/base.rb).
```
thommay: "dispatch"

thommay:
Besides the one trivial typo, I'm very much :+1: for this.

ranjib:
@thommay fixed

adamhjk:
I would like to say that @ranjib is awesome for all the great RFCs. Folks are going to dig this. Ship it, @chef/rfc-editors

PR #84, opened 2015-01-21
Add chef_version and ohai_version fields to cookbook metadata

jonlives:
Increasingly, as momentum on Chef development continues to grow, it is necessary to release modifications to the recipe DSL or core resources which break backwards compatibility and result in cookbooks potentially breaking under old versions of chef-client. Further, it is often necessary to make these changes in minor chef-client releases rather than waiting multiple months for the next major release cycle. There is currently no way to surface this version requirement to Chef users.

This RFC proposes the addition of the "client_compat" field (name up for discussion) to cookbook metadata to indicate the required version of chef-client needed to run that cookbook.


new/rfcXXX-client-compat-metadata.md
```diff
@@ -0,0 +1,29 @@
...
+RFC: unassigned
+Author: Jon Cowie (jcowie@etsy.com)
+Status: Draft
+Type: Standards Track
+---
+
+
+# Add client_compat field to cookbook metadata
+
+Increasingly, as momentum on Chef development continues to grow, it is necessary to release modifications to the recipe DSL or core resources which break backwards compatibility and result in cookbooks potentially breaking under old versions of chef-client. Further, it is often necessary to make these changes in minor chef-client releases rather than waiting multiple months for the next major release cycle. There is currently no way to surface this version requirement to Chef users.
```
lamont-granquist: So as far as wording goes I still strongly object to this being described as introducing resources which "break backwards compatibility."  That is not what we're talking about.  In core chef these are all minor releases with *STRICTLY* backwards compatible changes in them.  As those changes are adopted by *COOKBOOKS* it produces backwards-incompatible cookbook code which cannot be used on prior versions of chef-client.

jonlives: Fair point, I've clarified the wording to indicate that we're only talking about cookbooks here.

lamont-granquist:
This is an often requested feature:

https://tickets.opscode.com/browse/CHEF-2251
https://tickets.opscode.com/browse/CHEF-3591
https://tickets.opscode.com/browse/CHEF-4524

FWIW, I like "ohai_version" and "chef_version" better than "chef_compat"

There's a possibility that a feature could be backported to a prior version of chef so that you'd need two ranges of support (chef_version >= 11.18.0 || chef_version >= 12.2.0), I don't know if we care about that.  I know we've got an RFC now that states that we don't backport features, but that wasn't what happened with chef 10+11.

jonlives:
Yeah I'm not totally sold on "client_compat" myself, chef_version probably makes more sense here.

lamont-granquist:
Another way to do this would be to support closed ranges and invert it and mark it as "broken" instead of "supported" -- idea that just popped into my head thinking about the fact that it could be used to mark individual chef-client versions as broken due to regressions.

jonlives:
If other folks agree on chef_version and ohai_version, I'll amend the RFC to add both of those fields. 

@lamont-granquist interesting idea there, could possibly even stick with regular version specifiers and allow ! to indicate specific broken versions?

jeremyolliver:
For specifying multiple version ranges, supporting semver style pessimistic version constraints is probably a good idea in the similar way that ruby/bundler does (eg. `chef_version ["~> 11.18.0", ">= 12.2.0"]`) where `~> 11.18.0` means `>= 11.18.0 && < 11.19`

coderanger:
Worth noting that it we use the existing solver stuff for this, it only supports a single version constraint per entry. Might mean either piggybacking some general improvement work there or that multi-range stuff might come later. Either way, just a mildly relevant impl detail.

coderanger:
:+1: :+1: :+1: on the idea. As a workaround for old Chef versions we could use a pattern like `chef_version '~> 12.0' if defined?(chef_version)`.

jaymzh:
:+1:

jonlives:
I've pushed a commit to change the originally proposed field name to "chef_version", and added the "ohai_version" field per @lamont-granquist's suggestion

thommay:
:+1:

miketheman:
:+1: curious, though, what the behavior would be when you get into dependency trees, and it doesn't match. Does the entire run fail right there?

jonlives:
@miketheman the logic for making the run fail is up for discussion, but I'm envisaging the entire run failing if any cookbooks fail the chef_version check, the same as if a cookbook dep is missing.

jonlives:
This RFC is just to add the fields, initially btw :)

coderanger:
An easy first pass would be not involve the version solver, just when it goes to run it could check and fail the run with an error. At least that would be better than mystery failures later on. Would be valuable even if it was just human-facing though.

lamont-granquist:
The depsolver runs on the server now, and would run on the workstation in the future.  Neither of those places have accurate information about the chef-client that is going to run the cookbook.

jonlives:
I've added clarification of how chef-client will behave w.r.t these attributes, y'all lemme know if further detail is needed or anything is unclear :)

miketheman:
:+1: to the changes. As to the depsolver stuff, I'm find with the concept of the client not doing depsolving for this, rather a sanity check of sorts, pre-run validation that the cookbooks compiled in have metadata flags that don't restrict those cookbooks from being run.

jonlives:
@miketheman right - that's very much the sort of thing I was thinking of too.

jonlives:
@adamhjk this is all good to go, just needs a decider seal-of-approval :)

danielsdeleo:
I'm totally :+1: on this.

That said, I think compatibility will be a bit tricky. To my knowledge, Chef Server disallows unknown keys in cookbook metadata when uploading, so it will return a 400 if we try to post that to the server. There's a couple of ways to handle this. I'm working with some folks at Chef Software to implement [RFC022](https://github.com/chef/chef-rfc/blob/master/rfc022-arbitrary-cookbook-identifiers.md) right now; since that's a new API we can greenfield and do whatever we want. I'll make a note to ensure that we relax the validation the new API.

To implement this RFC for non-policyfile usage, though, we'll probably need to modify the validation behavior for the existing /cookbooks endpoint and have some fallback mechanism for `knife` and friends to talk to older Chef servers. Maybe we catch the 400 and then regenerate the cookbook JSON in "metadata strict mode?"

lamont-granquist:
It might be very confusing to have policyfile behaviors and non-policyfile behaviors.  Even if we get to greenfield it for policyfile stuff, we may want to make sure that we can make the behavior in non-policyfile-land consistent.

Probably we should either start versioning the metadata schema and/or relax the validation completely and get back to having a schemaless database.  I'd like to see the validation dropped almost entirely except for those fields which the server actually cares about (name, version, deps (for now)).

danielsdeleo:
@lamont-granquist yep, I agree. It's just about how we get there from where we are now, especially with folks who may be on Chef Server 11 (or private chef variants), or on Chef Server 12.0.x. Even if we did a major version bump, it'd be really painful to have a flag day. On the client (knife) side, I think this general logic would work?

```ruby
begin
  upload_with_all_metadata
rescue HTTPServerException => e
  raise if e.code == "400"
  # another thing we can do is check the headers in the response
  upload_with_only_strict_metadata
  # we only get here if the first 400 response was caused by "extra" metadata fields
  warn("Your chef server doesn't support extended metadata fields. Features X,Y,Z won't work")
end
```

lamont-granquist:
Yeah, you had the idea before that knife could get back the chef server version string while its uploading the sandbox data and then could switch behaviors for the last POST/PUT.  We could also have a config option / command-line-flag to force knife to strip new metadata out of a cookbook (and warn as it does) and then POST/PUT that to an old server.  Old servers with old knife versions and new cookbook metadata might be able to be handled by guarding the metadata with respond_to? even though that's kinda gross.

ekrupnik:
A big :+1: on this idea. It would be great to have the minimum Chef version defined in a cookbook's `metadata.rb` file, where other versions are already defined (like cookbook and dependent cookbook versions).

gregkare:
As more methods are added to a cookbook's metadata it will become really ugly if every new attribute needs to be guarded with an if statement (see chef/chef#2937)

Here's a naive way to avoid the problem:

```ruby
class Chef::Cookbook::Metadata
  def method_missing(method, *args, &block)
    Log.warn("#{method} is not defined in this version of Chef")
  end
end
```

I had never looked at the code for `Chef::Cookbook::Metadata` and was surprised to see the metadata DSL just crashes when it encounters an unknown method. What do you think?

jonlives:
@gregkare that's a bit out of scope for this RFC - all we're talking about here is adding metadata fields that'll make it possible for chef-client to gracefully fail if any cookbooks being run use more recent features than it implements.

adamhjk:
I am the decider, and I approve this message. :+1:

PR #85, opened 2015-01-22
RFC000: Updating RFCs

btm:
Specify that 'Standards Track' RFCs require a new RFC to change, while
Informational and Process RFCs may be updated by pull request.

### DIFF
--- a/rfc000-rfc-process.md
+++ b/rfc000-rfc-process.md
@@ -116,6 +116,18 @@ the original one should be marked as Replaced.
 * **Final** – The RFC has been implemented. *(Standards Track type only)*
 * **Replaced** – The RFC has been superseded by another RFC.
 
+### Changing an Accepted RFC
+
+An accepted RFC may be modified in two ways, depending on the type of RFC:
+
+1) To support software implementations meeting the specifications of an RFC,
+Standards Track RFCs must be replaced by a new RFC. This is done by specifying
+the RFC that is being replaced using the Replaces header in the metadata of the
+new RFC.
+
+To simplify their evolution, existing Informational and Process RFCs may also
+be updated by pull request.
+
 ## RFC Template
 
 ```markdown
@@ -124,6 +136,7 @@ RFC: unassigned
 Author: Alan Smithee <asmithee@example.com>
 Status: Draft
 Type: <Standards Track, Informational, Process>
+<Replaces: RFCxxx>
 ---
 
 # Title
--- a/rfc000-rfc-process.md
+++ b/rfc000-rfc-process.md
@@ -120,13 +120,15 @@ the original one should be marked as Replaced.
 
 An accepted RFC may be modified in two ways, depending on the type of RFC:
 
-1) To support software implementations meeting the specifications of an RFC,
-Standards Track RFCs must be replaced by a new RFC. This is done by specifying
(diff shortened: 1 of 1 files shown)


rfc000-rfc-process.md
```diff
@@ -124,6 +136,7 @@ RFC: unassigned
 Author: Alan Smithee <asmithee@example.com>
 Status: Draft
 Type: <Standards Track, Informational, Process>
+<Replaces: RFCxxx>
```
coderanger: Shouldn't be in the template since this is very rare.

rfc000-rfc-process.md
```diff
@@ -116,6 +116,18 @@ the original one should be marked as Replaced.
 * **Final** – The RFC has been implemented. *(Standards Track type only)*
 * **Replaced** – The RFC has been superseded by another RFC.
 
+### Changing an Accepted RFC
+
+An accepted RFC may be modified in two ways, depending on the type of RFC:
+
+1) To support software implementations meeting the specifications of an RFC,
+Standards Track RFCs must be replaced by a new RFC. This is done by specifying
```
coderanger: Standards Track can be edited as needed. Only Final RFCs are locked down because that means the feature or change is in a release build.

jonlives: Yeah I think we need to clarify the language here - new RFCs for every change to non-implemented RFCs would be a nightmare.

rfc000-rfc-process.md
jonlives: We document "replaced" elsewhere in RFC-000, for consistency IMO it makes sense to include it here too

btm: I'm looking for a establishing a backwards link here. Without something like this, it's hard to know that this RFC has a relationship with an older one. For some time cases, it's likely you'll find the newer one but want the older one, like if we just created a newer version of an API, but someone is working on software that used the older version.

I was trying to infer that the metadata line was optional with the <> characters, since the above line infers that you should pick one of the above. Using [] characters instead might make more since to those who are used to that syntax.

tyler-ball:
We could add a top level `replaced` folder where older RFCs are moved to prevent the accepted RFC list from becoming too cluttered.  Thoughts?

jonlives:
@adamhjk this is ready to rock, deciderating needed.

jonlives:
@tyler-ball I like that idea :)

rfc000-rfc-process.md
coderanger: The point is that the example template is supposed to be a copy-paste-go thing, not an exhaustive documentation. This will not be present on the vast majority of RFCs so it should not be in a the template.

jonlives: My issue with that is that RFC-000 pretty much *is* the documentation, and explains further down that RFCs can be "Replaces" but doesn't then include that as a valid field in the metadata. I think something like adding <Replaces> to indicate it's an optional would be enough, but not including it at all seems to me to make things a little more opaque.

coderanger: It should be documented in this RFC, just elsewhere. Make a new section.

adamhjk:
I am the decider, and I approve this message. /cc @chef/rfc-editors

PR #86, opened 2015-01-28
Let chef-client consume github release tarballs in local mode

ranjib:


ranjib:
https://github.com/chef/chef/pull/2827

lamont-granquist:
We already decided to replace chef-solo with local-mode, adding more unique features to solo just makes the parity problem worse: https://github.com/chef/chef-rfc/blob/master/rfc031-replace-solo-with-local-mode.md

ranjib:
@lamont-granquist any idea when chef-solo will be replaced with chef-client -z ? chef 13 ? will it support recipe tarballs (--recipe-url) ? Does it make sense to rephrase the RFC to specify chef-client in local mode instead? Given the chef-solo patch is lot smaller, can we maintain both, while the migration happens?

This RFC's goal was to ease consumption of github releases, so that we dont have to host asset servers separately, I'd be fine with implementing it in local mode, if thats the future.

lamont-granquist:
recipe-url has already been merged to chef-client master on monday.

i'm more in favor of freezing the features of chef-solo completely and any new features would only go into chef-client.  if chef-solo users then have blocking issues which prevent them from moving to local-mode to get those new features then those blocking issues should get raised.  imo, new code should not be going into chef-solo after we've already deprecated it.

ranjib:
@lamont-granquist i have updated the PR and RFC to reflect chef-solo in local mode. I'll nuke the solo related code after this thrursday's rfc meeting (we'll get some more comments, feedback etc)

btm:
Along the lines of a local-mode future, we should call this (filename) something like 'github-cookbook-tarballs'. I'm not sure we need another flag like --github. Can we just turn on github mode for a URI with "github.com" or do the right thing from the directory structure?

Doing the right thing from the directory structure for --recipe-url from GitHub seems like an obviously useful feature to me.

nathenharvey:
Conversation from the Developers' IRC Meeting on 5-March-2015

* **ranjib**:  im still working on btm's comment.  i have to update the rfc text as well as the PR.  to reflect the proposed flags, instead of --github. but i like btm's suggestion
* **holoway**: one thing I want to say, Im against any kind of "freeze" of solo.  it makes our lives easier to get to a place where we can have only local mode filling all the requirements, but until then, I'm not going to condemn that use case to purgatory.  and, just so we're clear, when we get to a place where local mode can replace the guts of solo, we're keeping a solo binary anyway, because we're not going to break the world
* **kallistec**:  the overall behavior requested is totally reasonable, but the way all the tarball stuff happens in solo now is pretty bad from an implementation standpoint. would it be bad to move fetch-code-from-random-places to some other program and have ppl do fetch-my-code && chef-solo ?
* **ranjib**:  Ya, moving the fetch code is a good idea
* **kallistec**: for example, the code download stuff all happens before we get our error handling framework initialized so the UX is bad for failures that happen there
* **mivok**:  kallistec: I like that idea - it would allow for alternate fetch code too such as git pull
* **ranjib**:  How u deploy the fetch code, if its outside solo?
* **holoway**:  it would have to be in chef, chef-fetch or whatever
* **kallistec**:  yeah, it would ship with chef, whether it's a different git repo I don't care so much about
* **holoway**: it shouldn't be a different git repo
* **jonlives**: as long as we don't start needing a Fetchfile :P
* **holoway**: I'm not sure I think it isn't just a feature of the chef-client
* **thom**: that seems like it would break people almost as hard as losing a solo binary
* **holoway**: right?  that's good for us, I'm not sure it's good for solo users
* **kallistec**: thom: moving code fetch stuff outside?
* **thom**: yeah
* **holoway**: yeah
* **mivok**: would an option such as 'chef-solo --fetch-command' work?
* **holoway**: it's already inside.  you can't remove it in this iteration
* **thom**: if we can remove it for local-mode, then great.  but i think solo has to keep it
* **kallistec**: right, don't remove the tarball stuff that's there without at least long deprecation cycle
* **holoway**: I really think btm's suggestion on the PR is right, andw e should just fix the UX stuff, and make it work for solo + local
* **mivok**: you could implement backwards compatibility by making the tarball fetch option expand to --fetch-command='chef-fetch URL' or similar
* **holoway**: but lets take this to the PR
* **kallistec**: right, the old tarball thing could delegate to the new stuff and give us a big improvement, then new features require you to use chef-fetch
* **holoway**: I'm not sure I like the ux of chef-fetch

ranjib:
im closing this one. along with partials in bootstrap templates. The current bootstrapping implementation forces us to swell the erb rendering logic, as well as imposes major UX challenge for new features. But we are likely to need more and more complex bootstrapping options . We need a broader discussion around this, before we start adding individual features.

PR #89, opened 2015-02-04
Allow directories to be managed idempotently

withnale:
## Motivation

    As a SysAdmin,
    I want to be able to manage directories idempotantly,
    so that I know the exact state of a directory over time and no drift occurs.

## Specification

When a directory resource is specified, it should allow for an optional attribute 'purge' to be specified. This would dictate that the only files/directories present within this directory tree are those that have been placed there as part of the chef run.


someara:
Fist, "idempotently" is not a word. "Idempotent" does not have anything to do with what you're after. 
Forget that word.

You want a managed set of files.

It's definitely possible to manage a directory like this per run_context. For example, a provider using use_inline_resources can only see the resources in it's current run_context. I can see this causing a lot of confusion and even wreaking accidental havoc.

lamont-granquist:
Yeah we don't currently have any resources which do deep inspection of the other resources on the resource collection, and the use_inline_resources case really kills this for now.

nvwls:
As the maintainer of the zap cookbook (https://github.com/nvwls/zap) which implements this functionality, use_inline_resources does make this tricky.  We have to have intimate knowledge of various LWRPs to know how certain files are created.

jonlives:
I chatted a little with @withnale about this at the London Chef meetup last night and highlighted a number of the same concerns mentioned above, namely that the introspection needed into the resource collection for this would potentially produce some surprising side effects - hadn't thought of the ```use_inline_resources``` case but that's definitely a good point.

lamont-granquist:
We probably should think about some way of having a global way to view and search all of the resources that chef runs in all its sub-resource-collections, and possibly to do notifies and subscribes off of those, but in some way that doesn't just reintroduce a bunch of crazy.  I'm also very wary of introducing functionality that is going to tie us to compile/converge mode separation even harder than we are now, when there's a near consensus that chef would be much better off without that split.

ranjib:
use_inline_resources do make the problem hard, but i think we have much bigger issue here.
Chef allows resources to be dynamically added via LWRP /HWRP. So, if there exist a sub-resource (like file or template) which writes inside the zap directory, and its parent resource is declared after the zap directory resource, then it will go un-detected.
Deletion of un-declared files in such directory can only be done at the end of the run, post convergence, but that sounds very hacky. I use a confd_handler to do same, but again, this is hacky. I dont think this is possible at all from resource/provider layer, unless you hook some logic

Will it be better to offer a resource that basically adds a handler instead?

lamont-granquist:
We could add another singleton object to the run which would be a Chef::FileTracker or something and have all the file resources directly update a list of files that were touched there.  There might be a larger abstraction here as well -- it might wind up just being a Chef::ResourceLog.

someara:
Also complicating things:

```
file 'arbitrary_name' do
  path '/etc/whatever/foo.conf
end

template 'arbitrary_name' do
  path '/etc/whatever/foo.conf
end
```

adamhjk:
Thanks for taking the time to write up this RFC. We're going to say no, for a couple of reasons:

1) The most common case is the remote_directory case, where you know what you are populating. For that use case we already support purge.

2) For the more complex cases, you need to get very specific about what populates that directory - and for that case, the zap cookbook is more than sufficient.

Feel free to re-open this PR with more comments if you disagree.

btm:
@withnale I feel like we jumped on you a bit with some of our responses coming from our deeper knowledge of Chef internals. I'm sorry about that.

In the case where you don't create a directory with remote_directory, but you do manage everything in it with Chef, the problem is creating a way that one resource knows what all the other resources are going to do and have that be a promise. The system is going to change from one resource running to the next. You can look at some of the discussions about dry-run mode versus why-run mode for a deeper conversation about that.

I think we feel that refactoring resources to make those kinds of promises isn't going to be as sure of a thing in reality, like why-run, so it's not worth the huge amount of engineering time relative to the use case.

Hopefully that makes sense and feels right to you.

lamont-granquist:
Also we should try to make the zap cookbook better and see about getting it the right information on updated files from core chef.

nvwls:
Please open issues against zap as I will against core chef ;-)

PR #91, opened 2015-02-12
Policyfile HTTP Document API

danielsdeleo:
This proposes a standard format for Policyfile lock documents and an HTTP API for storage and retrieval these documents. We have already implemented a "preview" version of this API in both chef-zero and erchef (see here for chef-zero version: https://github.com/chef/chef-zero/pull/111), however that implementation is somewhat simplistic. We intend for it to be replaced with the implementation described here.

I still have some cleanup to do in the rationale section.

Comments welcome.

/cc @chef/delivery @chef/client-core

### DIFF
--- a/new/rfcXXX-policyfile-endpoint.md
+++ b/new/rfcXXX-policyfile-endpoint.md
@@ -69,17 +69,14 @@ A Policyfile lock is a JSON document that has the following structure:
 #### `revision_id` Field (required)
 
 The `revision_id` field is used to distinguish revisions of a Policyfile lock
-document. It is a JSON String. It may contain alphanumeric characters, hyphens,
-underscores, the dot character and the colon character. In regular expression
-form: `/^[\-[:alnum:]_\.\:]+$/`. The revision ID is limited to a maximum of 255
-characters. The Chef Server must reject a Policyfile lock document when the
-revision ID does not meet these requirements.
+document. It is a JSON String containing a hash of the canonicalized content of
+the Policyfile lock, given in hexadecimal form. A fully compliant server MUST
+independently compute the `revision_id` and reject a creation request if the
+submitted `revision_id` is incorrect.
 
-In ChefDK the revision will be calculated by hashing other content in the
-Policyfile lock document and thus will be a hexidecimal number. In order to
-allow alternate implementations to generate revision IDs according to different
-schemes (perhaps timestamps or Ci build numbers), this field will not be
-required to be a hexidecimal string.
+The canonicalized form of the Policyfile lock is not finalized at this time,
+but will be provided in a future update to this document. Note that the format
+will not be based on JSON, as JSON has no canonical form.
 
 #### `name` Field (required)
 
@@ -157,6 +154,10 @@ Example cookbook locks including optional fields:
     },
 ```
 
+Note that a future update to Chef RFC 022 will introduce a standardized format
+for cookbook identifiers based on the content of the cookbook and will
+introduce strict server-side checking of the identifier field.
+
 #### `named_run_lists` Field (Optional)
 
 Named run lists provide a replacement for the override run list feature in Chef
@@ -182,20 +183,11 @@ cannot exceed 255 characters.
   },
 ```
 
-TODO: if named run lists are computed by expanding a run list containing roles,
-they will have different attributes from those associated with the default run
-list (see attributes below). This will need to be accounted for, either by
-making the fields for each named run list a JSON object with `run_list` and
-attributes fields or having a corresponding top-level field for named run
-lists' attributes.
-
 #### Attributes Fields (Optional)
 
 A Policyfile lock may contain default and override attributes. Chef Client will
-treat these attributes the same as it currently treats role attributes. For
-attributes associated with the top-level run list, there are two proposals for
-how these are specified in the Policyfile lock JSON. One proposal is to store
-default and override attributes as separate top-level keys:
+treat these attributes the same as it currently treats role attributes. Default
+and override attributes are stored in separate top-level keys:
 
 ```json
 {
@@ -203,23 +195,6 @@ default and override attributes as separate top-level keys:
   "override_attributes": {}
 }
 ```
-
-The alternative is to store both in a top-level "attributes" key:
-
-```json
-{
-  "attributes":
-  {
-    "default": {},
-    "override": {}
-  }
-}
-```
-
-TODO: The author will do UX testing to determine if either proposal generates
-more readable diffs (particularly in the context of a git repo). If one
-proposal tends to result in more readable diffs, it will be favored.
-
 
 #### Other Optional Fields
 
@@ -422,6 +397,10 @@ Object-level Authorization:
   list of active policies and revisions.
 * An actor must have `Update` permission to set the active revision for a given
   `policy_name` to a specific revision via `POST /policy_groups/:policy_group_name/policies/:policy_name`
+* An actor must have `Update` permission to set the active revision for a given
+  `policy_name` to a specific revision via the promotion API.
+* An actor must have `Update` permission to modify the promotion policy for a
+  given `policy_group`.
 * An actor must have `Read` permission on a specific `policy_group` and `List`
   permission on the `nodes` container to list nodes in a policy group via
   `GET /policy_groups/:policy_group_name/nodes`
@@ -470,7 +449,7 @@ actions individually,
   it exists, or `Create` on the `policy_groups` container to create the desired
   group, AND either `Update` permission on the given `policy_name` if it exists
   or `Create` on the `policies` containter to create the the desired policy if
-  it doesn't exist.
+  it doesn't exist. The same restrictions apply to the promotion API.
 * An actor must have `Read` permission on a sepcific `policy_name` and `List`
   permission on the `policy_groups` container to list the groups associated
   with a given policy at a given revision via `GET /policies/:policy_name/revisions/:revision_id/policy_groups`
@@ -527,12 +506,9 @@ relational database query for use in a web-based UI.
 
 #### Data Cleanup
 
-TODO: this needs to be figured out
-
-* auto-cleanup?
-  * what happens to a group with no nodes?
-  * what happens to a group with no policies?
-  * what happens to a policy lock that is not active anywhere?
+Although the design of Policyfile and related APIs greatly simplifies the task
+of finding and deleting unused cookbook versions (cookbook artifacts in this
+case), no such feature is proposed at this time.
 
 ### Authorization Design Goals and Considerations.
 
--- a/new/rfcXXX-policyfile-endpoint.md
+++ b/new/rfcXXX-policyfile-endpoint.md
@@ -1,0 +1,367 @@
+---
+RFC: unassigned
+Author: Daniel DeLeo <dan@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Policyfile HTTP Resource API
+
+Policyfiles are a new feature of Chef that allow the user to specify a
+`run_list` and exact set of cookbooks that `chef-client` will use to
+converge a node (host) to a desired state. In contrast with the current
+behaviors of Chef Client and Server, Policyfiles provide a way to
+describe and manage configuration code on a whole-node basis, reducing
+Chef Client's use of shared, mutable resources hosted on the server and
+providing operators easier visibility into configuration code and
+changes. For further description of the Policyfile feature, read the dedicated
+[POLICYFILE README.](https://github.com/chef/chef-dk/blob/master/POLICYFILE_README.md)
+
+A preview implementation of the Policyfile feature is currently available; in
+this implementation, Policyfile documents are stored as data bag items.
+Although this implementation works, it does not provide the baseline quality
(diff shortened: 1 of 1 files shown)


josegonzalez:
As a confused chef user, what is the point of the Policyfile and the associated endpoints?

spheromak:
@joedevivo there are a couple other documents that cover the why. https://github.com/chef/chef-dk/blob/master/POLICYFILE_README.md does a good job of explaining the rationale.

danielsdeleo:
@josegonzalez if you still have questions after reading that, please start a thread on the Chef or Chef Dev mailing list so I can improve the documentation.

josegonzalez:
Nope, that explains it all. My only comment on your actual RFC is that it appears some api endpoint sections are empty, but I'm sure that's on purpose :) .

danielsdeleo:
I'll think some more about how best to format that section. I recently started using RAML for this kind of thing and liked it quite a bit, so I tried to emulate RAML's format where you start by naming the URL path and then describe which HTTP methods it supports and what they do. So I originally wrote it like this:

----

#### `/path/to/thing`

##### GET

"describe what GET `/path/to/thing` does

----

 But then I got feedback from a colleague that it was hard to understand that the method and URL path go together, so I repeated the path with the HTTP method.

josegonzalez:
What I've been doing - and it's a bit heavy - is using [API Blueprint](https://apiblueprint.org/) syntax, which also makes it easy to generate digestable api docs. But what you're doing makes sense, so feel free to ignore :)

new/rfcXXX-policyfile-endpoint.md
```diff
@@ -0,0 +1,367 @@
...
+characters. The Chef Server must reject a Policyfile lock document if the name
+does not meet these criteria.
+
+#### `run_list` Field (required)
+
+The `run_list` field gives the ordered list of recipes that Chef client should
+apply to converge a node. Each item MUST be in fully qualified form, that is,
+`recipe[COOKBOOK_NAME::RECIPE_NAME]`. Roles are not valid run list items in the
+Policyfile lock document, and the Chef Server MUST reject run lists that
+contain roles.
```
tyler-ball: It seems like this could be a point of contention for some users.  In the README you said role support was coming - was that idea dropped?  Could you give some explanation here of why roles are not supported and whether they will be in the future?

danielsdeleo: Roles are 'compiled in' to the Policyfile lock document (or will be when this gets supported on the ChefDK side). So your Policyfile.rb run list can contain roles, and can use any of the syntax shortcuts for run list items. Then `chef install` or `chef update` will expand the run list, solve dependencies, etc., and emit only the expanded run list with fully qualified recipe names in the Policyfile lock document.

tyler-ball: Ahhh, yes, I understand - thanks!

new/rfcXXX-policyfile-endpoint.md
```diff
@@ -0,0 +1,485 @@
...
+
+#### `/policies/:policy_name/revisions/:revision_id`
+
+##### `GET /policies/:policy_name/revisions/:revision_id`
+
+Returns the policyfile lock document with the given name and revision ID.
+
+##### `DELETE /policies/:policy_name/revisions/:revision_id`
+
+Deletes the policyfile lock document with the given name and revision ID.
```
danielsdeleo: From @manderson26 via chat, this implies that a malicious/mistaken/drunken actor can delete a policy that's in-use in a group they don't have access to (see authz section). This should probably return 409 if the revision is active in any group. We maybe could offer a force option, which would require that you be authorized to update every group currently using the policy.

new/rfcXXX-policyfile-endpoint.md
```diff
@@ -0,0 +1,485 @@
...
+This implies that a PUT to /policy\_groups/foo/policies/bar has to authorize
+against both policies/\_acl and policy\_groups/\_acl which maybe is weird (?).
+
+#### `/policies/:policy_name/revisions/`
+
+##### `POST /policies/:policy_name/revisions/`
+
+Create a new Policyfile lock document. The POST body MUST be a valid Policyfile
+lock document, as described above. The document can subsequently be retrieved
+at the relative URL `/policies/:policy_name/revisions/:revision_id`
```
danielsdeleo: From @manderson26 via chat, this section needs to clarify that policyfile lock revisions cannot be updated (you have to delete and re-create). Attempting to update should be a 409.

new/rfcXXX-policyfile-endpoint.md
```diff
@@ -0,0 +1,485 @@
...
+
+Returns the policyfile lock document at the revision that is currently
+associated with the given policy group and policy name.
+
+##### `PUT /policy_groups/:policy_group_name/policies/:policy_name`
+
+Sets the active revision of `:policy_name` for the policy group
+`:policy_group_name` to the policyfile lock document in the request body. The
+policy group, policy name, and policyfile lock revision will be created if any
+does not exist.
```
danielsdeleo: via @manderson26 in chat, this section needs to clarify that existing revisions cannot be updated. Since this is the "do what I mean" endpoint, if a given revision already exists, the outcome of the request should be the same as `POST /policy_groups/:policy_group_name/policies/:policy_name` with the `revision_id` in the request body (see next endpoint description).

new/rfcXXX-policyfile-endpoint.md
```diff
@@ -0,0 +1,482 @@
...
+underscores, the dot character and the colon character. In regular expression
+form: `/^[\-[:alnum:]_\.\:]+$/`. The revision ID is limited to a maximum of 255
+characters. The Chef Server must reject a Policyfile lock document when the
+revision ID does not meet these requirements.
+
+In ChefDK the revision will be calculated by hashing other content in the
+Policyfile lock document and thus will be a hexidecimal number. In order to
+allow alternate implementations to generate revision IDs according to different
+schemes (perhaps timestamps or Ci build numbers), this field will not be
+required to be a hexidecimal string.
```
manderson26: I think the revision id should be a strict cryptographic hash of it's canonicalized contents, and that should be enforced at the server. Allow a metadata field for user defined things like timestamps/CI build numbers, but don't conflate the revision_id with the metadata. The lockfile goes to some effort refer to the exact content of the cookbook (the identifier) and moves away from using the version to ; let's carry that all the way through into the lockfile id scheme as well. 

Otherwise you could create a policyfile lock with some collection of cookbooks, and replace it with a different one with a different collection of cookbooks, defeating much of the benefit of moving away from version numbers. This is especially important because the policy groups seem to essentially refer to lockfiles by reference to that id. 

Having the policy groups include policy lockfiles by reference to the id, and allowing the id to be arbitrary creates some possible issues around permissions. Anything short of requiring the policy lockfile revision id to be a strict hash of its contents, and enforcing that at the server level is going to create nasty opportunities. For example dev creates policy id foobar, which passes acceptance, but deletes/recreates it with a policy with the same name but some sneaky property changes in between acceptance and prod deploy.

jonlives:
@danielsdeleo apologies if I'm missing something here - the original https://github.com/chef/chef-dk/blob/master/POLICYFILE_README.md writeup mentioned that role attributes etc will continue to exist as the runlist in <policy>.rb can contain roles...the lockfile however doesn't store any concept of roles.

I'm assuming here based on the docs that the lock file is what the node actually uses to determine its runlist etc - if I'm correct in that, at what point does it know it has any attributes other than those in recipes?

thommay:
@jonlives I think, based on what @danielsdeleo said previously, that the lock file contains expanded roles, such that any roles get resolved to cookbooks and attributes, and those are then included directly in the lock file.

jonlives:
@thommay possibly, but the spec for the lockfile would place attributes under optional fields, if they go in there at all - it's a bit unclear to me exactly how attributes would be treated and the relationship they'd have with the proposed lockfile :)

danielsdeleo:
@jonlives @thommay Thanks for the review. The lack of attributes in the lockfile is indeed an oversight. Somehow I only managed to remember that I forgot to include this when I was away from the computer :wink: In addition to attributes for the default run list, I think there also needs to be a map of attributes that correspond to the named run lists, since you could have different roles in those.

Do you have any thoughts on how this should be structured in the lockfile? The two obvious options are:

```json
{
  "default_attributes": { },
  "override_attributes": {}
}
```
Or have a top-level `attributes` key:

```json
{
  "attributes":
    {
      "default": {},
      "override": {}
    }
}
```

I would tend to favor whichever produces the most readable diffs so users can most easily compare and review changes from one revision of a lock to another, but it will take some research to figure out how git's diff algorithm handles those.

jonlives:
@danielsdeleo I'd lean towards the former purely because that's the same naming scheme as people are already used to from roles, environments etc - don't feel particularly strongly about that though as long as the diffs are readable, as you say :p

new/rfcXXX-policyfile-endpoint.md
danielsdeleo: @manderson26 I'm coming around to thinking that what you're requesting should at least be possible.

My concern is that the current implementation isn't designed to provide strong integrity guarantees. For example, cookbook identifiers are a SHA-1 of the MD5s of the files in the cookbook (code here: https://github.com/chef/chef-dk/blob/b227c377703625ba5da2744d5dbc623c6ba923be/lib/chef-dk/cookbook_profiler/identifiers.rb#L41-L64) which is fine for well-intentioned users, but I would certainly change to a SHA-2 of SHA-2s if I were to make any claims about integrity. Likewise, I would make the revision id a SHA-2 of the canonicalized content if I wanted to make claims about integrity.

Speaking of the canonicalized content, the current implementation of this doesn't account for attributes at all (code: https://github.com/chef/chef-dk/blob/b227c377703625ba5da2744d5dbc623c6ba923be/lib/chef-dk/policyfile_lock.rb#L145-L176), which it certainly will have to when that feature is added. Also I could be doing something dumb in the canonicalization process.

Given that I'd like to be shipping sooner rather than later, that means I can either spec out two behaviors or just ship the feature in violation of the spec and get into compliance later.

Another thought I had (and maybe this is terrible) is that API consumers opt-in to strict crypto checking by adding a `!` to the end of the cookbook identifier or policyfile lock revision ID. Since this character is currently reserved, this would be backwards (but not forwards) compatible, and since the character is part of the ID, it would prevent the delete/recreate threat you described.

danielsdeleo:
Alright, the latest update includes sections about attributes and named run lists.

jonlives:
Awesome thanks @danielsdeleo!

nathenharvey:
As discussed in the Developers' IRC meeting on 5-March-2015, we agreed to go with strict checksumming only, we're okay with shipping early implementations without it, strict checksumming should not prevent shipping.

danielsdeleo:
Hey, how do you all feel about the promotion API? Awesome? Meh? Nah?

jonlives:
Definitely a :+1: on the promotion API - it's a nice way of making this easy to use IMO.

new/rfcXXX-policyfile-endpoint.md
thommay: I think explicitly not providing a force option is probably more healthy here.

new/rfcXXX-policyfile-endpoint.md
```diff
@@ -0,0 +1,558 @@
...
+#### `GET /policy_groups/:policy_group_name`
+
+Returns the policy group configuration for `:policy_group_name`. The document
+will contain the name of the next policy group in the deployment cycle. Exact
+format is TBD, but it could look like:
+
+```json
+{
+  "name": "development",
+  "next_group_name": "qa"
```
thommay: Is there any context in which `next_group_name` might be better as an array?
I sort of started thinking of policy groups as being analogous to environments, in which case you might have a have promotion that looks like: 
```
dev -> qa -> [production, disaster recovery]
```
but actually there's no reason for the set in the array not to be a single policy group. (So I've mostly answered my own question, but I thought I'd throw it out there)

new/rfcXXX-policyfile-endpoint.md
```diff
@@ -0,0 +1,558 @@
...
+  groups.
+
+Object-level Authorization:
+
+* An actor must have `Read` permission on a specific `policy_group` to read the
+  promotion configuration for that group (promotion API only)
+* An actor must have `Read` permission on a specific `policy_group` to read the
+  list of active policies and revisions.
+* An actor must have `Update` permission to set the active revision for a given
+  `policy_name` to a specific revision via `POST /policy_groups/:policy_group_name/policies/:policy_name`
```
thommay: Should also mention promotions.

thommay:
:+1: on the promotions (and generally)

adamhjk:
By the power invested in me by being the decider, I say that when @jonlives is happy with the next round of updates to this RFC from @danielsdeleo - he can merge it.

new/rfcXXX-policyfile-endpoint.md
danielsdeleo: I could see a case for it. For now I'm going to leave it as is and I'll make updates when I get the chance to create this API, since I think I need to start implementing to get a better feel for how these details should work.

danielsdeleo:
Updated.

PR #92, opened 2015-02-18
rfc server api versioning

marcparadise:
RFC for versioning API behaviors in chef-server. 

Note that I'm considering moving section "New Endpoint: /server_api_versions" to its own RFC, since it's implementation isn't required for the core functionality, and it's currently shown as optional.

### DIFF
--- a/new/server-api-versioning.md
+++ b/new/server-api-versioning.md
@@ -1,0 +1,294 @@
+---
+RFC: unassigned
+Author: Marc Paradise <marc@chef.io>
+Status: Draft
+Type: <Standards Track>
+---
+
+# Chef Server API Versioning
+
+The purpose of this RFC is to provide a graceful deprecation mechanism
+for the Chef Server REST API.
+
+As the capabilities of the Chef Server API expand, it becomes necesasry
+to deprecate and desupport old API behaviors.  However, consumers of the
+API such as chef-client may not yet be able to support changed API behaviors.
+The version indicator is used by a client to indicate which level of
+API behavior it can support.
+
+Any given release of the Chef Server will include documentation of the current minimum
+and maximum supported API versions, as well as behavior changes between any newly
+introduced version and the one prior to it. It will also include, if appropriate,
+which version has been deprecated, and which version has been retired.
+
+API Versioning is independent of product versioning.
+
+## Motivation
+
+    As a consumer of the Chef Server API,
+    I want to be able to consume significant changes at my own pace
+    so that they don't break my usage of the API.
+
+    As a user chef-client at a supported version level,
+    I want to ensure that changes to the server API don't affect me until I'm ready to upgrade
+    so that all my nodes continue to converge even as Chef Server API continues to mature.
+
+    As a developer of the Chef Server API,
+    I want to be able to add potentially breaking feabtures without breaking customers on supported versions
+    so that the state of the Server API can continue to advance.
+
+## Specification
+
+The Chef Server API Version (Version from this point forward) is indicated as a whole number,
+starting at zero. Zero indicates the behavior of the API at the time of this RFC's acceptance.
+
+If a client does not indicate a version preference, it MUST receive the behavior
+of the OLDEST supported API version.
+
+If a particular endpoint does not implement versioning due to remaining unchanged, it MUST
+behave in a forward-compatible way through the range of currently supported version. In
+documentation these will be specified as having compatibility with versions 0+.
+
+If a particular endpoint does implement versioning and remains unchanged, it MUST behave in a
+forward-compatible way through the range of currently supported versions. In documentation
+these will be specified as N+, where N is the version at which it was last modified.
+
+For this section, "API" refers to any combination of a method and endpoint, eg `GET /users`,
+`POST /organizations/$org/clients`, et al.
+
+### Request
+
+A client MAY request version compatibility via custom header as
+follows:
+
+`X-Ops-Server-API-Version: $version`
+
+`$version` MUST be one of:
+
+* the desired version number indicated by a whole integer
+* `current` - indicating the behavior of the highest supported version
+* `stable` - indicating the behavior of the lowest supported version
+* `next` - indicating current, plus any experimental/in-development behaviors.
+
+### Response
+
+If a client does include this header and `$version` is...
+
+* ... less than the minimum Version supported by the server product, the
+  server MUST respond with an HTTP 406 with `application/json` body specified below.
+* ... greater than the current maximum Version supported,
+  the server MUST respond with an HTTP 406 and an `application/json` body specified below.
+* ... of a value other than a valid version, `current`, `stable`, or `next`, the server
+  MUST respond with an HTTP 406 and the body specified below.
+
+
+If `$version` is valid or is not specified, the API response will contain the headers:
+
+````
+X-Ops-Server-API-Version: $actual_version
+Vary: X-Ops-Server-API-Version`
+````
+
+`$actual_version` MUST indicate the API version level of the response.  The `Vary` header is included
+in the event it becomes desirable to support caching of specific responses.
+
+#### 406 Response Body
+
+When the server provides a 406 response due to unsupported API version, the response body
+will be of type `application/json` and MUST look as follows:
+
+    { "error" : "invalid-x-ops-server-api-version",
+      "message" : "Specified version $version not supported",
+      "min_api_version" : $x,
+      "max_api_version": $y }
+
+`$version` indicates the requested version. `$x` and `$y` are the current minimum and
+maximum supported versions, respectively.
+
+### Server Implementation
+
+The minimum version MUST NOT be configurable, as it indicates the level supported in the code.
+
+The maximum version MUST be configurable, allowing an installation to explicitly
+limit the introduction of new behaviors.  The configured value MUST NOT exceed the maximum
+supported by the codebase; if such a value is specified, or if a value is not configured,
+the supported maximum MUST be assumed.
+
+The configured value MUST NOT be less than the supported minimum version. If such a value is specified,
+the supported minimum MUST Be assumed.
+
+Minimum and maximum values in effect MUST be logged at server startup.  Further, current API support levels
+MUST be exposed through a new endpoint, `GET /server_api_version` as specified below.
+
+The server MUST resolve and validate the version at the start of each request, and retain it in request state
+for resources to make use of.
+
+The implementation MAY require addition of version support callbacks in erchef webmachine resources.
+
+Given no `X-Ops-Server-API-Version` header or a `null` value of version, it MUST resolve to the
+minimum supported version.
+
+Given a value of version, it MUST be resolved as follows:
+
+* The version label `stable` shall resolve to the minimum supported version.
+* The version label `current` shall resolve to the maximum supported version.
+* The version label `next` shall resolve to the internal identifier `next`.  When no such behavior
+is applicable, the resource MUST treat this as if it were the maximum supported version.
+* A numeric version MUST be validated against current minimum and maximum versions as
+specified under section "Response".
+* Any other value is invalid and the server MUST respond to the client with a 406
+and the associated response body specified above.
+
+### Effects on Unmodified APIs:
+
+APIs that are not modified are not required to support Version.  They are assumed to be
+compatible and unchanged through the current range of min/max supported versions.
+
+### Effects on Modified APIs:
+
+Any API modified as follows MUST use `api_version` to determine which behavior to
+present to the client:
+
+* new required field on input
+* change in type of an output field
+* change in client-facing behavior of the API
+
+Any API modified as follows MAY use `api_version` to determine which behavior to
+present to the client:
+
+* removal of a required input field
+* addition of an output field
+* change in internal behavior of the API
+
+#### New API Implementations
+
+Any new API MAY implement checking of `api_version` to determine its availability.
+
+#### Deprecated API Behaviors
+
+Any API behavior that is to be deprecated MUST have the version for both its deprecation and
+its removal documented in release notes and in API documentation.
+
+After deprecated behavior has been removed, last supported API version MUST be updated in
+accordance with the published documentation.
+
+### Documentation
+
+When an API is modified in a way that requires support of Version, release notes and API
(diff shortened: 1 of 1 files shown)


jonlives:
Since we want to give users a stable baseline, what if 12.now becomes the "stable" API, and any new features returned in the API in 12.x releases result in an X-Ops-API-Features-n header being added with the contents being some indication of the feature.

On the next release, say 13.x, stable becomes 12 stable plus all the added features as default.

That way, clients can default to stable, and also optionally deal with additional features - this would mean that chef-client releases, say, could easily implement new API features while 3rd party clients would still work on the "stable" baseline.

The "stable" baseline changing between major versions is, I think, fair game.

nathenharvey:
Discussion from the Chef Developers' IRC meeting on 5-March-2015

* **mparadise**: So a bit of background on this one - we're in the middle of adding keys rotation, and this will require changing the behavior of a couple of existing endpoints.  We don't want to break various clients of the server, and so RFC-92 was born - providing a way to version server features in a default-conservative means by having the client sending version in a request header, and if no version is specified the oldest supported behavior for the endpoint * **jonlives**: my only concern there is that "oldest supported version" might be deprecated long after the release where new behavior was added.  ie, your api client might break when the version it defaults to changes could be a bit surprising
* **jkeiser**: Yeah, I'm concerned about "stable" -- are we planning to leave "oldest supported version" stable across releases?  If we're not, stable isn't really stable
* **sdelano**: jonlives: since the API has been pretty static, we could take the current implementation as version 0 (or -1) and decrement that when appropriate
* **nlloyds**: Any thoughts around including a list of capabilities rather than relying on version numbers?
* **jonlives**: I like jkeiser's idea of having stable stay stable for an entire release
* **jkeiser**: It feels like we might want to force people to just give a number if they want a particular version, that way they don't get surprised when it changes
* **mparadise**: This isn't specifically tied to product version (12.x, etc) for that reason - those behaviors and rquirements can be defined separately as part of  the  support policy.
* **jkeiser**: Actually, my concern was more that if you upgrade from 12 to 13, clients which have been expecting to continue to work because they specify "stable" will stop working in an unclear way
* **holoway**: yeah - I think in practice, though, you don't really ever want to use the 'stable'
* **thom**: i'm always a little wary of versioned server APIs, because it's incredibly hard to get people to upgrade without simply breaking them and waiting till they fix
* **kallistec**: it'll be very simple for the chef-client side stuff to add a version number and that's how I'd do it
* **holoway**: we're going to hard code the version in the client to a number
* **jkeiser**: If we shouldn't use stable, should we remove it?  Does it serve a purpose?
* **holoway**:  preciesly because of what jonlives said - stable can move, and you won't know the new endpoint, and blammo
* **kallistec**: also, on the client side we could add support for a your-version-will-die-soon header
* **jonlives**: I'm also thinking of third party API clients
* **jkeiser**: oh!  It's a client thing?  I'm OK with that I guess, though, who is supposed to use it?  knife plugins would have issues
* **holoway**: jonlives: me too
* **mparadise**: jkeiser: that makes sense to me in this context.  If there's no baseline for 'stable' then it's not stable.
* **holoway**: jonlives: they'll totally write to the 'stable' api, and then we'll move it, and then they'll break :)
* **mparadise**: Though we could define what 'stable' is for a given server product version?
* **jonlives**: I was more thinking that a given stable will remain supported for an entire release, eg 12.x.
* **mparadise**: ^
* **jonlives**: and the expectation will be that stable will change in 13.x
* **holoway**: I'm good with that.  the issue though is that you really want the integer for 12 ie: if you have client code that will break on 13
* **jkeiser**: OK, then how about stable-12 and stable-13 or something
* **mparadise**: Though again, I'd like to separate that out from what this RFC does - which is talk about the versioning mechanism. I agree that stable should be supported for a release,
* **mparadise**: but think that belongs in the support RFC
* **jonlives**: holoway: good point
* **jkeiser**: Or rename stable to something that doesn't imply that it's, you know, stable
* **sdelano**: if stable stays static across an entire release, then you're never going to get new features.  I don't like the word "stable" for that reason
* **jonlives**: sdelano: i mean stable will be a guaranteed baseline
* **thom**: having a numerical version and a deprecation schedule is probably most easily communicated
* **jonlives**: maybe call it "compat" or something
* **mparadise**: sdelano: stable is really minimum or compatibility. So - s/stable/baseline?
* **jkeiser**: For some definition of "guaranteed"
* **holoway**: well, new endpoints are safe to add
* **jonlives**: having a deprecation schedule is tough, because people won't find it and stuff will break in surprising ways.
* **jkeiser**: Yeah, I like the number.  I can't imagine specifying "stable" and really meaning "I'm OK with this changing across releases"
* **jonlives**: if at least it's tied to a release, that' a bit more expected I'd have thought.  I can live with numbers too
* **holoway**: right - and as maintainers, we can basically say
* **thom**: jonlives: indeed. but otherwise you end up with 30 different API versions and then you die
* **sdelano**: holoway: but for something like key rotation you need to change the users and clients APIs to make it all make sense (because those currently return one key)
* **jkeiser**: Like, if I specify a number, my intent is to specify the number I want, whose responses are the same no matter what.  And I want it to break the day the server stops supporting my number.
* **sdelano**: in this world, that doesn't become stable until 13?
* **holoway**: ahhhh
* **sdelano**: or 12.next?
* **holoway**: does the header apply per-endpoint?
* **thom**: jkeiser: exactly
* **holoway**: or is it universal?
* **mparadise**: It's per-endpoint holoway.  Well - to clarify:  the version value is universal. Each endpoint can implement support for it at current version level as the minimum.

* **jonlives**:  we're going to end up with a semver-api-versioning discussion, i can see it coming :p
* **jkeiser**: I'm just confused as to the use case.  If we can figure that out, we can decide whether "keep," "rename" or "remove" is the right thing with stable, which would let us move on and talk about the rest of this proposal, which is super useful
* **sdelano**: jonlives yay! calling schisamo!!!
* **mparadise**: jkeiser: specifics for our first use case: the clients endpoint needs to return a list of keys instead of a single key.  But existing clients will not like that. *chef-clients/web-ui, etc
* **mmzyk**: I struggle with this RFC b/c I feel it adds a lot of complexity that I’m not convinced is going to give us anything in return. I’m used to entire APIs being versioned, but this version at the endpoint level. It becomes a level of granularity that I think will introduce confusion.
* **mparadise**: So we only return the list if the client says it is ready for it (eg, server api version 1)
* **jkeiser**: I'm curious why we don't design the endpoint for backcompat, but that's a different issue (let's chat after)
* **sdelano**: and what mparadise is saying is that they'll need an easy way to opt into these new features (or get them by default when they upgrade)
* **sdelano**: jkeiser: this conversation is how we add back-compat into the endpoint.  if the server gets no version header, behave the old way, otherwise behave the new way.  we've done this in the past by using the chef-client version header as a key
* **jonlives**: yeah my concern was more around what determines the old way and when it changes
* **jkeiser**: Yeah.  I'll chat with you after about what I mean, it's irrelevant to this discussion, which is useful on its own
* **kallistec**: jkeiser: at some point you really will need to change the document format that you return from GET /clients/foo
* **sdelano**: but there are so many tools outside of chef-client now that that way no longer makes sense
* **holoway**: sdelano: for that use case, wouldn't feature flags be better than versions?
* **kallistec**: otherwise you're just moving the versioning around
* **jkeiser**: kallistec: yep.  I think we have to now, too.  Just not in as jarring a way as implied.  But, irrelevant
* **jonlives**: here's a quick thought
* **ssd7**: holoway: By feature flags, what do you mean?  If they are server-side flags, how would support a mix of client versions?
* **jkeiser** So, what we're saying is "by default, clients will not get Chef Server 13+ behavior, unless they opt in with a flag, and we think the 'stable' version is the right way to specify that default state".  Yes?
* **holoway**: ssd7: I mean rather than saying I support version X, say I support feature Y
* **adamedx**: ah.  like html5 recommendations
* **holoway**: "I can deal with multiple clients"
* **ssd7**: holoway: Yeah, I think a neat alternative would be a header that specified some set of descriptive behavior
* **ssd7**: right
* **sdelano**: holoway: and those become headers?
* **holoway**: yes
* **mparadise**: holoway: wouldn't that get cumbersome as the quantity of changed  and new APIs grows? Too, if it's for a new API, there's no reason for a client to say it's supported - they're already asking for it.
* **holoway**: have N X-Ops-Feature headers
* **ssd7**: X-Features: multi_key; awesome_new_thing; wombats
* **kallistec**: holoway && ssd7 that would basically be moving versioning into content negotiation
* **holoway**: kallistec: precisely
* **jonlives**: I like the idea of stable + feature headers, then next version becomes all those feature headers as stable
jonlives: PR :) (because I like that too)

danielsdeleo:
If we want per-feature stuff, then I'd advocate to move versioning into content negotiation as discussed in the "A custom `Accept` header" section of the rationale. That becomes quite a bit more complex on the client side, as for each request we need to know what kind of resource we're asking for (this is kind of ugly given that you can just do `chef_rest.get('/clients/foo')`) and look up the accept headers we want, whereas with a single version number we probably just assign that in a constant and call it good.

That said, I don't like the alternative idea of a "capabilities endpoint" or at least I would argue that should not be the default way to do versioning. I think the general outline of the way HTTP does content negotiation (make a request optimistically for the thing you really want, if the server can't do it then client can fallback and retry if it wants) is the ideal way to go.

danielsdeleo:
To clarify my last statement. One of the proposals was to have something like `X-Chef-Feature: foo` which enables non-default behavior on the server. This has some downsides:

For one, you will have to maintain a growing list of X-Chef-Features for a long time. For example, if we add `X-Chef-Feature: client-multikey` somewhere in 12.0, and then we want that to become default in 13.0, chef-client 13.0 must still send the header or else it will get an unexpected response from a chef 12.0 server. And when we get to Chef 14 and some hapless user tries to connect to a Chef 12 server which doesn't support client-multikey, we still have to send that header in order to induce a "not supported" response, even though it's been the default for some time. Likewise there are problems with an old client talking to a much newer server; ideally the server should be able to say it doesn't support that old of a client, but that requires we send this header for every request for eternity. Of some concern also is that headers aren't compressed (we don't have client compression on in the HTTP sense of things for most endpoints now, but it would be nice if we did).

Secondly, `X-Chef-Feature` implies you can mix and match the things you want, which greatly increases the number of possible response formats you could have (and would need to test). It's also more complicated when X-Chef-Feature "next" relies on feature "previous".

To fix those things, you probably want to force changes to be linear, at which point you probably want your feature level to have some sorting semantics for developer sanity reasons, which leaves you at having a `X-Chef-Resource-Version` header with a sortable value (integer or whatever) for each resource, at which point you could just fold it in to a custom accept header.

mmzyk:
I've been struggling with this RFC, because my gut reaction is that this adds a lot of complexity that I'm not convinced of the value of. This doesn't strike me as versioning the API, but versioning each of the end points (and at that point, do we really have a versioned API any more, since each endpoint could vary (mostly) independently?).

The Chef server itself can be thought of as nothing more than an API. We've traditionally had an API version, although we haven't made heavy use of it. I think it might be beneficial to step back and look at what value might come out of just implementing a better entire API versioning scheme. Then each time there is a change just bump that. So we get 1.0 for today. 1.1, then 1.2. etc. A client can then just say I need API 1.2 and be done and there is a clear set of endpoint behaviors that correspond to that API version. This strikes me as simpler from an implementation standpoint and from a human understanding endpoint. We'd be able to easily doc what each API version allows, where as with the current proposal I'm struggling to answer that question.

My comments seem to gel with what @danielsdeleo said. However, I'm open to the fact that I might be missing something.

danielsdeleo:
@mmzyk are you saying the server shouldn't support multiple active API versions at the same time?

marcparadise:
@mmzyk The intent isn't per-api versioning, but rather that the server as a whole support a given range of versions.  Current version is 0, and a change in behavior would set that to 1.  In this case, the change of behavior would apply to just one endpoint - but globally the max server version would be 1. In future revisions, we may update behaviors across multiple endpoints concurrently, and that would map to API version 2, and so on. 

As an aside, I deliberately avoided "x.x" style so that there is no confusion with semver behavior - the range of versions defines the behaviors supported by the server, without need for interpration of major/minor changes.   Backwards compatibility isn't implicit in a version number, but explicit based on the version range supported at any given time.

marcparadise:
I updated this RFC as follows based on feedback received here and via other venues:

* Changed and expanded wording to make more clear that this is not
  per-endpoint versioning, but rather is global range across the server,
  with each version defining a set of behaviors across all APIs.
* Expanded on the notion that while product version must specify
  supported and deprecated Server API Versions, the relationship was not reciprocal - Server API Version can be incremented independently of product version, although 
no versions can be retired or deprecated except through  appropriate product releases according to 
accepted support lifecycle documentation. 
* Added example request-responses to demonstrate how version fits in,
  the responses expected for a given version in a changed endpoint, and how the 
 established minimum and maximum supported versions will affect this over time.
* Clarified that there is not a 1:1 relation between a new version
  number and a single endpoint change - a new version number can include
  many new behaviors.
* Removed special version labels based on discussion in the IRC
  developers meeting - they add potential sources of confusion more than
  they simplify usage.

/ping those who had feedback but haven't weighed in here:  @adamhjk @stevendanna  @sdelano @jkeiser @thommay (Sorry, I don't know nlloyds's github name :(  )

mmzyk:
@marcparadise nlloyd's is @smith. Now you know.

new/server-api-versioning.md
```diff
@@ -0,0 +1,371 @@
...
+a given product release.
+
+Arbitrary examples with no significance:
+
+* Chef Server 12.0 - API Versions 0-3
+* Chef Server 12.1 - API Versions 0-4
+* Chef Server 13.0 - API Versions 3-15.
+
+This means that if you are using Chef Server 12.1, any supported client will behave
+as it did with Chef Server 12.1.  However when you upgrade Chef Server 13.0 you will need to
```
stevendanna: "as it did with Chef Server 12.1" -> "as it did with Chef Server 12.0"

adamhjk:
I am Adam Jacob, and I am the Decider, and because I can't let my people down, I decide we should merge this RFC.

The one caveat is that we aren't actually sure if we've covered all our bases here - but the best way to find out is to implement and then revise, rather than talk about the PR. Lets do it, and if we ahve to make changes, we send follow-up PRs to the merged RFC to reflect reality.

/cc @chef/rfc-editors

PR #93, opened 2015-02-21
On Demand Cookbook Libraries RFC

jkeiser:
**Breaks Backwards Compatibility? (Workstation):** No
**Breaks Backwards Compatibility? (Cookbooks):** No
**User Benefit:** Med

### DIFF
--- a/new/on-demand-cookbook-libraries.md
+++ b/new/on-demand-cookbook-libraries.md
@@ -40,9 +40,9 @@ eager_load_libraries false
 
 If `eager_load_libraries` is false, Chef Client would instead append it to the ruby load path.  This would occur in the phase where the `libraries` files of a cookbook otherwise be automatically required in alphabetical order.  All recipes, libraries, and other Ruby code running in the Chef Client can then use `require 'filename'`` and it will load `cookbooks/mycookbook/libraries/filename` if present.
 
-If a file named `default.rb` exists in `libraries`, it will be required automatically.  No other files will be automatically loaded if `eager_load_libraries` is false.
+If `eager_load_libraries` contains a list of files or globs, the library path is appended to the Ruby load path, and the listed files are loaded during the library load phase.
 
-If `eager_load_libraries` is true or not specified, Chef Client loads the top level of files in alphabetical order, as before.
+If `eager_load_libraries` is `true` or not specified, Chef Client loads the top level of files in alphabetical order, as before.
 
 ## Rationale
 
--- a/new/on-demand-cookbook-libraries.md
+++ b/new/on-demand-cookbook-libraries.md
@@ -35,14 +35,14 @@ When a user creates a cookbook, they can add the following directive to the meta
 # metadata.rb
 name "mylibrary"
 version "0.0.1"
-load_libraries_on_demand true
+eager_load_libraries false
 ```
 
-If `load_libraries_on_demand` is true, Chef Client would instead append it to the ruby load path.  This would occur in the phase where the `libraries` files of a cookbook otherwise be automatically required in alphabetical order.  All recipes, libraries, and other Ruby code running in the Chef Client can then use `require 'filename'`` and it will load `cookbooks/mycookbook/libraries/filename` if present.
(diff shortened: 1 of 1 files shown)


ranjib:
:+1:  a shorter an sweeter method name will be awesome :-)

jkeiser:
@ranjib totally cool with other names than `load_libraries_on_demand`.  But not sure what :)  I'd like to pick something that is specific enough to say what it does as well as simple enough for someone who isn't familiar with Ruby load paths and requires to understand basically what's going on.

coderanger:
`on_demand` implies some kind of autoload-style behavior, such as how Rails loads library code. That should be avoided at all costs as a system-level feature. It has led to tons of issues in Rails.

jkeiser:
There is no lazy loading in this proposal, thankfully.

coderanger:
Should probably rename it then :-) Or at least make it very clear that "on-demand" != "autoload".

ranjib:
```ruby
add_library true
```
?

jeremyolliver:
Documentation of this feature should probably point out that files intended for use in this manner with `require`, will need to be named to avoid collisions with other cookbook library files, and even trickier to avoid collisions with ruby stdlib, and gem names in the chef omnibus install correct?

ranjib:
@jeremyolliver your concerns are right, but this risk portfolio is same as any other gem or library, so im unsure if we have to explicitly call this out

coderanger:
Having mulled it, I would probably call this feature "eager loading". So `eager_load_libraries false` to disable it. That also nicely sets the stage to make it the default in Chef 13 (a man can dream).

jonlives:
I like this idea, again though I'm not sure what the best name for it is - not being familiar with the rails context of the term I actually quite liked `load_libraries_on_demand`.

jkeiser:
`eager_load_libraries` is fine with me, and fits with the idea that eagerly loading stuff is the unusual action (which IMO it is).

coderanger:
:+1: Also as a corollary between this and the chef_version field being added, I think it shows we should reconsider how to handle unknown metadata on the server side, maybe make it more of a non-validating hash. That's out of scope for this RFC but will probably factor in to the implementation :-)

thommay:
:+1:

new/on-demand-cookbook-libraries.md
```diff
@@ -0,0 +1,58 @@
...
+```ruby
+# metadata.rb
+name "mylibrary"
+version "0.0.1"
+eager_load_libraries false
+```
+
+If `eager_load_libraries` is false, Chef Client would instead append it to the ruby load path.  This would occur in the phase where the `libraries` files of a cookbook otherwise be automatically required in alphabetical order.  All recipes, libraries, and other Ruby code running in the Chef Client can then use `require 'filename'`` and it will load `cookbooks/mycookbook/libraries/filename` if present.
+
+If a file named `default.rb` exists in `libraries`, it will be required automatically.  No other files will be automatically loaded if `eager_load_libraries` is false.
```
lamont-granquist: Not wild about the magic default.rb here.  I see where you want to bootstrap the library loading with a hook, but maybe this should go into the metadata parameter?

`eager_load_libraries [ 'default.rb' ]`

Allow the filenames to contain globs and then the user has control and its explicit with no magic.

jkeiser: Sure.  With eager_load_libraries, that actually makes even more sense.  I'll change it.

lamont-granquist:
:+1:

jonlives:
:+1:

adamhjk:
I am the decider, and I approve this message. GoGo Gadget @chef/rfc-editors

martinisoft:
:+1:

jtimberman:
Per IRC, known tools that might have their own custom metadata loading that will need to be aware of this new metadata field:

* [stove](https://github.com/sethvargo/stove/blob/master/lib/stove/cookbook/metadata.rb)
* [ridley](https://github.com/reset/ridley/blob/master/lib/ridley/chef/cookbook/metadata.rb) (via berkshelf)

I'm not aware of other tools offhand that implement their own metadata parsing class. The librarian-chef project uses [Chef::Cookbook::Metadata](https://github.com/applicationsonline/librarian-chef/blob/603de858cef9e4e4a25f5013f28d43e218b942b7/lib/librarian/chef/manifest_reader.rb#L29).

PR #96, opened 2015-03-04
Restrict search to current environment by default

thommay:
Restrict search to node's current environment (cc @alexpop )

### DIFF
--- a/environment-restricted-search.md
+++ b/environment-restricted-search.md
@@ -1,0 +1,44 @@
+---
+RFC: unassigned
+Author: Thom May <tmay@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Environment Specific Search
+
+Chef searches in recipes should be restricted to the node's current environment by default.
+
+## Motivation
+
+    As a sys admin,
+    I want to ensure that my searches are accurate,
+    so that my applications aren't confused.
+
+## Specification
+
+By default, searches return all possible results in an organisation.
+This RFC proposes to change that so that searches are restricted to the
+environment that the searching node is in.
+For backwards compatability, this feature would be enabled by a config
+option in Chef 12, becoming the default in a future version of Chef.
+
+## Rationale
+
+Essentially every search I have ever written contains:
+
+```
+"chef_environment:#{node.chef_environment} AND …"
+```
+
+This is wasteful and also introduces the possibility of nasty bugs
+should one forget, probably introducing site reliability issues. Doing
+this by default is both more delightful and more safe.
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.
+


alexpop:
I support this change.

The options to limit the search could go even further.

For organizations where node search is not used, it would be more secure to limit the scope of the search to the node level. I would like to see this as the default option on a future major Chef server release.

someara:
It should probably also use partial search by default

ranjib:
this will break backward compatibility.
Not everyone uses chef environment same way. Lot of users uses one chef server per environment, and chef_environment inside one chef server for cluster/service (like service A = db + web+lb).
This is also somewhat strange as now for listing all servers one has to explicitly specify chef_environment:*. 

I have used, and am still using *:* style searches to create full mesh networks (like iptables, ipsec etc) using chef. Its the only decent way to create security group like ipchains in non-aws providers.

jonlives:
This also breaks searches in the workflow where a cookbook might be tested in an unconstrained environment versus the production environment with rigid constraints....in our case, because we *usually* have everything in one environment we *don't* usually specify the environment in search queries, so this change would result in any node not in the prod environment not being returned.

I'm very much against this being an "enabled-by-default" change as it has the potential to alter the results of a lot of search queries in subtle errorless ways.

jonlives:
I think essentially if you come down on the side of @thommay's case where environments *are* usually specified in search queries, this change makes a lot of sense. If you are in our case where environments are *not* usually specified, this change is scary and would require a lot of recipe fixing.

lamont-granquist:
This should probably be implemented as an API extension that makes it easy to do searches restricted by the environment by default.   Doing this by default will create a flag day where everyone has to audit all their recipes and include some magic button on all their search results which are cross environment by design and that the downside of that change seems to greatly outweigh the benefit in this case.  Some kind of chef-sugar-ish extension would better.

lamont-granquist:
I've also considered doing node-local storage which was unindexed and by default restricted access to the node in question and admins which would actually solve @alexpop's issue, but I think the search ACL filtering that is in progress will also likely address that.

thommay:
OK, so consensus seems to be that 
* there's no sensible default for this
* we should make it easy for an author to write searches that are environment specific

alexpop:
I would like to see a Chef server setting or ACL to limit the search results to the current environment.
From a "secure by default" perspective, this should become the default in a future Chef server major release. ACLs or an organization level setting will be provided in order to enable cross environment searches.

jonlives:
@alexpop again that's only viable when you don't already use multiple environments in one infrastructure / org etc. I'm categorically against implementing this as an enabled-by-default change which will cause cookbook behaviour to change silently when searches all of a sudden start returning different results.

jonlives:
@alexpop to clarify slightly, I understand the need to enable the workflow you're describing, I just think we need to be very careful about how we expose it, and when it gets enabled and in what form. This has the potential to provide a somewhat hostile user experience.

alexpop:
I totally agree @jonlives. Such a change needs to be carefully implemented and properly communicated.

sc0ttruss:
Folks, I am on board with this whole change.  A lot of folks have been using environments for a very long time, and  having to create an organisation for Production just so that the dev/test searches cannot see Production, is a pretty painful way of separating concerns.   
Note: Splunk guys got this right by having separation of concerns at the index level within splunk.

lamont-granquist:
For the security issues, @thommay's work on correctly ACL'ing search results is the way to solve that problem.  Then if none of your nodes in the int environment have permissions to retrieve your nodes in the prod environment, the search results for your int environment will be pruned.  However since its ACL based it is more flexible than imposing a flat dumb restriction that nodes can only search for other nodes in their own environment.  I may not want that since I may have infrastructure in datacenters which supports multiple environments (particularly preprod int/dev/test/qa/whatever infrastructure) which is served by single monitoring servers where the nagios servers need to be able to issue search requests across different preprod orgs that they may not be in.  Imposing a requirement that I have one nagios server per environment and cannot have one for multiple environments may increase my management burden substantially.  With ACLs it is much easier to implement policies like this, and that is probably just one example base on prior experience.   For people who are using fine-grained environments to do deployments, restricting search results to the environments they are using will just break their ability to use search and that's a pattern that we'd like to support as well.

So, you can't impose search restriction by environment by default, and its not really up for debate, it would break too many people and too many reasonable use cases.  It'd be nice to see some sugar around the search API to make it easy to issue searches that are restricted by environment just to reduce typing and reduce errors.  It'd also be nice to see some easy way of configuring search ACLs to start with a baseline where search results were restricted by environment.  That is clearly where most people will start with their ACLs, so it should be easy to setup.

miketheman:
I'd much rather see a search helper that would change the behavior than redefining the default.

adamhjk:
Ok, we are -1 on this RFC. The short story is that changing this as the default has pretty drastic back-compat implications.

However - we think there is a bunch of really healthy, good work that can be done to improve the semantics of search. A short list: safe results, partial by default, environment restricted, etc. @ranjib has a good tack here, which is to create a 'query' API that implements these best practices, and migrate to it over time.

Issue #30, opened 2014-07-24
[Chef 12] Create a `package_repo` resource


jtimberman:
Potential providers:

* yum
* apt
* homebrew (taps)
* IPS 
* rubygems

This was brought up on issue #28. Moving it here for additional discussion.

#### Create a package_repo resource

* **Breaks Backwards Compatibility? (Workstation):** No
* **Breaks Backwards Compatibility? (Cookbooks):**  No
* **User Benefit:** High
  * Short description of the benefit: Having package repository resource/provider available in core Chef would be a huge benefit to remove the dependency on the `yum`, `apt`, and other cookbooks.
* **Impact:** Low
  * Short description of the impact: The only impact is that users who wish to use the new resource would opt into using it by writing the new resource in their recipes. Users can continue using the existing cookbooks until they're ready to migrate.

thommay:
Closing in favour of the pull request version #98

PR #98, opened 2015-03-04
Create a package_repo resource

thommay:
This is a conversion of #30 into RFC form.

### DIFF
--- a/package_repo_resource.md
+++ b/package_repo_resource.md
@@ -1,0 +1,71 @@
+---
+RFC: unassigned
+Author: Thom May <tmay@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Add a resource to manage package repositories
+
+Provide a core resource to allow a user to add a package repository to
+their system.
+
+## Motivation
+
+    As a cookbook author,
+    I want to write cookbooks that rely only on core resources,
+    so that I can minimise my dependencies.
+
+## Specification
+
+Create a new `package_repo` resource, similar to 
+```
+package_repo "foo" do
+  description "A descriptive string"
+  base_url "https://some/url"
+  gpgkey "https://some/other/url"
+end
+```
+Providers are free to add additional platform specific attributes, but
+must attempt to standardize where practical.
+
+## Rationale
+
+Currently, to add a package repository in a cookbook, one must depend on
+the relevant cookbook for that repository type. Supporting multiple
+package types greatly increases the dependency footprint of a cookbook,
+whilst simultaneously requiring the author learn the details of many
+disparate resources.
+
+ie, the `yum` cookbook provides the `yum_repository` resource
+```
+yum_repository 'zenoss' do
+  description "Zenoss Stable repo"
+  baseurl "http://dev.zenoss.com/yum/stable/"
+  gpgkey 'http://dev.zenoss.com/yum/RPM-GPG-KEY-zenoss'
+  action :create
+end
+```
+while the `apt` cookbook provides the following
+```
+apt_repository 'cloudera' do
+  uri          'http://archive.cloudera.com/cdh4/ubuntu/precise/amd64/cdh'
+  arch         'amd64'
+  distribution 'precise-cdh4'
+  components   ['contrib']
+  key          'http://archive.cloudera.com/debian/archive.key'
+end
+```
+
+Providing a single `package_repo` resource with standardised
+attribute naming would decrease the cognitive burden of writing cross
+platform cookbooks
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.
+
+


thommay:
@jtimberman because i love you I have written you an RFC

jonlives:
I wish I could :+1: this more than once. Oh wait, I can!

:+1: 
:+1: 
:+1:

danielsdeleo:
Absolutely :+1: on the general idea. Have some questions about a few details.

package_repo_resource.md
```diff
@@ -0,0 +1,71 @@
...
+    so that I can minimise my dependencies.
+
+## Specification
+
+Create a new `package_repo` resource, similar to 
+```
+package_repo "foo" do
+  description "A descriptive string"
+  base_url "https://some/url"
+  gpgkey "https://some/other/url"
```
danielsdeleo: probably everything on Linux will use GPG, but are there any other cases which use non-gpg signatures (anyone know how Windows nu-get works)? What I'm getting at is, should this just be `key` ?

package_repo_resource.md
```diff
@@ -0,0 +1,71 @@
...
+    I want to write cookbooks that rely only on core resources,
+    so that I can minimise my dependencies.
+
+## Specification
+
+Create a new `package_repo` resource, similar to 
+```
+package_repo "foo" do
+  description "A descriptive string"
+  base_url "https://some/url"
```
danielsdeleo: What makes this "base_url" instead of just "url" or the pedant's choice "uri"? Does that reasoning apply everywhere?

package_repo_resource.md
someara: Are the semantics around "package_repo" congruent across all the different
types?

I can't imagine they are.... this smells like a leaky abstraction to me.

-s

danielsdeleo: @someara are you referring to the feature in general? I'd guess that there's a fair bit of difference between different platforms, but generally you get a list of available packages over the network and then you can do `package-manager install $thing` and it works. Even if the usage will be quite a bit different between different platforms, this is such a common config management task that we should have it in core (IMO).

someara: I'm all for moving yum_repository and apt_repository into core, but a
generic "package_repo" abstraction feels wrong to me.

The "base_url vs uri" question that came up in the first 5 minutes here is
a perfect example to why.

danielsdeleo: If that's our worst problem, then I think we have a pretty good chance of succeeding. Personally I don't think it would be any worse than packages or services.

package_repo_resource.md
jtimberman: Aliases aren't particularly tricky, right?

image

jtimberman:
Thanks @thommay :cake:

someara:
I'd like to see this attempted in a cookbook first. This is abstraction for
the sake of it.

system users are similar cross platform.... the all have standard posix
bits.
providers abstract useradd vs adduser vs whatever.

system services facilities have a service_name and a small amount of states
to put them in.

network packages_repositories vary WILDLY in their capabilities and
interfaces.

apt_repository has ~10 properties
https://github.com/opscode-cookbooks/apt/blob/master/resources/repository.rb#L30-L40


yum_repository has ~36 properties
https://github.com/chef-cookbooks/yum/blob/master/resources/repository.rb#L26-L62

ips_repo has ~5
https://github.com/clintoncwolfe/ips/blob/master/resources/repo.rb

pkgin? aix? chocolatey? other windows?

-s

package_repo_resource.md
lamont-granquist: Yeah, but service is horrible, and package isn't much better...   =)

I do think the implementation should either copy how the package resources + providers work, or else it should be one resource and one provider which uses composition to get multiple different behaviors in the provider.

lamont-granquist:
It does look like a severely leaky abstraction to me.  Without seeing the code I'd say it should be written as separate resources and providers for each repo type.  We can then alias package_repo up to yum_repository and apt_repository via `provides` lines in the resource for the appropriate platform.  Some inheritance of base classes could be used where its truly appropriate to DRY up the code, but our codebase is littered with issues caused by trying to treat RHEL and Solaris abstractions like Ubuntu abstractions.  I'm very confident that starting with that design will fail to be delightful in the end.

package_repo_resource.md
danielsdeleo: Sure. At this point, all that's been proposed is to have a common resource API for the things that are common. If we implement it such that calling `package_repo` in the recipe returns an `AptRepo` or `YumRepo` resource, we still achieve what this RFC proposes.

lamont-granquist: Yeah, but its written as "create a single packge_repo resource" which certainly sounds like one big leaky abstraction that won't work well.   If the RFC was stated as to pull yum_repository, apt_repository and friends into core chef and keep as much of their API consistent as possible so that a package_repo alias made sense I'd be less worried about it...

coderanger:
In addition to being a leaky abstraction in terms of varying capabilities, there is no such thing as a multi-system repo. The code is very often going to need different values of URI, GPG key, etc for each package system and if you are already doing `if ubuntu? ... elsif redhat? ... end` using the existing separate resources is just as easy and a lot more clear.

thommay:
Let's put this on hold pending how #101 turns out. I still think there's value in getting the abstraction right, but let's do it bottom up.

lamont-granquist:
Totally agreed.

btm:
@thommay as the person who opened this, do you want to close the PR for now so we're clear it's not being considered right now?

PR #100, opened 2015-03-05
Lcg/node state separation

lamont-granquist:


lamont-granquist:
I didn't mention auth on this at all.  The new endpoints should ideally just check perms on `/nodes/:name` and use those rather than introducing any kind of new primitives to manage and keep in sync, but I'm not familiar enough with erchef's auth model to know if that makes sense or if we do that anywhere else.

coderanger:
Given the scale of this change, that Rationale section needs to be way way bigger.

thommay:
I like this change in concept, and will properly review the details in a bit.
@alexpop and I were talking in the office yesterday about wanting to ensure that nodes couldn't alter their run_list or which environment they're in, and so this is wonderfully timed.

lamont-granquist:
There's perfectly rationale reasons, though, for having nodes modifying their configuration data, particularly their run_lists.  We can't prevent nodes from doing that, and user's write cookbooks that use that feature.  The node itself is the best source of truth about itself and programmatic inspection of the node's environment by the node itself will generally be better than enforced policy from an overseer (which I think falls out of Promise Theory, or is probably derivable from it).

coderanger:
If this was done, nodes editing their definition vs. state could probably be handled via ACLs.

alexpop:
Nodes changing the run_list or environment is a valid use-case, but it shouldn’t be allowed by default for "secure by default" reasons. This will reduce the attack surface for Chef environments.
For example, by having access to a load balancer node in the 'staging' environment, I can rename client.rb to knife.rb and change the node to become a DB in the 'production' environment.

Providing organization settings or ACLs to allow the current behavior will ensure backwards compatibility.

lamont-granquist:
Feels like you're breaking the whole infrastructure as code model and walking back to infrastructure as configuration.  Its definitely a breaking change and can't be introduced until Chef 13.

I think it might be possible that it could be implemented as a container around the configuration sub-objects of nodes.  However, this proposal doesn't implement that separation strictly enough because there is still /nodes/:name which the node has write access to and which supports writing the configuration data there.  We'd have to drop the PUT endpoint there (which would completely break old knife versions talking to new chef servers).

coderanger:
The specific reasons I want to block nodes changing their own run list is that I would like to use role/recipe assignments to control secrets access and similar stuff and right now that's not viable.

nvwls:
Besides the security concerns, there are "race" conditions as well.  If I flip a bunch of nodes from one `chef_environment` to another and a node happens to be running chef-client, the `node.save` at the end of the run will "reset" `chef_environment` back to the original value.

We currently monkey patch `node.save` to get around this.  I would **really** like to remove that code.

lamont-granquist:
So @nvwls this RFC addresses exactly that use case.  The common case will be that the environment, run_list, etc weren't 'dirtied' by the chef-client run and they will not be saved with the node.save, so it eliminates the race condition in the common case.

We can include the different ACLs, I'm going to have to leave the implementation TBD though until I research it.

lamont-granquist:
```
holoway: ok, so for node state separation
[09:38am] holoway: we have to do this
[09:38am] thom: it feels like 100 needs more fleshing out
[09:38am] holoway: but this isn't quite ready to merge
[09:38am] lamont_oc: well do we want to do the separate ACL'ing or not?
[09:38am] holoway: I feel like this was my largest mistake
[09:38am] btm: kallistec: and I promise to work on it within 24 hours.
[09:38am] kallistec: btm: ok
[09:38am] holoway: we need to separate desired from current state
[09:38am] thom: holoway: openid is swept under the rug now?
[09:39am] holoway: I think we need to address acls
[09:39am] holoway: thom: we fixed that mistake!
[09:39am] lamont_oc: i could use some more feedback on if desired and current are separate ACLs
[09:39am] • jtimberman chuckles
[09:39am] holoway: (and I was ahead of mine time - the whole world uses token auth now, so fuck the haters)
[09:39am] thom: we really need to adress ACLs here
[09:39am] thom: lamont_oc: i think they aare
[09:39am] holoway: (I was a genius)
[09:39am] holoway: (take that)
[09:39am] someara: how about we just make node attributes immutable for the duration of the chef-client run and call it a day
[09:39am] jonlives: lamont_oc: one question on that one - it says node.save saves to the runtime data...I don't see anything saying when then configuration data will be updated, if at all.
[09:39am] holoway: someara: that is not a solution
[09:39am] holoway: someara: it's fraught with edge cases
[09:40am] chip-: someara: doesn't fly when you've got 125 CCRs/sec reporting in
[09:40am] lamont_oc: node.save would update the configuration data if it was dirtied
[09:40am] thom: doesn't immutability solve everything though?
[09:40am] holoway: putting a big lock around it just leads to the problems that locks have in distributed systems
[09:40am] ranjib left the chat room. (Ping timeout: 256 seconds)
[09:40am] miketheman: what if the run never completes, and the node is in immutable state forever?
[09:40am] holoway: ask cheeseplus how he feels about that
[09:40am] jonlives: lamont_oc: ah - I think that needs clarified a little, the languages used in the node save section was a little unclear on that.
[09:40am] chip-: I feel strongly that these should be separate ACLs. It's a very common request from THE ENTERPRISE that run_lists and environments be immutable.
[09:41am] holoway: so - I think thom is right, the ACLs have to be separate
[09:41am] btm: holoway: +10
[09:41am] jtimberman: chip-: +LONG_INT
[09:41am] thom: chip-: yar
[09:41am] • someara retracts his comment.. E_DISTRIBUTED_SYTEMS_ARE_HARD
[09:41am] holoway: with the 12 implementation defaulting to open
[09:41am] holoway: and 13 defaulting to closed
[09:41am] lamont_oc: hmmmm, okay maybe when i changed i from explicitly PUT'ing to /nodes/:name/configuration and /runtime to just using /nodes/:name that became less obvious what i was doing...
[09:41am] kallistec: I think in that case though you need to make the configuration part more granular
[09:41am] miketheman: lamont_oc: node.save in a recipe, or node.save at the end of a run? Won't some attributes always be dirtied?
[09:42am] jonlives: lamont_oc: I understand "If it is detected that the configuration data has been dirtied then node.save will instead PUT to /nodes/:name. " once you explained it, but yeah I think it needs clarifying a little in the RFC
[09:42am] holoway: I think part of ti is /configuriation and /runtime isn't quite right languatge
[09:42am] holoway: it's really /desired and /current
[09:42am] kallistec: like tags and normal attributes you probably need to be able to modify
[09:42am] lamont_oc: the runtime attributes are always dirtied and always written, the configuration attributes usually aren't
[09:42am] jonlives: could we use "desired" and "current"?
[09:42am] holoway: to put it another way - you want to let the 'current' state be mutable and is often dirtied
[09:42am] jonlives: oh holoway just said that. +1 to holoway's suggestion 
[09:42am] holoway: the 'desired' state is often not mutable by the node
[09:42am] chip-: This also opens the gateway to having node objects in the repo, and modifying/uploading those as files without worrying about wiping current state, which would be 
[09:42am] lamont_oc: yeah i copied that from john k's original gist, i can use desired and current instead
[09:42am] ssd7: I'd like to see the client-side accessors for this data split spelled out as well
[09:43am] miketheman: holoway: disagree - nodes may make changes to their own run lists, changing the desired state for next run
[09:43am] thom: +1 to desired and current
[09:43am] martinisoft: Ooo, I like that chip-
[09:43am] holoway: miketheman: which is fine, they can do that
[09:43am] miketheman: (and such)
[09:43am] holoway: you have ACLs on Desired and Current separately
[09:43am] holoway: implementation 1 is back-compat, with the default acl on desired being open
[09:43am] martinisoft: I like desired and current as well
[09:44am] holoway: but you move to desired being closed to the node itself
[09:44am] holoway: by default
[09:44am] thom: yes
[09:44am] jonlives: I think given this is a pretty large change, expanding the rationale and motivation sections would be a good idea
[09:44am] holoway: it's mostly a plumbing change - I bet you can land the entire feature without breaking back-compat
[09:44am] holoway: jonlives: yes
[09:44am] someara: I've never understood how a node attribute can simultaneously be input and output of a chef-client run
[09:44am] jonlives: ie, for somebody reading the RFC who wasn't familiar with this issue
[09:45am] miketheman: I'll also state like current debugging attribute level overrides is complex, this should have very clear debugging/troubleshooting to help the operator figure out what exactly they are getting.
[09:45am] someara: attributes in general are probably the biggest source of confusion and pain in chef
[09:45am] someara: we should be very thoughtful about changes to them
[09:45am] holoway: we have a spearate RFC about that
[09:45am] holoway: (attributes)
[09:45am] thom: i think also we're going to need to do some work on ACL management, because there really isn't a good story there right now
[09:45am] holoway: this isn't that
[09:45am] holoway: this is simply saying setting desired state is separate from reporting current state
[09:46am] nathenharvey: 15 minute warning
[09:46am] holoway: so - lamont_oc, please copy-pasta this conversation in to the RFC comments
[09:46am] jonlives: this wins the "why wasn't it already this way" award for 2015, IMO.
[09:46am] holoway: and make with the typing
[09:46am] lamont_oc: cool
[09:46am] holoway: jonlives: it wins it from 2008
[09:46am] jonlives: heh
```

stevendanna:
The discussion in IRC moved rather quickly so I wasn't able to express this well:

I think we need to think about how this feature interacts with: #77 

The IRC led me to believe that the consensus is moving towards:

- renaming the endpoints to /current and /desired
- this being very separate from #77 

I am having a hard time imaging how that won't lead to a lot of confusion down the road.

danielsdeleo:
I think normal attributes and tags (which currently are just normal attributes) should be a different thing. It's totally reasonable to do something like adding a tag in a recipe that you later search on, for example so your LB search only finds application servers that are fully configured, but you still might want to restrict run list updates. So from an authz perspective you need to have different ACLs on those, hence they should be separate. This overlaps a little bit with some of the things we say we wanted in #9 around making normal attributes a separate thing entirely. Of course we need to keep the old APIs working as a facade around the new stuff. I would like to at least talk about whether any of the "normal attribute" replacement stuff should be bundled in, since it would be kinda sucky to thrash the api multiple times and then provide two or more layers of back-compaq facades.

stevendanna:
For the record, I am a big proponent of both this and #77, I just think that we should think about about them together.

danielsdeleo:
As far as this and #77, I think in #77 we'll make access to normal attrs, merged default/override attrs, and ohai data go though separate ruby methods. Whatever happens in chef-client's internal representation and the server's representation can be papered over. The conceptual goals are similar (make desired state/"aspirational data" look different than "facts about the system or a thing that happened in the past") but the implementation parts are all independent.

lamont-granquist:
One question that this conversation made me think of:

With the ACL changes that people want around the run_list, that would also block the node from writing to normal attributes.  Granted, breaking normal attributes completely might be a feature and not a bug.  But there might be two different things here.  People don't want nodes to be able to update their own run_list, environment and presumably tags, since those wind up in search requests on other nodes and there's a risk that a hostile attacker might change those attrs on a node and "Dependency Inject" the node into another nodes config (redirecting LB traffic, etc).   I don't see that risk from normal attributes.

And with #77 we definitely want to push normal attributes off into its own API land so that its not part of the common API path and people don't use it unless they go looking for that feature.

Do we actually want to go down the path where if we're in secure-by-default mode that the client will not be able to set normal attributes on the node?

danielsdeleo:
I think by default you want nodes to be able to set their own normal attributes, (even more so if tagging remains a thing that's implemented as normal attrs). Run list and other stuff (in policyfile land this would be policy_name and policy_group), you're much more likely to want to restrict.

lamont-granquist:
It sounds like we probably have at least a third endpoint that we need then.  We'll need to be able to separate out normal attrs from run_list/environment/tags.  And it sounds to me like normal attributes will need to be composed of normal-desired (node immutable) and normal-"current" (node mutable) state.

lamont-granquist:
Aside from this discussion, some notes from talking with Mark Anderson:  We tentatively concluded that the implementation should split up the underlying node object into desired and current state.  A migration on the server would need to be written to do this.   In the process, the existing ACLs would get taken off of the node object and applied to both the desired and current objects.  The endpoints for the full node object would behave the same way but would return the composed data.  The ACLs for the full node objects would likewise be composed of the ACLs on the components, and both ACLs would need to pass for an operation to succeed or else it would fail (so GET /nodes/:name would require read on both the desired and current state).

danielsdeleo:
Splitting normal attributes into two things seems pretty ugly. Is there a compelling case where a large quantity of nodes need snowflake attributes and should be prohibited from updating them? Why not use a data bag?

In #77 I propose that in the new API, "normal" attributes aren't treated as attributes at all, so you don't see them when you do `node.get_the_attribute_value_for("foo", "bar")`, instead you have to do something like `node.separate_node_specific_data_storage.get_value("foo", "bar")`. Since you have to specifically ask for a used-to-be-normal-attribute, you can choose at cookbook-authoring time to point to a data bag or other read-only location if you want that.

lamont-granquist:
So in the future then the run_list, environment, tags and normal attributes in general would all become "desired" state, and the ACL'ing to prevent the node from updating desired state would break normal attributes.  That means that at some point `node.normal['passwd'] = whatever` would break by default.

Then we could introduce in #77 or a related RFC that there would be an API to get and set local storage (which might just be a data bag, implementation TBD).  All of the existing use of normal attributes as current state would need to be converted to the new API (or if it was all mistaken use of normal attributes it would need to get converted to default/override attrs).

And to be clear the problem here is that tags being in normal attributes means that you've got "desired" state in normal attributes, while the use of setting passwords and stuff as normal attributes means that you've also got "current" state.

new/node_state_separation.md
```diff
@@ -0,0 +1,287 @@
+---
+RFC: unassigned
+Author: Lamont Granquist <lamont@chef.io>
+Status: Draft
+Type: <Standards Track>
+---
+
+# Node State Separation
+
+The purpose of this RFC is to make it possible to manipulate node objects
```
thommay: "seperately manipulate"? I'm not sure, just the language feels a bit off

new/node_state_separation.md
```diff
@@ -0,0 +1,287 @@
...
+as their desired "configuration" state and actual "runtime" state.
+
+## Rationale
+
+This proposal fixes [CHEF-4978](https://tickets.opscode.com/browse/CHEF-4978)
+and the associated tickets and issues that have occurred over years.
```
thommay: "the years"

new/node_state_separation.md
```diff
@@ -0,0 +1,287 @@
...
+
+## Motivation
+
+    As a user of the knife CLI
+    I want to be able to update my nodes run_list
```
thommay: "node's" or "nodes'"

new/node_state_separation.md
```diff
@@ -0,0 +1,287 @@
...
+
+This endpoint is unchanged, and is only included here for context.
+
+##### `GET /nodes/:name`
+
+This method will return the combined configuration and runtime data and its semantics will be unchanged.
+
+##### `PUT /nodes/:name`
+
+This method will create a new node with the combined configuration and runtime data and its semantics will be unchanged.
```
thommay: This doesn't tally with the text of "Node Save" below; needs to mention that in the dirty config case the node will fall back to PUTing here. 
Also, what happens if the actor has permissions to update `/nodes/:name/runtime` but not `/nodes/:name/configuration` and attempts to PUT here?

new/node_state_separation.md
```diff
@@ -0,0 +1,287 @@
...
+
+* 200 OK.  The request was successful
+* 401 Unauthorized.  The user or client who made the request could not be authenticated. Verify the user/client name, and that the correct key was used to sign the request.
+* 403 Forbidden.  The user which made the request is not authorized to perform the action.
+* 404 Not found.  The requested object does not exist.
+
+##### `PUT /nodes/:name/runtime`
+
+The `PUT` method is used to update the runtime data on an already created node.
+
```
thommay: Is it worth noting that this is expected to be only used by nodes?

new/node_state_separation.md
lamont-granquist: All of this needs to be updated with the discussion on ACLs.  I'm not planning on documenting how the API is used in the API documentation since that may change, the dirty config stuff is entirely client-side so wouldn't be documented here.

jaymzh:
For the record, lots of people *DO* want nodes to be able to update their own runlist. In fact, it's the ONLY thing that ever sets a runlist for us. We have a bootstrap script on the host that installs the chef RPM, generates a run_list.json, and the generated client.rb has an 'json_attrs' that points to it. That file is then on generated by Chef as a template, and if we need to push out a new runlist, we update that template.

So we need a way to have nodes update runlists.

jtimberman:
I also want nodes to update their own run lists sometimes.

lamont-granquist:
I think that question is out of scope for this RFC.  We can build the ability to ACL it, and then we can construct a Thunderdome later to figure out which configuration is the default.

jaymzh:
As long as you don't release a version in which I can't make this work. we will always want our nodes to be able to come up, register themselves and generate their own runlist. In no world do I want to have to tell my chef servers about them separately.

lamont-granquist:
Updates:

- Description section updated now that I have a better idea of the scope.
- Changed the endpoints to "/desired" and "/current" from "/configuration" and "/runtime"
- Added a bit of description in a server section about how the nodes will be composed and a bit on composing the ACLs.

I think there's still at least a piece missing on how to configure the containers so that the ACL'ing of desired state can be made to work easily.

thommay:
:+1:

lamont-granquist:
Mark is out on vacay, and I won't be able to get the last piece in place until I talk to him about it.

lamont-granquist:
Added a section on how the read-only desired state node configuration *might* be implemented.

The point is that there are a few ways forwards and we didn't see how any of them generated new requirements for this RFC, so there's no implications we're aware of.

The future implementation of read-only desired state node configuration is considered out of scope for this RFC.

new/node_state_separation.md
```diff
@@ -0,0 +1,350 @@
...
+If the node does not exist it will still be created with a POST to `/nodes`
+
+#### Node Save
+
+Provided the desired endpoint is supported, the Chef::Node#save method will be
+updated to typically PUT its data only to the `/nodes/:name/current` endpoint.
+This will affect both the final `node.save` of the chef-client run, and any
+user called `node.save` in recipe code.
+
+If it is detected that the desired data has been dirtied then `node.save` will
```
jaymzh: what does this mean? How does one detect dirty? Since the client gets a composed object, how does it no which pieces are from current vs desired?

Finally... when I request a runlist at the beginning of a run using the client, at the moment I don't believe that actually affects the node, it means the server does the dependency resolution based on that data instead of what it knows about the client... in the new model does that still work?

jonlives:
By the powers vested yada yada I hereby decide that this RFC is approved for merge. Take it away @chef/rfc-editors

btm:
rebased into 3e60a69abeab278284260546a969e7abbe714618 and accepted in 04aba6ea36cf81c18fc5846748f1b41a9a906f27.

new/node_state_separation.md
lamont-granquist: I was planning on saving a deep-dup'd copy of the desired state to compare against.  Dan suggested doing a checksum of the data which I was leaning towards thinking was premature optimization, but the implementation might be easier.  Either way I think that level of detail is out of scope for the RFC, its only important that the chef client implement that contract, the exact implementation can be left for a PR.

The client also POSTs its own run_list to the cookbook_versions endpoint, so the server behaves 'statelessly' there and does not use the node object it holds to compute the cookbook versions.  That won't need to be touched.

Issue #24, opened 2014-07-22
[Chef 12] Support nested directories


sethvargo:
#### Support nested directorires
 * **Breaks Backwards Compatibility? (Workstation):** No
 * **Breaks Backwards Compatibility? (Cookbooks):**  No
 * **User Benefit:** High
  * Short description of the benefit: As cookbooks become more complex, having the ability to logically group files in folders is incredibly useful, especially libraries. However, it would also be useful to group recipes and other resources as well.
 * **Impact:** None
  * Short description of the impact: This should be fully-BC, so there is no impact. There is a residual impact that cookbooks which implement this pattern would not be BC.

jkeiser:
I don't see an acceptance for this, but I'd love to see something more detailed--did I miss an RFC?  Is there one?

ranjib:
@jkeiser nope, there isn't one. @sethvargo do you plan to reformat this one? or raise another one?

This one provides the template:
https://github.com/chef/chef-rfc/blob/master/rfc000-rfc-process.md

jkeiser:
I'm building one now :)  It turns out we've already merged something that allows Chef client to upload a cookbook with nested library files; I'm adding something to make it useful in a way I want.

jkeiser:
Thus the RFC I'm building doesn't intersect with this, but it does make this RFC (and the merged patch) much more useful.  One minute.

jkeiser:
https://github.com/chef/chef-rfc/pull/93 is filed for autoloading.  If I read https://github.com/chef/chef/pull/2129#issuecomment-57237381 correctly, nested libraries are already possible (though I'm unclear on the load order or whether the nested directories are loaded).

jkeiser:
I think that might mean we can close this issue--@sethvargo does chef/chef#2129 fulfill what you were after?

johntdyer:
if I am reading chef/chef#2129 correctly its scope is limited to strictly libraries, and I believe @sethvargo's intent on this issue was to support other resources as well, including such as recipes.

dblessing:
@jkeiser I agree w/ @johntdyer - this should be kept open. https://github.com/chef/chef/pull/2129 does not address this RFC.

jkeiser:
Ah!  I missed that this was for all directories.  We need a real RFC though.

- files/ and templates/ are already hierarchical
- libraries/ is covered by chef/chef#2129 and autoloading behavior is addressed in #93.
[...]

PR #102, opened 2015-03-07
[WIP] Nested recipe support

dblessing:
A very fast draft as a result of conversation in #24. I wanted to get something rough up here for conversation so we can keep things moving. Please provide feedback and suggestions.

### DIFF
--- a/new/nested-recipes.md
+++ b/new/nested-recipes.md
@@ -1,0 +1,77 @@
+---
+RFC: unassigned
+Author: Drew Blessing <drew.blessing@mac.com>
+Status: Draft
+Type: <Standards Track, Informational, Process>
+<Replaces: RFCxxx>
+---
+
+# Title
+
+Add nested recipe support
+
+## Motivation
+
+As a cookbook developer,
+I want to be able to organize recipes into a nested directory structure
+so that recipes are more organized.
+
+## Specification
+
+Load files in the `recipes` directory recursively.
+
+Example directory structure:
+
+```
+|------ my_cookbook
+|        |------ recipes
+|        |        |____ default.rb
+|        |        |____ config.rb
+|        |        |____ config
+|        |        |         |______ client.rb
+|        |        |         |______ server.rb
+|        |        |____ install
+|        |        |         |______ client.rb
+|        |        |         |______ server.rb
+```
+
+Recipes will namespaced based on the directory structure and referenced using the namespace.
+
+`my_cookbook/recipes/install/client.rb` will be referenced as `recipe[my_cookbook::install::client]`
+in run lists and `my_cookbook::install::client` in recipes.
+
+Additionally, a recipe and a directory may have the same name and won't conflict.
+
+`my_cookbook/recipes/config.rb` and `my__cookbook/recipes/config/` are treated as both a
+recipe *and* a namespace. Therefore, `recipe[my_cookbook::config]` and `recipe[my_cookbook::config::server.rb`
+are valid references.
+
+## Rationale
+
+Organization can be a problem, especially in a complex cookbook with many recipes. With a nested directory
+structure the cookbook developer can split recipes up into logical groups. This also serves
+to make the cookbook easier to understand for other developers and for cookbook users as well. Consider
+that this is similar to the way Ruby library classes can be organized.
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.
+
+```
+
+## Influences
+
+[chef/chef-rfc#24](https://github.com/chef/chef-rfc/issues/24) - The issue has a lot of discussion
+about this proposed feature. Seth Vargo originally raised the issue. Jean Mertz work on
+[chef/chef#2129](https://github.com/chef/chef/pull/2129#issuecomment-57237381) is very similar. John
+Dyer also provides some examples in chef/chef-rfc#24.
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.
--- a/new/nested-recipes.md
+++ b/new/nested-recipes.md
@@ -42,8 +42,8 @@ in run lists and `my_cookbook::install::client` in recipes.
 
 Additionally, a recipe and a directory may have the same name and won't conflict.
 
-`my_cookbook/recipes/config.rb` and `my__cookbook/recipes/config/` are treated as both a
-recipe *and* a namespace. Therefore, `recipe[my_cookbook::config]` and `recipe[my_cookbook::config::server.rb`
+`my_cookbook/recipes/config.rb` and `my_cookbook/recipes/config/` are treated as both a
+recipe *and* a namespace. Therefore, `recipe[my_cookbook::config]` and `recipe[my_cookbook::config::server.rb]`
 are valid references.
 
 ## Rationale


coderanger:
I think this is out of scope for the moment given the minimal use case. Cookbooks big enough to need this kind of stuff are almost certainly in need of refactoring and moving logic to resources anyway. I think the conversation from #24 was aimed more at getting rid of the restrictions in the uploader code and how chef server handles files so we can build on things like this in the future. Cookbooks should allow nested folders in the recipes folder (and all other folders) even if there is no specific way to use them. You can then build on that low-level support to experiment with different layouts et al.

dblessing:
@coderanger I agree that the first step is getting Chef to load the resources in subdirectories. I'm not sure I agree that the impact is minimal. There are likely many more use-cases than I've outlined here so far. Another I've heard is that users would rather nest 'private' recipes in directories than have them be at the top level with an underscore at the front.

Chef has a ton of built-in resources that beginner-intermediate users can use to build about 90% of what they need. While LWRPs are generally not that difficult you're expecting the user to understand how they work as opposed to simply using the built-in Chef resources in nested recipes. 

FWIW, coming from an almost 5 year background in Puppet I was astonished when I tried the nested recipes and had no luck. Maybe that's just my broken thinking coming form Puppet but I think this is something that should be there. Also, any person familiar with Ruby might expect this behavior to be there.

coderanger:
Stuff like "private" recipes is a great indication you shouldn't be writing that as a recipe :-)

johntdyer:
@dblessing I think this looks great.   In our organization we have a large "monitoring" cookbook which has nearly a hundred recipes, and gaining the ability to separate these recipes into logical groupings based on concern ( server, client, ect ) would be a big win, even if only for readability sake.  It can be very intimidating for someone to pick up this large cookbook, open up the recipes folder, and get slapped in the face with almost a 100 recipes.  :)

nathenharvey:
@dblessing this PR is on the agenda for this week's Chef Developers' IRC meeting.  https://github.com/chef/chef-community-irc-meetings/blob/master/2015-03-26-agenda.md

btm:
I'm in support of loading recipes from subdirectories, and the specification looks like a reasonable way to do it. It just makes sense that you'd be able to. I think #24 also wants a separate RFC that says the client/server should change to get rid of the concept of segments, but we still need this to specify what happens when you have multiple recipes which is what this does.

thommay:
:+1:

someara:
-1 on this one... 

It sounds good on the face of it, but starts to fall down when thinking about the UX implications.

Recipes are referred to from many places:
1) recipes via include_recipe ```"cookbook_name::recipe_name"```
2) roles via ```recipe[cookbook_name::recipe_name]```
3) environment via ```recipe[cookbook_name::recipe_name]```

Currently, the cookbook artifact structure is ```cookbook_name/recipes/```

If that was changed to ```cookbook_name/recipes/{sub1,sub2,sub3}```, what would the implications be for the referencing objects?

Would it be ```include_recipe[cookbook_name::subdirectory/recipe_name]```? 
Would it be ```include_recipe[cookbook_name::subdirectory::recipe_name]```?
how about...
Would it be ```include_recipe[cookbook_name::subdirectory::subdirectory2::recipe_name]```?

The A::B syntax already means "subclass" in Regular Ruby, and making it mean something else in (parts of) Chef is already confusing enough... 

I see this opening a wormhole to confusion town.

FWIW, I've always found it easier and cleaner to break cookbooks with large amounts of recipes up into multiple cookbooks.

/2 cents
-s

luckymike:
As a co-maintainer of many community cookbooks, I'm in agreement with @coderanger. Nested recipes seem like window-dressing to address a systemic problem. Making it easy to hide the whole world in a cookbook will only encourage the wrong style of cookbook development. 

If a one-cookbook-per-repo strategy results in internal org cookbooks that have dozens of recipes, that's a workflow issue that should be addressed internally, not draped on the community as a whole through an RFC.

:-1:

martinb3:
As someone who maintains various community cookbooks, I'd also give this a -1. Recipes are a DSL designed to simplify a complex problem. At some point, it becomes easier to go to a Ruby solution than it does to make the DSL more complex. We should keep the Recipe DSL simple, and encourage the more complex use-cases to revert to pure Ruby.

If we had to implement it, I'd favor this syntax over the others:
```
include_recipe[cookbook_name::subdirectory/recipe_name]
```

dblessing:
Puppet has solved this issue long ago and there haven't been any issues with it. Nested manifests are referenced as 'module::segment1::segment2::manfiest'. That's exactly what you would expect from Ruby, too. There are plenty of reasons to have a nested structure that do not necessarily mean the cookbook is too complicated. I don't see how this structure could cause confusion. 

I agree that using resources for more complex uses is the best method and I can see the argument for breaking into more cookbooks, too. But all of these things come at a higher cost. The user must now understand Ruby development and/or spend the extra time configuring a new cookbook end-to-end which may include a repo, CI pipeline, etc. 

Additionally, I remind Chef users that one of the main arguments for Puppet and against Chef for inexperienced developers (or sysadmins not experienced with CM) is that you need to know Ruby. The Chef concept of 'just enough Ruby' isn't enough for new resource development. 

I hope this helps provide perspective from someone who has only recently joined the Chef community.

dblessing:
@nathanharvey darn! I didnt see your message in time. Sorry I missed it. Will it come up again?

btm:
@dblessing it's on next weeks agenda already: https://github.com/chef/chef-community-irc-meetings/blob/master/2015-04-16-agenda.md

dblessing:
Thanks. I'll be there. I also look forward to continuing discussion here.

someara:
It's very important to understand that Chef != Puppet.

Puppet modules are "class" based, and the directory structure has meaning
to way those are constructed.

my_module/manifests/implementation/foo.pp maps to
my_module::implementation::foo maps to

Recipes do not work that way *at all*..
That structure might make potential sense for resource/provider
construction, but recipes are a much flatter concept.

The resources in recipes are (by default) compiled into a single, flat
resource_collection.
Recipes are referenced from various places (run_list, roles, other recipes).

The reference syntax and artifact file location is the real issue here.

PS: "One of the main arguments for Puppet and against Chef for
inexperienced developers (or sysadmins not experienced with CM) is that you
need to know Ruby." has always been a straw man. Puppet is its own weird
language that needs to be learned. Puppet has just as many flow control
constructs, lambdas, variables, and scopes as the next thing.

dblessing:
Resources can still be compiled into a flat resource collection. Yes, reference logic has to be changed. Is there something especially complex that would prevent this from being accomplished?

ranjib:
Following up on what i have said on IRC:
resources are not always compiled into flat resource collection. depending upon how and where you declare them (use_inline_resources exists for similar reason). 
code organization is a generic problem across resource/provider/recipes. Nested recipe -> namespace mapping needs to be done cleanly. From the recent resource/provider mapping overhaul, we have learned lot. I think it will be better if we introduce a resolver like abstraction, and allow include_recipe to fetch recipes via that. Chef client can do the namespace /class construction magic. This will also allowed writing curried recipes (aka recipes with certain variables pre-defined). 
include_recipe 'bar', from: Foo::Bar, with:(x: 1, y:z). we'll need few more helpers similar to use_inline_resources to deal with different notifications scenarios. If a community cookbook comes with nested recipes i should be able to decide whether or not expose the resource list in top level resource collection.

in its current format, this RFC only encourage building deep nesting and its possible to address that without any new feature as well (split up client and server into different recipe etc).

we should also note that in chef directories have special values depending upon what segment you use. files, templates etc uses directory to pick up platform specific options.

someara:
It's probably important to define what exactly is meant by "nested recipes"
here.

Are we talking resources / recipes-as-provider-actions?

Here be dragons, and a lot of prior work ala poise/crazytown.

From the original message, I got the impression that it was the mere
ability to have subdirectories in the recipes/ folder for organization
purposes.

jtimberman:
I'm :-1: on this feature.

My primary concern is regarding community cookbooks. If a popular (or unpopular, too) cookbook is refactored to be "more organized" using this feature, users who are using older versions of Chef will be unable to use that cookbook. Yes, they certainly can pin their version of the cookbook, but that's additional burden put on them when they may need / want features or bug fixes to said cookbook(s) for their infrastructure. One of the points of community cookbooks is they're reusable, and people gain the benefit of community maintenance, rather than having to maintain their own (forks, copies, wrappers, etc). I feel that this can make the problem with community cookbooks worse, not better. It seems to me that this would be harder to adapt one's local run lists to use, unlike other changes that break compatibility for old existing versions of Chef.

While it *is* best practice to pin versions of cookbooks to the precise version that is tested, in reality we know full well that people don't do that until they hit something like this where something breaks and they spend all day fixing it. This is a really poor user experience.

Recipes aren't like libraries in other software development contexts. They are not Ruby classes. I don't see the value that this brings for organization of a cookbook. Why do we need this 6 years into Chef's lifespan? As near as I can tell we only had [one CHEF ticket in Jira](https://tickets.opscode.com/browse/CHEF-3775) bringing this up, but we never moved forward with it back then (January 2013), either. In my opinion, if you have such a large number of recipes that you need to organize them, I think that the cookbook's complexity is probably sufficient to abstract things out to resources, or separate cookbooks.

In the conversation during the developer's IRC meeting,

> [10:25] <jonlives> we have the RFC to add chef_version metadata to stop that, in a turtles-all-the-way-down twist.

For reference, this is [RFC 037](https://github.com/chef/chef-rfc/blob/master/rfc037-chef-ohai-version-metadata.md) which was accepted on Feb 19, 2015. Older versions of Chef wouldn't make use of that either, but I'd like to have that feature above this one wrt where contributors spend time implementing new features.

And TL;DR, I agree with words that @someara said.

dblessing:
After hearing all arguments I'm OK with not adding nested support. However, I would like to see improvements to LWRP documentation. The current docs only provide a very simple example and I'm left with questions like what the heck `use_inline_resources` really does and the implications it has on resource notifications. Also, when and when don't I need to worry about `updated_by_last_action` stuff? Clarifying all of these things will help lower the barrier for new Chef users.

The poise stuff looks like it's actually going to make HWRP easier than LWRP and it takes care of the confusing notification problems w/ LWRPs. I'm definitely going to try that out soon.

jonlives:
After taking all feedback into account, we're going to close off this PR. An action item from the same community meeting has been created to file an issue against the chef-docs repo to look at improving the LWRP documentation, and I'll be taking the lead on getting that started.

jonlives:
Issue filed for improving LWRP docs at https://github.com/chef/chef-docs/issues/499

PR #103, opened 2015-03-10
Reopen RFC 28, Mailing List Migration

nathenharvey:
Migrating to Google Groups failed
Suggest migrating to Discourse

### DIFF
--- a/rfc028-mailing-list-migration.md
+++ b/rfc028-mailing-list-migration.md
@@ -1,14 +1,15 @@
 ---
 RFC: 28
 Author: Julian C. Dunn <jdunn@aquezada.com>
-Status: Accepted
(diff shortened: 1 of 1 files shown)


coderanger:
Should probably add a note about what failed with Groups.

jaymzh:
indeed. and how interaction looks via email for the more rich things people can do in discourse which is going to be a deal breaker for a large portion of the community.

nathenharvey:
I've updated the RFC with some more info on discourse and why the migration to Google Groups failed.

adamhjk:
By the power vested in me by having written Chef in the first place, and continuing to not be an asshole, I approve this RFC. Take it away, @chef/rfc-editors

PR #104, opened 2015-03-16
Combine the rationale and description

thommay:


lamont-granquist:
Yes :+1:

tyler-ball:
:+1:

coderanger:
Because yodawg, whats the description and/or rationale for this?

jaymzh:
:+1:

jonlives:
:+1:

coderanger:
Was this discussed somewhere I can't see? We should really be recording the "why" of changes like this for the future. /cc @adamhjk

thommay:
Sorry, I didn't realise you were asking a question. 
Because most of the time the description is the rationale, and even when it's not you want to read the rationale before the specification.

PR #108, opened 2015-03-23
ohai identity attributes RFC

lamont-granquist:


new/ohai_identity_attributes.md
```diff
@@ -0,0 +1,124 @@
...
+
+This changes the semantics of the `fqdn` attribute and may be a breaking change, but would need to be
+introduced in a minor release of Chef 12 in order to fix the issue for Solaris customers.  The risk is low.
+
+## Second Proposal (Long Term Fix)
+
+Since the short term fix breaks the strict semantics of the `fqdn` attribute, that opens the door as to if
+those semantics were right in the first place.
+
+The `hostname -f` lookup has an inherent weakness in that it relies on DNS, which may fail.  It is also obvious
```
btm: Isn't this "may rely on DNS" depending on host.conf order?

thommay:
:+1:, and I tend to agree that the risk is low on the short term fix.

btm:
I think we need some way to configure it, especially since overriding Ohai plugins is still non-trivial. I have memories of lots of people struggling with Chef/hostname issues but I can't recall the exact breadth of the issues right now. 

Some are easiest to fix by fixing your machine, like users testing the Chef Server on a virtual machine with name resolution barely configured, and changing /etc/hosts fixing the issue.

But we've seen users with huge infrastructures with DNS resolution configured "incorrectly" and telling them to fix their infrastructure before they use Chef is a non-starter. 

Something like a host.conf/nsswitch.conf order?

new/ohai_identity_attributes.md
lamont-granquist: And depends on what you have in your /etc/hosts file or NIS or LDAP or whatever you are using.

If all of your servers have their FQDNs resolvable in /etc/hosts then we don't depend on DNS, but before we run Chef the first time that isn't guaranteed, and even afterwords people don't understand how /etc/hosts works and how to configure it, and some applications straight up break on standard /etc/hosts files and so their hosts have to mangle them to make the apps happy (thinking of one particularly nasty third party proxying software that a large internet retailer used in its image serving stack -- crazy shit happens in IT).

lamont-granquist:
A lot of the hostname issues were fixed or papered over (mostly) by the changes in Ohai 7.0 and the introduction of the machinename parameter and corresponding change to initial node_name guessing so that Chef could use the static hostname.

I'm open to suggestions on how to configure it, I don't think we currently do that for anything in ohai?

new/ohai_identity_attributes.md
```diff
@@ -0,0 +1,124 @@
...
+```ruby
+name = Chef::Config[:node_name] || ohai[:fqdn] || ohai[:machinename] || ohai[:hostname]
+```
+
+It was discussed making the `fqdn` attribute to prefer the return of `/bin/hostname` when that returned a
+FQDN, but this felt to break some edge conditions in DNS related cookbooks (djbdns?) and so this was forced
+to be the fqdn which was resolved via DNS resolution across all platforms.  For platforms that do not support
+the `-f` flag to `hostname` a pure ruby version is used by ohai.
+
+The changes in Ohai 7.0.0 to the `fqdn` attribute broke Solaris people with busted DNS who expected the
```
adamedx: To clarify -- do these Solaris users agree their DNS is incorrect? If not, I'd rather stay neutral here with the description of the DNS configuration.

new/ohai_identity_attributes.md
btm: Yeah, I ran across a Stackexchange issue recently where someone said they'r hosts file had to be "ip_address alias canonical_hostname" for some other software, but Chef was using "alias" for the fqdn after the first run. Which of course, is expected behavior, and it's the other application that's doing it wrong.

---
The `hostname -f` looking has an inherent weakness in that it relies on the system's resolver. Depending on the resolver configuration, we may receive incorrect answers due to intentional misconfiguration of the hosts file, or even failures due to unavailable DNS, NIS, etc. It is also obvious that both cookbook users and authors expect the `fqdn` attribute to DWIM.  The use case for having the `fqdn` attribute be strict is a very small subset of a few cookbooks and is the uncommon case.

new/ohai_identity_attributes.md
lamont-granquist: I believe they do.  Broken is broken, though.  I can be more verbose and talk around it, but if people didn't have broken DNS this wouldn't be an issue since the forward+reverse lookup would not fail.

new/ohai_identity_attributes.md
lamont-granquist: Yeah, sometimes its application authors that don't understand /etc/hosts and force the short name alias first, which breaks expected name resolution and breaks `hostname -f`.  We can't change that.

lamont-granquist: I also consider that to be "broken DNS" even though its /etc/hosts which is broken.

btm:
There aren't that many [Ohai settings](https://docs.chef.io/ohai.html#ohai-settings-in-client-rb) right now. I think the most common one is disabled_plugins.

These Ohai proposed configuration options  were part of the Ohai 7 internal RFC, but were not implemented:
* required_attributes: Array of strings List of attributes that will be marked as critical.
Chef will automatically set the list of attributes it depends on without action from user.
* disabled_attributes: Array of strings List of attributes that will not be collected by OHAI.
* run_mode: [symbol: :all_plugins, :required_plugins] Flag that specifies OHAI to run either all plugins or only required ones. Default is set to all_plugins in order not to break existing logic that depends on optional plugins.

In all cases, the "trick" is how do you get them into your client.rb on a node that you're bootstrapping. If you use the [chef-client cookbook](https://github.com/opscode-cookbooks/chef-client/blob/master/recipes/config.rb) you've already loaded Ohai in the first run, so you potentially have different values on your first run than your second. Short of adding more flags for individual settings we put into the bootstrap template, I'm hopeful @ranjib's RFC to replace #82 will give us a framework for improving this.

Of course, I'd like to avoid abusing the hints system into being a configuration system. If we can make it easier to get configuration values into client.rb, I think it's easy to use Ohai::Config.

btm:
And if we're greenfielding plugin configuration (rather than hints), we should come to agreement on something like Ohai::Config[:plugin] namespace or Ohai::Config[:something_config][:name_of_attribute] to be more Ohai 7.

lamont-granquist:
Shower thought:

So we could refactor this whole proposal into a single non-breaking config setting.

```ruby
Ohai::Config[:fqdn_using] = :dns
Ohai::Config[:fqdn_using] = :hostname
Ohai::Config[:fqdn_using] = :nis
Ohai::Config[:fqdn_using] = :dwim
```

Where `:dwim` might be `:hostname`, then `:nis`, then `:dns`.

Then we tell the affected customer with the Solaris hosts to set it to `:nis` right now and they're fixed.  At some point in the future we switch the default from `:dns` to `:dwim`.

adamedx:
Yeah, by "right now" you mean once we release this refactor, right @lamont-granquist? They still need some other workaround until we release a new Ohai, right?

lamont-granquist:
I don't see a workaround unless we patch ohai.

btm:
Let's keep specific solaris hostname discussion and workaround over here: https://github.com/chef/ohai/issues/420.

jonlives:
@lamont-granquist I really like this config setting idea, we can have defaults that make sense for various platforms and then people can change it if they want.

lamont-granquist:
Make it an array, then `:dwim` just becomes:

```ruby
Ohai::Config[:fqdn_using] = [ :hostname, :nis, :dns ]
```

btm:
It was agreed upon in today's IRC meeting that we'd like to have a table of what would change, by platform, so we can estimate how much of a breaking change this would be.

lamont-granquist:
In looking over the Ohai code it looks like ohai never parses the `/etc/chef/client.rb` file itself, so it relies on chef-client parsing that file and the Ohai::Config[:whatever] values being set by chef-client.

This will lead to different behavior between ohai invoked on the command line and ohai invoked via chef-client (an issue that we already have when it comes to disabled_plugins and a couple other config values).

Having a difference between what the command line shows and chef-client gets for the fqdn seems to me to be a fairly major discrepancy, though, and something that should probably get addressed.

We could try to have ohai load the client.rb file, but most of the methods there are going to be client setttings which should be ignored by ohai.  We could load it up into an object that method_missing'd all of the config settings and that should allow the `Ohai::Config[:thing] = :whatever` setting to apply to the command line.   We could reuse Chef::Config to fully parse and load the config file, but that introduces a circular dependency and does a lot of work we'll just throw away.

Thoughts?

jonlives:
I'm kind of errring on the side of using method_missing to handle the irrelevant-to-ohai config settings, seems like overkill to use Chef::Config for this IMO - thoughts from others?

btm:
I think we need to make a separate Ohai Config RFC, then specify using that here as @lamont-granquist keenly suggested. I wonder then how that affects the first and second proposal. Can we drop the first proposal and zero in on the second one for "permanent" or long-term agreement on what these attributes should be?

---
Ohai Config RFC thoughts:

Having Ohai read ```/etc/chef/client.rb``` itself and ignore the Chef config settings seems fine. I assume we'd do that in Ohai::Application#configure_ohai so it won't affect a Chef run, e.g. we're not overwriting any config settings that came from who knows where in someone's Chef run. We'll need to add a -c argument to change the file that we load too.

I'd really like to namespace the plugin configuration values since we're the first ones introducing plugin configuration to Ohai, but it's a little wordy. This should help protect us from taking patches that overwrite someone elses custom top level configuration value down the road.

```
Ohai::Config[:plugin][:hostname][:fqdn_using] = [ :hostname, :nis, :dns ]
```

Eventually we could do helpful things like warn if Ohai::Config[:plugin][:foo][:anything] is set but plugin foo isn't loaded. 

Since we would only be using the namespace for humans, I think we're safe using the name of the plugin and don't need to go through the trouble of namespacing by attribute (since we previously came to the conclusion that we cared more about specific attributes than plugins in the Ohai 7 design process). @mcquin any thoughts on this?

jonlives:
Per the community meeting on 09/04/2015, this RFC is on hold pending @btm submitting an RFC to propose a format for specifying Ohai config to make implementation more easy here. I'll update further with details of this RFC once it's submitted.

new/ohai_identity_attributes.md
```diff
@@ -0,0 +1,124 @@
...
+a problem because Chef::Client uses it internally as first precedence for guessing the `node_name` and `fqdn` is
+used by many cookbooks.
+
+# Specification
+
+## First Proposal (Short Term Fix)
+
+The first proposal is to modify the `fqdn` attribute to be:
+
+# Equivalent to `/bin/hostname -f` if that algorithm succeeds in returning a FQDN
```
jaymzh: At the moment we (I added this PR) try `/bin/hostname -f` a few times to avoid errant issues... we should keep that as part of the algorithm and then continue down the list.

new/ohai_identity_attributes.md
```diff
@@ -0,0 +1,124 @@
...
+Since the short term fix breaks the strict semantics of the `fqdn` attribute, that opens the door as to if
+those semantics were right in the first place.
+
+The `hostname -f` lookup has an inherent weakness in that it relies on DNS, which may fail.  It is also obvious
+that both cookbook users and authors expect the `fqdn` attribute to DWIM.  The use case for having the `fqdn`
+attribute be strict is a very small subset of a few cookbooks and is the uncommon case.
+
+It is proposed that the `fqdn` attribute semantics be changed to:
+
+# Returning `/bin/hostname` if that returns a FQDN
```
jaymzh: how do you know if it's an FQDN? I've seen people do weird shit where it returns foo.bar but the hostname is foo.bar.baz.bing.

jaymzh:
Also, I'm pretty stoked about @btm's idea, looking forward to that RFC.

(sorry, catching up on RFC PRs...)

mcquin:
@btm I'll wait to add my thought until your RFC for an ohai config comes out. If you'd like to collaborate on it, you know how to reach me. =)

nathenharvey:
Pending #118

btm:
This needs an update specifying how something like `ohai.plugin[:hostname][:fqdn_using] = [ :hostname, :nis, :dns ]` would work, per [rfc053](https://github.com/chef/chef-rfc/blob/master/rfc053-ohai-config.md#plugin-namespacing).

PR #109, opened 2015-03-25
adding reclassify_amazon_linux_platform_family.md

someara:


zts:
I use Amazon Linux, and I think I'm +1 on this - for a time it was possible to treat it as roughly equivalent to RHEL6, but it has diverged far enough to need special treatment more often than not. (In my experience, anyway)

That said, this will break a bunch of community cookbooks which work today, on the basis that they'll break at some point in the future anyway.  If we made this change in a 12.x release, I'd hope we included a config option to maintain the lie a little longer (while folks are fixing the issues this will cause).

lamont-granquist:
Technically, it never actually worked back when Amazon Linux was a hybrid of RHEL5+RHEL6.   Trying to pull sane Java packages from EPEL5/6 back in 2010 or so was terrible since whichever one you picked you'd eventually wind up with some kind of conflict when you hit the wrong package in the base that conflicted.  It actually got better for awhile as everything moved onto RHEL6 and has now reverted to being worse as its morphing into a RHEL6+RHEL7 hybrid.

It was originally added as a fedora 'family' as well but that was changed to RHEL in 

https://tickets.opscode.com/browse/OHAI-331

It does mean that cookbook authors will need to do more work to support the amazon distro and can't just lazily rely on 'rhel' but in reality you never could quite do that.

:+1:

And, yes, we probably need a lie-about-amazon-being-rhel config switch that defaults one way or the other.

thommay:
:+1: as long as we optionally lie.

philoserf:
In 12.x I lean toward 'the lie should be the default'. Warn,, opt-in, then later, deprecate.

someara:
Not sure how to implement the lie, as this change would be in Ohai

zts:
How about an ohai hint?

lamont-granquist:
#108 has a similar issue with ohai.  abusing the hinting system is probably not the best idea, but something like adding more `Chef::Config[ohai_*]` options or something...

btm:
We definitely need an Ohai Config RFC as we've been discussing in #108, then it's an easy

```
when /amazon/
  if Ohai::Config[:plugin][:platform][:amazon_is_amazon]
    platform_family "amazon"
  else
    Ohai::Log.warn("this train is going elsewhere, read this thing on the internet:")
    platform_family "rhel"
  end
```

jonlives:
This would make things considerably easier, and a lot easier to ensure this isn't a breaking change until we want it to be.

new/reclassify_amazon_linux_platform_family.md
```diff
@@ -0,0 +1,31 @@
+---
+RFC: unassigned
+Author: Sean OMeara <sean@chef.io>
+Status: Draft
+Type: <Standards Track, Informational, Process>
+<Replaces: RFCxxx>
```
btm: Can you can change Type to 'Standards Track' and remove the ```Replaces``` line when you edit this next to save some work on merge later?

jonlives:
Per the community meeting on 09/04/2015, this RFC is on hold pending @btm submitting an RFC to propose a format for specifying Ohai config to make implementation more easy here. I'll update further with details of this RFC once it's submitted.

nathenharvey:
Pending #118

btm:
This needs a reference to RFC053 now that it's accepted.

https://github.com/chef/chef-rfc/blob/master/rfc053-ohai-config.md#plugin-namespacing

    ## Specification
    
    Over time, change Amazon Linux platform_family from 'rhel' to 'amazon'.
    
    Initially default `Ohai.config[:plugin][:platform][:amazon_is_amazon]` to `nil`, which will preserve behavior and log a warning about the future change. If the value is `false`, the existing behavior will be maintained but no warning will be logged. When `true`, platform_family will be returned as `amazon`.

new/reclassify_amazon_linux_platform_family.md
```diff
@@ -0,0 +1,30 @@
...
+   As time passes, Amazon Linux looks less and less like a RHEL clone.
+   Release version numbers do not match EL5, EL6, EL7.
+   Amazon is now a rolling release
+   Different package sets are available
+   Package names and contents are different
+   Service providers are different
+
+## Specification
+
+Change Amazon Linux platform_family from 'rhel' to 'amazon'
```
lamont-granquist: So I think we should change the platform_family here to 'fedora' rather than 'amazon'

Having platform_family and platform of fedora/fedora and amazon/amazon doesn't allow leveraging any existing code, and we have to update all the if cases.  If we reuse 'fedora' then code like this just works:

```ruby
case platform_family
when 'rhel', 'fedora'
  # amazon also works fine here
end
```

For code that needs to specifically look at fedora as a distro and cares about things like platform_version then this code will still work:

```ruby
case platform
when 'fedora'
  # inappropriate to match amazon here
  platform_family >= 21 ? "systemd" : "sysvinit"
end
```

There might be some (bad) code which matches on platform_family of 'fedora' and then does very fedora-specific code which would break on 'amazon', but that code should be changed to use platform of 'fedora' instead of 'platform_family'.

This gets us back to the original design goal of platform_family which is to let you be lazy with case statements and have good branching logic.  To that goal it makes more sense to only have the two 'family' buckets where 'rhel' is explicitly 'rebranded redhat enterprise linux with compatible platform_version numbers' and 'fedora' is 'all other fedora-derived distributions'.

someara: Amazon Linux isn't Fedora derived anymore.

It really truly is its own thing.

The "things" you case-statement around are not compatible with Fedora.

It's a rolling release.

Package wise, Amazon puts semantically meaningful information into package
names (much like Debian),
The package dependency graph deviated a very long time ago.

A Fedora package repo won't work on Amazon, and an Amazon package repo
doesn't work on Fedora.

Service wise, Amazon currently uses sysvinit. Fedora does uses systemd.

Amazon still uses Yum, Fedora use DNF now, etc.

It should be 'amazon/amazon'

lamont-granquist: Amazon using sysvinit and Yum is just like old fedora, that isn't a significant difference.  The Provider Resolver code can also just handle that and pickup the RedHat service provider or the systemd provider based on duck-typing the operating system.  We should also duck-type Yum-vs-DNF as well to make those differences transparent and not require hand-wiring up lots of platform-version switching logic.

And package repo compatibility isn't necessarily expected at the family level (e.g. "debian" vs. "ubuntu" both being in the same "debian" family).  That /is/ what makes the platform_family of 'rhel' work, but we don't expect it anywhere else.  And they're still both RPM based distrubutions with basically the same layout of config files as RHEL/Fedora, which makes it Fedora-derived.

someara: If we make Amazon Linux 's family "fedora", we'll just have to start
putting tons of exception logic around platform_family 'fedora' instead of
"rhel".


On Sun, Jul 12, 2015 at 10:57 PM, Lamont Granquist <notifications@github.com

lamont-granquist: In the case where you need to branch between fedora and amazon you have to branch anyway.  There's no net increase or decrease in work.  In the case where fedora and amazon match, however, then if they share the same platform_family then you can be lazy and you don't need to update your case statements and it'll just work.  With the 'amazon' platform_family approach now we have to through every cookbook in existence and update them to add `when "fedora", "amazon"` to every single fedora platform_family case.

someara: I want to get this right the first time... 

I guess it really comes down to the intent of the Amazon Linux maintainers on how far they plan on deviating from Fedora.

Reaching out...

someara: I received a response from the Amazon Linux AMI team with permission to post it here for transparency:

We consider the Amazon Linux AMI to be a distant cousin in the extended family of Red Hat-inspired Linux offerings.  We don't have plans to converge on a specific set of packages from RHEL 6, RHEL 7, or any particular Fedora release, because our customers ask us for features and functionality that exists in all of those different distributions.  Additionally, the Amazon Linux AMI includes packages and features for which Amazon is the upstream.  This allows us to support the latest features of EC2 and AWS as they are launched.

Because the roadmap for the Amazon Linux AMI is driven very strongly by customer requests, we have drifted from RHEL and Fedora in some important ways, including the kernel version that we provide and when we change it, providing multiple versions of languages (PHP, Python, Ruby, etc.) in the OS repositories directly, and the "aggressiveness" that we show towards certain packages in terms of keeping close to the latest from upstream, even when that means a lot of churn.  Another major difference is that we are still using sysvinit/upstart while the latest RHEL and Fedora have moved to systemd.

Based on the above reasoning, we include ID_LIKE="rhel fedora" in the /etc/os-release file.

someara: case statements around packages names are unique (debian-like alternative system)
case statements around packages versions are unique (less 3rd party repos like remi, ius, oracle)
case statements around packages availability are unique (aws tools)
case statements around service management are unique (if unimportant)

-s

lamont-granquist: So again this sounds like less of a difference to me than ubuntu-vs-debian, both of which are platform_family of debian.   The package repos and package versions along with package availability are completely different.  There has been at least one totally different behavior in the core service resource:  upstart.  For a lot of case statements though a platform_family of 'debian' that covers the latest thing that ubuntu and debian both do the same (as opposed to fedora-based, suse-based, solaris/aix/bsd, etc.) will work fine.

lamont-granquist:
So this still sounds like platform_family of 'fedora' to me.  And the platform_family concept is designed around the case statements that users have to write and making it easy to get idioms correct.

Before platform_family we had code like this:

````
case node['platform']
when 'ubuntu', 'debian', 'raspbian', 'linuxmint'
when 'redhat', 'centos', 'scientific', 'oracle', 'fedora'
else
end
```

And every time a new platform got added all the case statements needed to be upgraded.  And the problem here is that there's no way to do a useful default-else-for-fedora-like things.  What we really need is that ability to group defaults according to very rough and major differences in O/S distributions.

```
case node['platform_family']
when 'rhel'
else
  # this is everything else -- solaris, aix, bsd, whatever we don't currently support.
end
```

While this works as a default-else:

```
case node['platform_family']
when 'rhel'
when 'fedora'
  # this catches fedora, amazon, anything else fedora derived in the future
when 'debian'
end
```

We also have this problem in solaris land where all the solaris-ish operating systems have been given unique platform_family identifiers making it impossible to write a default-else to catch them or to catch any future solaris-ish operating systems which are introduced.  We wind up doing this:

````
        provides :mysql_service, os: %w(solaris2 omnios smartos openindiana opensolaris nexentacore) do |node|
          File.exist?("/usr/sbin/svccfg")
        end
```

When we could instead do this:

```
      if defined?(provides)
        provides :mysql_service, platform_family: "solarisish"
          File.exist?("/usr/sbin/svccfg")
        end
      end
```

As a rule, I would actually argue that having single-platform-per-platform-family is nearly always codesmell (except for completely proprietary O/Sen like aix or hpux that nobody has forked or legally can fork).

And like the Amazon team posted their /etc/os-release contains:

```
ID_LIKE="rhel fedora"
```

It seems like there shouldn't be any argument even that platform_family of 'fedora' works because it exactly what they're declaring there.

And this gets back to where we meant to be initially with this platform_family:

```
case node['platform_family']
when "rhel"
  # I can match on platform_version.to_i == 6 here and it is RHEL6
when "fedora"
  # You configure my hostname in /etc/sysconfig/network, but my pv doesn't follow RHEL
when "debian"
  # You configure my hostname in /etc/hostname
end
```

Otherwise, to get the easy case correct then we will force everyone to update all the case statements everywhere to this:

```
case node['platform_family']
when "rhel", "fedora", "amazon"
  # behave like redhat
when "debian"
  # behave like debian
end
```

And I think if you poll a reasonably experienced cross-platform sysadmin who sits down and tries to administer an Amazon Linux box as to what kind of distro it is, they're going to tell you that its a fedora/redhat distro.  It'll be significantly different from SuSE, from the BSDs, from Solaris/AIX/etc, but if they know RHEL/fedora it'll be easiest to describe the operating system as "Its like RedHat, except the kernel is different, and...." which is exactly how the Amazon team described it.

Also, from a backwards compatibility perspective changing the platform_family to 'fedora' will be significantly less disruptive to existing cookbook code.  Case statements around platform_family that handle 'rhel' and 'fedora' will at least have a much better chance of working correctly -- Amazon linux will at least fall into redhat-ish codepaths.  If we introduce a new platform_family for "amazon" then case statements that do not have a global else that happens to guess redhat-like will definitely be broken (and many default-else's out there are debian-like due to initial debian/ubuntu bias in Chef).

lamont-granquist:
It also seems like there's a view out there which is that platform_family MUST only be applied to distros which are nearly straight recompiles.  If we take that to its logical conclusion then it seems like the only really useful platform_family is 'rhel'.  While that was a major use case of platform_family that was never intended as the only use case.  Following that to its logical conclusion we should extract ubuntu, linuxmint, raspbian, etc out of the "debian" platform_family.   But the result would render platform_family only useful for RHEL.  That is would render it a fairly useless construct.

Since that definition of what a platform_family "is" produces a useless construct, it seems that the better definition of what a platform_family "is" revolves around more of a "smell test" based on how a sysadmin would describe the distro to another sysadmin.  If that description starts out with "its like XXX only they've changed X, Y and Z" then you've probably got a platform_family.  It cannot be sufficient to simply to point to X, Y and Z and claim that there's two distinct platform_families.

zts:
Coming at this from the opposite direction, what problems would we solve (or ameliorate) by changing Amazon Linux's platform_family to `fedora`?

The problems I've encountered have generally been around packages.  For example, a cookbook adding a vendor yum repository whose RHEL packages were incompatible with the OS, or attempting to install (broken) packages from EPEL for software that is available in the distro, or assuming an older/newer major version would be installed.  I can't readily think of any examples where treating the platform as fedora would have helped.

As a user, I feel the pain when I encounter these issues, but I don't give much thought to things that "just work".  I'm certainly using cookbooks that happen to do the right thing on Amazon Linux by assuming it's RHEL, and I'm sure they'll break if that changes.  And that would be sad, but it would require only small fixes for those things that were already working, while those that weren't would (perhaps) fail more obviously.

To the smell test.  Prior to using Amazon Linux, I was at a shop running Centos for a number of years.  If you'd asked me then, I'd have described Amazon Linux as RHEL-derived, and assumed `platform_family == "rhel"` would do the right thing.  Today, I need to run many of my test suites against Amazon Linux because a pass on Centos in development doesn't mean something will work in production.  I wouldn't call it closer to Fedora than RHEL, and neither make a consistently reliable approximation for the purposes of cookbook development.

lamont-granquist:
The EPEL problem is solved by moving the platform_family out of 'rhel'.  And in fact, I designed the original 'rhel' platform_family to not have 'amazon' because I knew from the RHEL5/RHEL6 hybrid versions of Amazon Linux that it wasn't compatible there.

Where the platform_family of 'fedora' is useful is for the cases where the two distros do overlap.  For example you want to set the hostname by using /etc/sysconfig/network and not /etc/hostname.  In that case the platform_family of 'fedora' becomes the bucket for "I set my hostname via /etc/sysconfig/network and smell like fedora/redhat, but I can't use the EPEL repos".  And this is exactly the same kind of bucket that we have with the platform_family of "debian" which is "I set my hostname via /etc/hostname and smell like debian, but might be debian or ubuntu, or linuxmint, or raspbian", etc.

If we have a no bucket for fedora-like distros, then the case statement for the "sorta like redhat, but not RHEL" will grow without bounds which is the problem that platform_family is designed to solve.  This is simple and can be made future proof:

```
case node['platform_family']
when 'rhel'
  # recompiled RHEL that can use EPEL
when 'fedora'
  # smells like redhat, but not compatible with EPEL
when 'debian'
[...etc....]
end
```

You can still get specific:

```ruby
case node['platform_family']
when 'rhel'
  # recompiled RHEL that can use EPEL
when 'fedora'
  case node['platform']
  when 'fedora'
  when 'amazon'
  when 'nexus'
  else
  end
when 'debian'
[...etc....]
end
```

The alternative basically forces you to always be explicit:

```ruby
case node['platform_family']
when 'rhel'
  # RHEL stuff
when 'fedora'
  # fedora stuff
when 'amazon'
  # amazon
when 'nexus
  # nexus
when 'debian'
end
```

What we'll start to see will also be this:

```
case node['platform_family']
when 'rhel'
  # RHEL stuff
when 'fedora', 'amazon', 'nexus'
  # THIS IS THE PROBLEMATIC USE CASE
when 'debian'
end
```

That use case is problematic because of what happens when a new fedora-ish operating system becomes supported where the bulk of those case statements would be fine to use on the new operating system.  It becomes non future proof.  And it regenerates the same issue that we had with all the rhel-like operating systems:

```
case "fedora", "redhat", "scientific", "oracle", "amazon", "centos"
```

Where that was often the correct case statement but you'd randomly see 2 to 4 of those in community cookbooks and rarely would the author get them all correct.  What I want is a way for authors to be able to write:

```
case "fedora", "rhel"
```

And in the trivial case that gets it right, and its future proof and gives the cookbook and the author the best chance of supporting future distros without edits.  And that is what `platform_family == "rhel"` really gives you.  If you use that logic and tomorrow we support a new rhel-like platform then your cookbook works for free with no edits by you.

lamont-granquist:
An additional reply from Amazon:


    Hi Lamont,

    By the criteria that you are most interested in here, I suppose it's accurate to say that we're a Fedora-based distro from your perspective.  Our support for Python alternatives is a significant departure from Fedora / RHEL, though that should be a minor from a configuration standpoint.  It's possible to change what /usr/bin/python points to; some users will want to do that, but the rest can safely ignore it.

jonlives:
I think as a general rule I'm in favour of having ```platform_family``` be ```fedora``` providing, as is the case, that we can get more specific with ```platform```. To my mind there's a benefit to being able to write common code for fedora-like systems, and adding more specific exceptions for amazon etc when needed.

zts:
I still don't see the point in changing the platform family from "rhel" (which allows some things to work, while getting it wrong a bunch of the time) to "fedora" (which will apparently have the same behaviour).  

But as long as there's a configuration option to temporarily revert to the current value, I can't feel strongly about it.

jonlives:
I kind of see it as a case of having platform_family/platform of fedora/amazon as making it possible to try and fix that properly - in the short term it may not have much impact, but it definitely feels like a better solution in the long term. caveat: I don't use Amazon Linux, so my opinion here isn't actual-use driven.

zts:
If that's the goal, perhaps the RFC should be redrafted it make it clear.  As written, the motivations for changing it away from "rhel" would also support changing it away from "fedora".

I do use Amazon Linux, and my preference for moving it to a separate platform_family is driven by the fact that the categorisation is not reliable, and I've been burned too many times by wishful thinking to the contrary.  If there's a plan to make that better, then awesome.

jonlives:
@zts you don't feel that having the separate ```platform``` option for amazon would be enough? My thinking was that this would hit the middle ground between being able to say ```platform_family fedora``` to catch the stuff that *is* the same, and ```platform amazon``` for the exceptions or amazon specific stuff. 

My concern is that by *totally* separating both platform and platform_family, we're potentially pretty much breaking every community cookbook for anyone using amazon linux.

zts:
@jonlives the `platform` for Amazon Linux is already `amazon`, so there's no change there - ie, already folks can use `platform_family rhel` for the stuff that is the same, and `platform amazon` for the exceptions.  That's why I don't see `s/rhel/fedora/` making much of a difference to anything - it might break some things that currently work, or fix some things that currently don't, but it will largely be the same.

I agree with your concern, and I think that's the crux of the issue.  We have a few groups of community cookbooks:
1. Those that don't branch on `platform_family`.
2. Those with a branch for `platform_family == 'rhel'` that currently work on Amazon.
3. Those with a branch for `platform_family == 'rhel'` that don't work on Amazon.

Group 1 is obviously unaffected by any change.
Group 2 will definitely break if we set platform_family to `amazon`.  If it's changed to `fedora` instead, they may continue working or they may break (we don't know).
Group 3 will remain broken if we set platform_family to `amazon`.  If it's changed to `fedora` instead, some things may begin working (although I can't imagine what).

The user experience for Group 3 is arguably better (it becomes clear that something isn't supported yet because amazon != rhel), but whether that's enough to justify the disruption to Group 2 is hard to argue without knowing how many cookbooks are in Group 2 to begin with.

jonlives:
@zts gotcha - the problem is though, as far as I can tell Amazon is only going to diverge further and further from RHEL compatibility - in their own words, they could be more properly considered a fedora derivative.

My feeling is that this kind of falls under potential short term pain (as yet unquantified) for long term gain (so we don't always have this uncertainty of how much ```rhel``` stuff works on amazon).

How about if we, as was suggested earlier, made the change to platform_family fedora but included an option to keep the old behaviour for those that run into issues?

I kinda feel like keeping the existing inaccuracy just because we already have it in place has the potential to bite us further down the line.

zts:
The notion that Amazon Linux is (or will be) diverging away from RHEL _towards_ Fedora doesn't tally with my experience.  But if there's a option (for at least a few releases) to switch back to the old value to ease transition, I wouldn't vote against it.

jonlives:
I think it's more that they consider themselves closer to fedora than rhel :p

jonlives:
Summarising an offline convo with @zts, as long as there's a config option to override platform_family for those who ```fedora``` causes an issue for.

As @zts pointed out, an option to set the platform_family you want to be would also be handy for testing this sort of thing out in future.

lamont-granquist:
By putting it in the "fedora" bucket it does not mean that Amazon is becoming more like fedora.

People please read this and really understand it:  the platform_family concept becomes USELESS if we adopt the standard that platform_families members must be nearly identical.  The "rhel" platform_family is NOT the prototypical family by which the concept should be defined.  If you go down that road you're going to destroy the platform_family concept.

The prototypical platform_family is "debian" which contains platforms of "debian", "ubuntu", "linuxmint", "raspbian", etc.  If you look at those they are very different distros.  You can't use the same repositories for them, ubuntu has all kinds of added special sauce, ubuntu was the only one which used upstart, etc.  Yet it still works to have ubuntu and debian under the same platform_family because there are certain commonalities like using /etc/hostname to set the hostname and /etc/network/interfaces to set the ip address.

Similarly, everything in the "fedora" platform_family should be RPM-based yum/DNF distros that use /etc/sysconfig/network to set the hostname and /etc/sysconfig/network-scripts/ifcfg-ethN to set the ip address.

The "rhel" platform_family was designed to handle the unique case of the fedora distros that were straight recompiles of RHEL where the platform_versions were identical and where the EPEL repositories could be used without any issues.  And it was designed as a convenience to meet a need because there were a lot of fedora-ish distros that were additionally straight recompiles of RHEL so it just saved a lot of typing and creating some future proofing.  That was never Amazon Linux (although it worked for quite some time), and now that its become apparent that its wrong all we have to do is move it BACK into the bucket that we already had for it and revert OHAI-331.

And if you think I'm wrong about this, then I suggest we broaden this RFC into eliminating all of the platform families that aren't straight recompiles, at the very least to be consistent with the logic that we're using we need platform_families of "ubuntu" now, along with "linuxmint".  Presumably "pidora" and "raspbian" can stay around although we could argue about if those are substantially different from their parent distros as well.   Of course that is reducto ad absurdum trying to point out how we're going down the road of just turning platform_family into a useless alias for platform.

lamont-granquist:
And if you disagree with me, please first explain what the platform_family concept is.  I have a pretty good answer for that, which is that its a set of distros which has some basic commonalities around system administration.  I'd expect to see similar package managers, similar service management, similar ways to set the hostname, similar /etc layout, etc.  One or more of those may also be different (e.g. the upstart service in ubuntu) but enough of them should be the same to make it a useful concept.  To be in the same platform family it needs to pass my sniff test that when I sit down at it on the console and start trying to use it I'll be thinking to myself "oh its like X, only that they changed Y and Z...".  It is necessarily an IMPRECISE concept that categorizes distros in grey ways, not in black and white and is closer to taxonomy -- so that when Ubuntu takes a mutation in the service gene and implements upstart that doesn't mean that Ubuntu and Debian still aren't in the same family.  It also is designed around the case statements that we actually write.  When we start writing long case statements, then we've probably got a platform_family (to that extent the "rhel" platform_family should exist, because it makes a given case that we wrote a lot easier to write).

So can anyone start by explaining what a platform_family is, and distinguish that from the concept of a platform?  Until you can answer that question, then you're likely going to get the answer to this specific question about the platform_family of Amazon Linux wrong.

zuazo:
I agree more with Lamont on this.

So these are the conditions for the **RedHat** family:

- RPM-based, yum, DNF.

These are the conditions for the **Fedora** family:

- It looks like a RedHat, but:
  - Cannot use `yum-epel`.
  - Setups the hostname in */etc/sysconfig/network*.
  - Configures the network in */etc/sysconfig/network-scripts/ifcfg-ethN*.

Under these conditions, if Amazon Linux meets the Fedora family conditions, should be considered Fedora.

But I think it's important [that Ohai implements it that way](https://github.com/chef/ohai/blob/8.5.0/lib/ohai/plugins/linux/platform.rb#L136-L139) or at least to be documented [somewhere](https://docs.chef.io/dsl_recipe.html#platform-family). Because otherwise we can always end up discussing those conditions instead of the particular case we have here. So I think before deciding this proposal we require an RFC that clearly defines the families.

Other than that, I've always felt that the distinction between RHEL and Fedora was a little too specific. When I sit in front of a Fedora, I feel like I'm in a kind of a RedHat.

Searching for examples, [most use the `when "rhel", "fedora"` condition](https://github.com/search?utf8=%E2%9C%93&q=%22platform_family%22+fedora+language%3ARuby&type=Code&ref=searchresults). I feel like the */etc/sysconfig* is a rather weak condition because I think [few cookbooks should directly touch those settings](https://github.com/search?utf8=%E2%9C%93&q=%22platform_family%22+%22%2Fetc%2Fsysconfig%2Fnetwork%22+language%3ARuby&type=Code&ref=searchresults) (~100 results). But I understand the EPEL condition because I see [it has been quite useful for many cases](https://github.com/search?utf8=%E2%9C%93&q=%22include_recipe%22+epel+platform+language%3ARuby&type=Code&ref=searchresults) (~8,000 results, most use `platform_family` instead of `platform`).

*Argumentum ad GitHubiam* in all its glory. Somewhat misleading, I know :wink:

someara:
Closing this. The original motivation was Ease of Use for cookbook authors wanting to support Amazon Linux in addition to Other Platforms in reusable community cookbooks. 

The proposed solution to this was a chef-client configuration option that would let people "adjust" the platform_family of a given platform. This would result in hard to track down, head-scratching situations, here its not immediately obvious what's going on, and therefore detract from Chef's Ease of Use.

I'd rather see the exception logic in cookbooks that support Amazon Linux .

I'll revisit this next year.

-s

lamont-granquist:
Here's my final thoughts:

https://gist.github.com/lamont-granquist/460cbeb419309c4b8b31

PR #115, opened 2015-04-16
Release process RFC

btm:
I believe this is now complete.

### DIFF
--- a/new/release-process.md
+++ b/new/release-process.md
@@ -28,11 +28,22 @@ so I can react accordingly.
 
 ### Releasing
 
+#### Release Candidates
+
+Each major and minor release should have at least one week where a public RC is available
+for Chef users to test and report feedback.
+
+Patch releases should only include fixes for regressions of the current minor version.
+These releases should mostly include small changes and should be safe to release without
+a release candidate.
+
+#### Chef's Release Process
+
 * Verify CHANGELOG.md
 * Update lib/chef/version.rb Version RFC/Semver/something in another section here?
 * Create annotated git tag for the version
 * Kick of CI
-* Write a blog post
+* Write a blog post. Make it a public draft
 * push gem
 * have someone push to package cloud
 * have someone build AIX
--- a/new/release-process.md
+++ b/new/release-process.md
@@ -28,7 +28,7 @@ so I can react accordingly.
 
 ### Versioning
 
-We aim to adhere to the Semantic Version scheme published [here](http://semver.org/). However, there are existing exceptions. Most notably, we do not consistently and clearly delinate which public methods are also part of the public API and guaranteed to only change with major version releases.
+We aim to adhere to the Semantic Version scheme published [here](http://semver.org/). However, there are existing exceptions. Most notably, we do not consistently and clearly delineate which public methods are also part of the public API and guaranteed to only change with major version releases.
 
 From Semantic Versioning:
 
@@ -41,7 +41,7 @@ PATCH version when you make backwards-compatible bug fixes.
 Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
 ```
 
-Chef Client specific additions:
+Chef Client specific clarifications:
 
 ```
 Given a version number MAJOR.MINOR.PATCH, increment the:
@@ -51,7 +51,24 @@ MINOR version when you add any new features
 PATCH version when you are only fixing bugs, typically regressions
 ```
 
-FIXME: examples
+##### Prerelease versions
+
+Any prerelease must be versioned in the format MAJOR.MINOR.PATCH.PRETYPE.PREVER, where:
+
+PRETYPE is the description of the prerelease that the Releaser chooses based on the state of the prerelease, e.g. alpha, beta, rc, dev.
+PREVER is an incremented number starting at zero.
+
+For example: 12.0.0.dev.0, 12.2.0.alpha.2, 12.3.0.rc.3
+
+#### Examples
+
+Major: When changing the load order of any cookbook segments (e.g. attributes, templates), the major version number shall be incremented.
+
+Minor: When adding support to the mount provider for special filesystem types that were previously unsupported, the minor version number shall be incremented.
+
+Patch: When a minor release is made that contains a new feature to a service provider that causes the service provider to no longer work for users on an older but still supported version of the platform, the patch version number shall be implemented.
+
+Changes that require the minor or patch version number to be implemented may be included in a release that increments a higher version number without incrementing a lower version number. That is, new features that don not exist in version 1.1.0 may be released in 2.0.0 without any intermediary releases.
 
 ### Releasing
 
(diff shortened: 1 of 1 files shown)


new/release-process.md
```diff
@@ -0,0 +1,106 @@
...
+
+As a Chef user,
+I want to understand what happens when a critical bug is found,
+so I can react accordingly.
+
+## Specification
+
+### Versioning
+
+We aim to adhere to the Semantic Version scheme published [here](http://semver.org/). However, there are existing exceptions. Most notably, we do not consistently and clearly delinate which public methods are also part of the public API and guaranteed to only change with major version releases.
```
adamedx: minor: *delinate* should be *delineate*

martinisoft:
I like the versioning documentation in here. During one of the post-mortem discussions there was talk about nightly builds and doing more RC builds ahead of a release. How would the versioning support this?

btm:
@martinisoft I'm gyrating a little between writing how things currently work, and what we're still discussing over on https://github.com/chef/chef/issues/3205 regarding how things are about to work as soon as builds start coming out of the new CI/build cluster....

new/release-process.md
```diff
@@ -0,0 +1,106 @@
...
+
+From Semantic Versioning:
+
+```
+Given a version number MAJOR.MINOR.PATCH, increment the:
+
+MAJOR version when you make incompatible API changes,
+MINOR version when you add functionality in a backwards-compatible manner, and
+PATCH version when you make backwards-compatible bug fixes.
+Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
```
jaymzh: Can we include something about making sure that RC versions aren't versionly-greater than the final versions? For example, I've gotten RCs in the past that I then couldn't upgrade because 10.12.1rc1 is greater than 10.12.1. We need to make sure RPM/deb/etc. will actually be able to upgrade to the final version.

btm: added in 04cd78d

btm:
@martinisoft It looks like Chef Software wants to push towards Continuous Deployment for the client, so this will need updating soon to support the necessary versioning for that. See https://github.com/chef/chef/issues/3205. I think the people working on that should submit a PR against this RFC after it's approved to update it for whatever they come up with.

Nightlies are going out to https://packagecloud.io/chef/current and you can get them at the moment with "install.sh -n -p" but that combination of flags isn't very intuitive and that's what https://github.com/chef/chef/issues/3205 is meant to resolve.

new/release-process.md
jaymzh: (damnit, commented on the rev instead of the PR, sorry for the dupe)

This won't work - RPM will consider 12.3.0.rc.3 newer than 12.3.0. Please don't inflict that on people.

Fedora's guidelines for this are here: http://fedoraproject.org/wiki/Packaging%3aNamingGuidelines#NonNumericRelease

In Debian you can use ~ to denote these things because anything after ~ counts "negatively" - but the Fedora trick for this will work in Debian as well, so there's no need to use the ~ AFAICT (that way you can be consistent).

jonlives:
Broadly :+1: on this - I agree that we should modify this once CD of chef-client versions is actually nearer to being a thing.

btm:
@jaymzh I'm not making something new up to harm anyone, this is how it has always been. Because Chef was originally just shipped as a gem, our version specification has been built around [Gem::Version](http://ruby-doc.org/stdlib-2.0/libdoc/rubygems/rdoc/Gem/Version.html).

I'm a little reluctant to try to write a new specification for release candidates at this specific moment when we're sure that we need to redo it shortly for continuous delivery. Personally, especially since I originally started this just to document the release yank process that came out of a post-mortem, and it keeps growing. I'd rather iterate on it.

new/release-process.md
schisamo: @jaymzh The dashes (`-`) in pre-releases numbers are automatically translated to tildes (`~`) for the RPMs. This ensures the pre-release versions are sorted earlier than the associated final release. Take a look at the following for more details:
https://github.com/chef/omnibus/blob/master/lib/omnibus/packagers/rpm.rb#L472-L490

new/release-process.md
```diff
@@ -0,0 +1,121 @@
...
+* Trigger a build in CI (chef-trigger-release)
+* Write a blog post and share a public draft (e.g. a gist)
+
+Once the CI pipeline completes successfully:
+
+* Locally build and release the gems
+* Ask Chef Release Engineering to build and release AIX
+* Publish the blog post
+* Tweet the blog post
+* Email the chef and chef-dev mailing lists
```
coderanger: Not a blocker as this is accurate to the current process, but it would be nice to move the announce tasks a bit up the list as right now there is often a window where a new version is installable but not yet announced. The downside is that it would mean when the announce goes out, the new version might not be up for an hour or two, but maybe a pre-release email just to chef-dev? Something to make it slightly less confusing when Omnitruck suddenly starts serving a new version. Alternatively, kill omnitruck and add more explicit release control to packagecloud so this isn't needed :)

jonlives:
Approved for merge, take it away @chef/rfc-editors. Any refinements to version numbering etc can be submitted as PRs against this.

lamont-granquist:
The internal representation has to be a Gem::Version compliant version or code in cookbooks which checks that version will break, which would turn into a breaking change and a bit of a yak shave for lots of people if we tried to switch that.  Looks like omnibus should translate prerelease numbers into `-`'s for rpm, which is just a bug against that package.

schisamo:
Omnitruck already does the right thing WRT to translating dashes (`-`) to tildes (`~`). For example here are the RPMs and DEBs from Chef `12.3.0-rc.0` which @thommay released this week:

chef_stable_-_packages

jaymzh:
@btm  - I know you're not trying to hurt anyone. :) But RCs are (afaik) relatively new (on the non-gem side), and nightlies will be here shortly. We're specing out what that should look like here and everytime I've ever gotten an RC from Chef that I've rolled to more than a few machines it's been a nightmare to clean up.

@schisamo - yes it does, except that `~` doesn't work the same way on RPM that it does on Deb, so the RCs are still considered newer in RPM land.

The only really sane way to do this that I know of is to make rc's be `.0rc1` and make the final be `.1`

@lamont-granquist I wasn't proposing `~` to be clear, I was just pointing out it's what Debian typically does. As I mentioned... it doesn't work in RPMs anyway. I think the Fedora technique is valid for RPM, Debian, and Gem.

lamont-granquist:
Okay, the piece you didn't explain was where the '-' came from.  Thom actually released `12.3.0.rc.0` and some code changed that to `12.3.0-rc.0` which then hit the code you posted which did the '~'.

jaymzh:
Oh. According to that comment, @schisamo, and the attached URL, `~` is supported in newer RPM. That works with me. Thanks for enlightening me :)

schisamo:
@jaymzh No worries. I ensured this issue was fixed about a year ago when your team brought it up. :)

@lamont-granquist The `12.3.0.rc.0` is translated into `12.3.0-rc.0` because the Chef Omnibus project gets it's version from the Chef dependeancy and outputs a SemVer compliant version. Take a look here for more details: https://github.com/chef/omnibus-chef/blob/master/config/projects/chef.rb#L23-L29

jaymzh:
@schisamo remind me to buy you a beer or 3!

btm:
rebased into 08ec44511c4d68b83e33a2d7300749186e2ec603 and accepted as RFC047

PR #116, opened 2015-04-17
Add the ability to declare resources anywhere

jkeiser:


new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+
+Accessing top-level objects like `File` are one of the big reasons for this proposal.  Anything that extends from `Chef::Resource` *will* look up `Chef::Resource::X` before `::X` when a method types `X`.
+
+To fix this, we will move resources and providers in `Chef::Provider::*` and `Chef::Resource::*` into `Chef::Resources` (and use the inline syntax as well).  For addressing the obvious concerns, see the Errors and Deprecations section.
+
+### Removing `method_missing`
+
+Presently, when you type `my_resource` in the DSL, there is no actual method by that name; we intercept `method_missing`, perform a string transformation on the method name to get `MyResource`, and then try to get the `Chef::Resource::MyResource` class.  This causes confusing stack traces when the resource cannot be looked up, and is significantly more difficult to read.
+
+In this new world, a `Chef::Resources.my_resource` method is created, and `Chef::Resources` is mixed into the recipe DSL, so there is a real method.  `method_missing` never gets involved.  The methods generated look like this:
```
jaymzh: So. Much. Win.

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+  created_at ||= caller[0]
+  declare_resource(:my_resource, name, created_at, ::MyResource, &block)
+end
+```
+
+NOTE: This does not remove *all* of method_missing; in particular, we need to make definitions do something similar.  Add that to the proposal.
+
+### Declaring a Provider inline (Also Single-File Resources)
+
+In order to create a Resource, you must have a corresponding Provider.  The current algorithm takes `Chef::Resource::MyResource`, lops off `Chef::Resource::` entirely, and prepends `Chef::Provider::`.  This is The new algorithm will first look for a `Chef::Resource::MyResource::Provider` class, so that declaring the provider for a class will look like this for most people:
```
jaymzh: s/This is//

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+  class Provider < Chef::Provider
+    ...
+  end
+end
+```
+
+This has several nice benefits:
+- Supports the goal of resources outside of `Chef::Resource`.
+- Encourages the creation of providers in the same file as the resource.
+- An absolute minimum of extra typing when linking a resource to a provider.
```
jaymzh: how does this work if you want, say, 2 possible providers?

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+
+The deprecation warnings for `Chef::Resource::File` suffice for Chef 12 deprecation purposes.
+
+Now for the reverse compatibility part: we want `File` to work.  We will allow anyone who declares a resource *outside* of Chef to type `File` normally by including Chef::Resource::TopLevelConstants directly into their resource.  Whenever a resource is created in `Chef::Resource`, we will check if there is a top level constant with the same name, and add it.  Since only resources *outside* of `Chef::Resource` get this included, anything inside `Chef::Resource` will still refer to the original thing.
+
+LWRPs created via `resources/x.rb` in a cookbook are also exempt from `TopLevelConstants`.
+
+#### Name Conflicts
+
+Resources can try to take the same name in recipe DSL (such as when two cookbooks both implement a similar thing).  Fact of life.  Presently when two custom resources try to take the same name, presently they will do one of two unsatisfactory things:
```
jaymzh: presently is twice in that sentence...

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+
+- Raise a confusing "subclassing the wrong class" error on load; or
+- Silently overwrite and merge with one another, with the last one winning
+
+Now, when you are in separate namespaces, the first thing simply won't happen.
+
+We will avoid the second thing by checking whether the method `my_resource` is already defined, and if it is, we
```
jaymzh: did the sentence drop off there, or is the next section supposed to be the end of that sentence? If the later, perhaps a `...` or `:`

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+### Risks To Specific Software
+
+While we're trying to maintain full compatibility, there's risk here, particularly for people doing metaprogramming of resources.  We will run the tests of the following packages, and make sure current versions
+still work (possibly emitting deprecation warnings):
+
+- chef-sugar
+- chefspec
+- chef-rewind
+- Poise
+- Crazytown
```
jaymzh: can you add foodcritic? It digs into internals fairly heavily. CC @odcinek and @acrump

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+
+    As a <<user_profile>>,
+    I want to <<functionality>>,
+    so that <<benefit>>.
+
+## Specification
+
+### Resources outside of `Chef::Resource`
+
+Users who wish to register a resource named `my_resource` can presently do it by creating a class inside the `Chef::Resource` namespace and another in `Chef::Provider`.  Chef then uses magical namespace lookups to
```
lamont-granquist: The provider resolver and resource resolver in Chef 12 actually already accomplish this.  There's still a fallback to magic for stuff that hasn't been fully converted yet, but if that was done then the magical namespace lookups would all go away, along with ripping out the old platform mapping.  Classes still have to inherit from Chef::Resource or Chef::Provider, but they can be named anything.

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Accessing `File`
+
+Accessing top-level objects like `File` are one of the big reasons for this proposal.  Anything that extends from `Chef::Resource` *will* look up `Chef::Resource::X` before `::X` when a method types `X`.
+
+To fix this, we will move resources and providers in `Chef::Provider::*` and `Chef::Resource::*` into `Chef::Resources` (and use the inline syntax as well).  For addressing the obvious concerns, see the Errors and Deprecations section.
+
+### Removing `method_missing`
+
+Presently, when you type `my_resource` in the DSL, there is no actual method by that name; we intercept `method_missing`, perform a string transformation on the method name to get `MyResource`, and then try to get the `Chef::Resource::MyResource` class.  This causes confusing stack traces when the resource cannot be looked up, and is significantly more difficult to read.
```
lamont-granquist: That's not quite how it works right now with the resource resolver indirection, but I don't think it changes anything that you're doing.

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+
+The mechanism is:
+- An `inherited` hook on `Chef::Resource` automatically takes care of hooking up `MyResource` to all recipe DSL when the class is created.
+- The provider lookup for a resource looks first for a `Provider` class inside the Resource before moving on to other things.
+
+### Accessing `File`
+
+Accessing top-level objects like `File` are one of the big reasons for this proposal.  Anything that extends from `Chef::Resource` *will* look up `Chef::Resource::X` before `::X` when a method types `X`.
+
+To fix this, we will move resources and providers in `Chef::Provider::*` and `Chef::Resource::*` into `Chef::Resources` (and use the inline syntax as well).  For addressing the obvious concerns, see the Errors and Deprecations section.
```
lamont-granquist: This will break a ton of people.  The chef classes are commonly inherited from and the provider parameter to resources exposes the class name as a public API (highly unfortunately).  We probably should rewrite the existing chef classes as new resources, but we probably need to keep the existing class layout.  The disruption from changing that would likely be of the order of the Chef 11 attributes changes or maybe worse.  Making those changes just to fix `::File` (which is fixed in a lot of use cases like in not_if blocks) probably isn't compelling enough to break everything this badly.

new/resources-anywhere.md
lamont-granquist: Actually the algorithm doesn't quite look like that.   There's a fallback to that kind of logic, but most of the Chef 12 providers are using the provider resolver code now, and the first fallback is to the static platform mapping hash, and only after both of those fail does it throw up its hands and go looking for a class derived from the resource name.

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+
+Sometimes the name of your class isn't the name that you need.  In that case, you can change the DSL with `chef_recipe_dsl`.
+
+```ruby
+class SuperMountPointResource < Chef::Resource
+  chef_recipe_dsl name: 'mount_point'
+end
+```
+
+This will remove the `super_mount_point_resource` DSL and add `mount_point` as DSL.
```
lamont-granquist: I think you're colliding with provider_resolver and resource_resolver functionality that we already have here.

```ruby
class Chef
  class Resource
    class WindowsPackage < Chef::Resource::Package

      provides :package, os: "windows"
      provides :windows_package, os: "windows"
```

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+
+- All things that currently work in Chef 12 must continue to work in Chef 12.
+- 99.99% of things that currently work in Chef 12, must continue to work in Chef 13.
+- As much as possible should work in 12.X, so that people in Chef 12 have a story when we move to 13 and so people can begin converting now.
+
+#### Accessing `Chef::Resource::File` or `Chef::Provider::File` directly
+
+In Chef 13, core resources are simply moving out of the Resource and Provider namespaces.  However, in Chef 12, there will be resources and providers who extend from them or construct them directly.
+
+In Chef 12, we will create classes in `Chef::Resource` extending from the real Resources which emit a deprecation warning on initialize or subclassing (`class Chef::Resource::File < Chef::Resources::File ...`).
```
lamont-granquist: Still think this is a lot of work we're asking people to do, just in order to be able to get around the `::File` issue

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+
+Now, when you are in separate namespaces, the first thing simply won't happen.
+
+We will avoid the second thing by checking whether the method `my_resource` is already defined, and if it is, we
+
+#### Replaced Classes
+
+It is possible to replace a class in Ruby (you can simply go assign ).  Due to the way this is implemented, the new class will in fact be used (which is sort of what a user would expect).  However, it's janky and something we'd like to warn about.  When the user attempts to use the resource (by typing `my_resource`), we will check in `build_resource` that the class has not changed from what was originally registered, and emit a warning that the class isn't what we expect and that they should contact the author of the new class to get the warning fixed.
+
+To *legitimately* do this, you should use `chef_recipe_dsl` instead.
```
lamont-granquist: We already stopped doing this with Chef 12 and stopped undefining the constants and recreating classes.  This feels like we're reintroducing the same problem again.  Classes need to *not* be redefinable.  Allowing them to be redefined leads to tons of memory leaks in chefspec which runs mutiple converges rapidly in the same process.

I think this is also a solved problem with the resource resolver.   You can have multiple classes declare that they provide the same resource with `provides` lines, and then the user can determine via the resource_priority_map which class wins if there's a conflict.

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+
+As the resources are moving, *requires* are also moving.  In Chef 12, we will emit a deprecation warning when you require 'chef/resource/x' instead of 'chef/resources/x', but it will still work.
+
+#### Redefining `dsl_name`
+
+Redefining `dsl_name` is deprecated and will not work in 13.  In Chef 12, the base `Chef::Resource` will include a `method_defined` hook that lets us detect this, cause the DSL to change, and emit a warning.
+
+#### Creating anonymous resource classes
+
+This procedure: `Chef::Resource::MyResource = Class.new(Chef::Resource) { ... }` could be used to create a resource currently and `my_resource` would work fine in recipes.  In Chef 13, that will no longer work, because we don't do anything magical for `Chef::Resource`.  In Chef 12, we will still check for `Chef::Resource::MyResource` when `method_missing(:my_resource)` is called, and if it is found, will return it and emit a deprecation warning.
```
lamont-granquist: I tend to like this idea, but I think that Chef::Resource::File needs to remain even in Chef 13.  I can see LWRPs being anonymous classes, but think there needs to be an API to explicitly set the class so that normal ruby inheritance can override it.

lamont-granquist:
So generally:

* I'm very :+1: on eliminating method_missing and define_method'ing resources.
* More work needs to be done on integration with the provider_resolver + resource_resolver.  It seems like there's a lot of solved problems in Chef 12 that you're trying to solve again.
* I don't file the `::File` argument to be compelling enough to warrant some of the larger changes
* I don't like the idea of turning the internal providers and resolvers into anonymous classes because of the scope of the fallout to that
* I do like the idea of LWRPs being anonymous classes by default, with an API to set the classes explicitly

coderanger:
To be clear, this is already supported, and Poise already includes support for anonymous classes (only a [minor change](https://github.com/poise/poise/blob/master/lib/poise/helpers/resource_name.rb#L57-L62)). See most of my recent cookbooks for examples of non-Chef-namespace code, eg https://github.com/poise/poise-service/blob/master/lib/poise_service/user_resource.rb#L22-L39.

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+end
+```
+
+### Creating a DSL resource with a different name
+
+Sometimes the name of your class isn't the name that you need.  In that case, you can change the DSL with `chef_recipe_dsl`.
+
+```ruby
+class SuperMountPointResource < Chef::Resource
+  chef_recipe_dsl name: 'mount_point'
```
coderanger: How is this different from what we have now?

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+end
+```
+
+This will remove the `super_mount_point_resource` DSL and add `mount_point` as DSL.
+
+### Single File LWRPs
+
+Since HWRPs are now easily declared in a single file, it would be macabre of us not to allow single file LWRPs as well.  We propose two new syntaxes in LWRP resources to enable this:
```
coderanger: Poise implements this as the Fused helper.

new/resources-anywhere.md
```diff
@@ -0,0 +1,211 @@
...
+
+#### Accessing `Chef::Resource::File` or `Chef::Provider::File` directly
+
+In Chef 13, core resources are simply moving out of the Resource and Provider namespaces.  However, in Chef 12, there will be resources and providers who extend from them or construct them directly.
+
+In Chef 12, we will create classes in `Chef::Resource` extending from the real Resources which emit a deprecation warning on initialize or subclassing (`class Chef::Resource::File < Chef::Resources::File ...`).
+
+#### Accessing `File` from a resource or provider method
+
+Right now, when you type `File` in any resource or provider it always refers to `Chef::Resource::File` or `Chef::Provider::File`.  In Chef 13, it will refer to `::File`.
```
coderanger: This will require renaming the `File` class, anything that inherits from `Chef::Resource` gets the cranky symbol lookup.

new/resources-anywhere.md
coderanger: This doesn't seem related to the rest of this RFC, and rebuilding the internals of the DSL is a bigger topic.

coderanger:
Also check out https://github.com/poise/halite/blob/master/lib/halite/spec_helper.rb#L289-L337 for an example of using anon classes with Chef. It still patches them in to the globals but that shouldn't be required any more. The problem on that front is that NodeMap doesn't offer a delete/remove API.

lamont-granquist:
Offtopic, but what shape should the NodeMap delete/remove API be?  And what are the use cases?  I noted that it would most likely be needed, but skipped it because I didn't have a use case and didn't want to design against requirements I didn't have.

coderanger:
@lamont-granquist https://github.com/poise/halite/blob/master/lib/halite/spec_helper.rb#L150-L169 except instead of const_set and const_remove I would be adding/removing from the Chef::Resource NodeMap around each test.

new/resources-anywhere.md
jkeiser: Thanks!  I'll update.  It will still be super nice for stack traces and method_missingness if we can do the wire-up once (also for performance and readability) to actual methods.

new/resources-anywhere.md
jkeiser: There is a method you can override on resource to get the provider; I will make sure and add a section on how that works now and how it would work.  I think the change for the default behavior is still worth changing that.

jkeiser: Turns out `provides :blah` will work the same before and after.  I've added a very minor note, but there are no deprecations or impacts involved there, so all is well :)

jkeiser:
@lamont-granquist has convinced me that moving the resources out of `Chef::Resource` is too high a cost; I'm going to retool the proposal around that, which significantly reduces the risk.  Plan now is simply to remove method_missing and make real methods.  I'd *like* to make the recipe DSL scope-able to chef run / cookbook, but I'll save that for the next RFC to keep this simple.

jkeiser:
(I am adding definitions into the proposal too, so that I can *actually* remove method_missing.

ranjib:
This is awesome :-) . Extending chef will be lot easier with this. I'll +1 pretty much everything @lamont-granquist mentioned. I was wondering, may be a Chef::LWRPBase can be used to build a single class resource+provider entity.

coderanger:
I'm a bit confused. The original premise of this RFC was being able to declare resources outside of the Chef namespace, a feature which already works fine. It is now pivoting to be about reworking the internals of the DSL implementation? If so, we should probably rename things and remove everything not about reworking the DSL. Given the current NodeMap system I'm not sure how much value there is in removing the use of method_missing, but regardless this is a bit all over the place.

jkeiser:
Yes, when I say I am going to retool it, that is what I mean.

coderanger:
:+1: Should we close out this PR and open a new one so there is less noise on the revamped RFC?

jkeiser:
Happy to close it out.  I appreciate the comments, and have taken everything that seemed actionable into account.  Will post a new one shortly.

PR #117, opened 2015-04-19
Remove method_missing and make Chef::Resource unspecial

jkeiser:
- Use explicit methods for Chef DSL instead of method_missing
- Remove special treatment of classes that happen to be under the Chef::Resource namespace

cc @coderanger @lamont-granquist

### DIFF
--- a/new/missing-method-missing.md
+++ b/new/missing-method-missing.md
@@ -13,6 +13,20 @@ This proposal:
 - Removes all special treatment of classes in the Chef::Resource and Chef::Provider namespaces
 - Automatically adds resource DSL for all named descendants of Chef::Resource, no matter what their namespace
 
+## Compatibility
+
+In Chef 12, three behavior changes will be noted:
+
+- Stack traces for resource declarations will usually no longer contain
+  `method_missing` (and will instead contain the name of the resources).
+- Classes outside of the Chef::Resource namespace will now be placed in recipe
+  DSL automatically.
+- Warnings are issued for deprecated behavior that will change in Chef 13 (see
+  Deprecations section).
+
+In Chef 13, deprecated behavior will be removed (along with method_missing).  All
+of these have deprecation warnings enabled for Chef 12 (see Deprecations section).
+
 ## Motivation
 
     As a Chef user,
@@ -61,6 +75,47 @@ messages for cookbook conflicts such as "superclass mismatch for class MyResourc
 Chef will now automatically call `provides` on all new, named resource classes,
 using `ResourceClass.dsl_name` (a transform of the class name, i.e.
 MyPackage::MyResource -> my_resource) as the provided class.
+
+```ruby
+class MyPackage::MyResource < Chef::Resource
+end
+# now you can type 'my_resource' in a recipe.
+```
+
+#### Explicit vs. Implicit `provides`
+
+If you specify `provides :something_else` in a class, that will override the
+auto-class-creation behavior.
+
+```ruby
+class MyPackage::MyResource < Chef::Resource
+  provides :shazam
+end
+# Now `shazam` will work fine in a recipe, but `my_resource` will not.
+```
+
+If you want to still provide the original name, you must be explicit:
+
+```ruby
+class MyPackage::MyResource < Chef::Resource
+  provides :my_resource
+  provides :shazam
+end
+# Now both `shazam` and `my_resource` work in a recipe
+```
+
+#### `does_not_provide`
+
+If you want to (for example) create a base class that does not have an actual
+DSL method, you can specify `does_not_provide` or `does_not_provide :my_name` in
+the class:
+
+```ruby
+class MyPackage::MyResourceBase < Chef::Resource
+  does_not_provide
+end
+# my_resource_base will not exist in recipe DSL
+```
 
 ### Move LWRPs out of the Chef::Resource namespace
 
--- a/new/missing-method-missing.md
+++ b/new/missing-method-missing.md
@@ -1,0 +1,156 @@
+---
+RFC: unassigned
+Author: John Keiser <john@johnkeiser.com>
+Status: Draft
+Type: Standards Track
+---
+
+# Remove magic and method_missing from Chef::Resource lookup
+
+This proposal:
+
+- Switches to explicit methods for Chef recipe DSL instead of method_missing
+- Removes all special treatment of classes in the Chef::Resource and Chef::Provider namespaces
+- Automatically adds resource DSL for all named descendants of Chef::Resource, no matter what their namespace
+
+## Motivation
+
+    As a Chef user,
+    I want resources declared in my own namespace to be in recipe DSL automatically,
+    so that I don't have to write the name of the resource twice.
+
+    As a Chef user,
+    I want to declare resources in my own namespace,
+    So that I get better error messages when there is a resource conflict.
+
+    As a Chef developer,
+    I want meaningful names in stack traces for methods,
+    So that I can actually read the debug output.
+
+    As a Chef developer,
+    I want less if statements, case statements and metaprogramming in DSL method calls,
+    So that I can debug and follow logic more easily.
+
+## Specification
+
+### Create explicit DSL methods for resources
+
+When looking up resources, there is no method anywhere that corresponds to the
+actual resource name.  This can lead to a lot of unnecessarily heavy thinking
+when following some very typical stack traces.  It also means that the name of
+the resource being declared is often not on the stack, making debugging harder.
+
+Here we propose that all resource and definition DSL be added to
+`Chef::DSL::Recipe` as actual methods, as they are created, so that they can be
+subject to inspection and the stack traces can be followed.  This will happen
+automatically when the user declares resources and definitions.
+
+### Automatically add new resources and providers to DSL, regardless of namespace
+
+Presently, when you type `my_resource` in the Chef DSL, in the absence of any
+`provides` directives to direct Chef to the correct resource class, the
+namespaces `Chef::Resource` is explicitly searched for
+`Chef::Resource::MyResource`.
+
+This means that if you have seen examples of classes in Chef::Resource, and want
+to do the obvious thing and just move to your own namespace, it doesn't
+immediately work and it is far from apparent why. The result? A massive traffic
+jam of non-core resources in the `Chef::Resource` namespace and puzzling error
+messages for cookbook conflicts such as "superclass mismatch for class MyResource."
+
+Chef will now automatically call `provides` on all new, named resource classes,
+using `ResourceClass.dsl_name` (a transform of the class name, i.e.
+MyPackage::MyResource -> my_resource) as the provided class.
+
+### Move LWRPs out of the Chef::Resource namespace
+
+With Chef::Resource no longer a special place for class lookup, LWRPs no longer
+need to be given a namespace.  We will now make resources/x.rb and providers/x.rb
+an anonymous class, with string methods to make it easy to see where they come
+from.  This avoids a number of potential errors and warnings that can happen in
+the case of conflicts, and gives us control over how we handle conflicts.
+
+Users who wish to set the class of the resource can assign it directly in the
+LWRP with `Namespace::MyResource = self`.
+
+## Deprecations and Errors
+
+Removing `method_missing` means several methods of creating resource names cannot
+be caught.  These are the cases.  We will issue deprecation warnings for each one
+and remove them (and method_missing) come Chef 13.
+
+### Chef::Resource::MyResource = <resource>
+
+When you assign an anonymous class or a resource from another namespace into
+`Chef::Resource`, your new resource will currently be accessible via the DSL.
+Since there is no Ruby method that lets us catch the assignment of a constant,
+there is no way for us to assign an explicit DSL method to handle the resource.
+
+```ruby
+# Deprecated
+Chef::Resource::Blah = Class.new(Chef::Resource)
+```
+
+We will catch this in method_missing, issue a warning letting the user know they
+should use the `provides` directive on their class to get it into the DSL, and remove it come Chef 13.
+
+```ruby
+# Not deprecated (though the assignment becomes unnecessary)
+Chef::Resource::Blah = Class.new(Chef::Resource) { provides :blah }
+```
+
+### Calling `method_missing` directly
+
+If people call `method_missing` directly to invoke DSL methods, and we don't
+define it anymore, it obviously won't actually invoke the methods.  We will
+detect this in method_missing, issue a warning telling people to call the
+actual method instead or use public_send(), and remove it come Chef 13.
+
+### Custom `provides?` without corresponding `provides :name`
+
+If someone creates a resource class with a custom `provides?` method, they can
+currently
+
+```ruby
+class MyResource < Chef::Resource
+  def provides?(node, name)
+    name == :blah
+  end
+end
+```
+
+The implications of this are that we have to scan every single resource when
+you type `blah` in a recipe, just so we can catch this one.  In Chef 12 we will
+issue a warning and tell you to do this instead:
+
+```ruby
+class MyResource < Chef::Resource
+  provides :blah
+end
+```
+
+### Referencing Chef::Resource::MyLWRPResource
+
+When you create an LWRP, it is presently placed in
+`Chef::Resource::CookbooknameResourcename`.  In the spirit of not placing user
+defined stuff into the core Chef namespace, this will no longer happen.  Any
+user that calls `Chef::Resource::MyLwrpResource.new` or extends directly from
+the name `Chef::Resource::MyLwrpResource` will receive a warning.
+
+## Risks To Specific Software
+
+While we're trying to maintain full compatibility, there's risk here, particularly for people doing metaprogramming of resources.  We will run the tests of the following packages, and make sure current versions
+still work (possibly emitting deprecation warnings):
+
+- chef-sugar
+- chefspec
+- chef-rewind
+- Poise
+- Crazytown
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.
--- a/new/missing-method-missing.md
+++ b/new/missing-method-missing.md
@@ -81,17 +81,6 @@ automatically do this as well.
 When resource DSL becomes explicit, the idea that a class has a single DSL name
 no longer makes sense.  We will remove `self.dsl_name` from `Resource` and `self.resource_name` from `Chef::Resource::LWRPBase`.
 
-### `provides_nothing`
-
-If you want to turn off method DSL for a class, you can call `provides_nothing`:
-
-```ruby
-class MyPackage::MyResourceBase < Chef::Resource
-  provides_nothing
-end
-# my_resource_base will not exist in recipe DSL
-```
-
 ### Move LWRPs out of the Chef::Resource namespace
 
 With Chef::Resource no longer a special place for class lookup, LWRPs no longer
--- a/new/missing-method-missing.md
+++ b/new/missing-method-missing.md
@@ -1,8 +1,10 @@
 ---
 RFC: unassigned
 Author: John Keiser <john@johnkeiser.com>
-Status: Draft
+Status: Final
 Type: Standards Track
+Tracking:
+  - https://github.com/chef/chef/pull/3270
 ---
 
 # Remove magic and method_missing from Chef::Resource lookup
--- a/new/missing-method-missing.md
+++ b/new/missing-method-missing.md
@@ -59,33 +59,11 @@ Here we propose that all resource and definition DSL be added to
 subject to inspection and the stack traces can be followed.  This will happen
 automatically when the user declares resources and definitions.
 
-### Automatically add new resources and providers to DSL, regardless of namespace
+### Stop automatically adding resources to recipe DSL
 
-Presently, when you type `my_resource` in the Chef DSL, in the absence of any
-`provides` directives to direct Chef to the correct resource class, the
-namespaces `Chef::Resource` is explicitly searched for
-`Chef::Resource::MyResource`.
-
-This means that if you have seen examples of classes in Chef::Resource, and want
-to do the obvious thing and just move to your own namespace, it doesn't
-immediately work and it is far from apparent why. The result? A massive traffic
-jam of non-core resources in the `Chef::Resource` namespace and puzzling error
-messages for cookbook conflicts such as "superclass mismatch for class MyResource."
-
-Chef will now automatically call `provides` on all new, named resource classes,
-using `ResourceClass.dsl_name` (a transform of the class name, i.e.
-MyPackage::MyResource -> my_resource) as the provided class.
-
-```ruby
-class MyPackage::MyResource < Chef::Resource
-end
-# now you can type 'my_resource' in a recipe.
-```
-
-#### Explicit vs. Implicit `provides`
-
-If you specify `provides :something_else` in a class, that will override the
-auto-class-creation behavior.
+Presently, when you create a class under `Chef::Resource`, a magical Chef DSL
+based on the class name (`my_resource`) will automatically work.  Now you will
+need to use `provides` to explicitly mark the DSL you provide:
 
 ```ruby
 class MyPackage::MyResource < Chef::Resource
@@ -94,25 +72,22 @@ end
 # Now `shazam` will work fine in a recipe, but `my_resource` will not.
 ```
 
-If you want to still provide the original name, you must be explicit:
+We will ensure that all existing core Chef classes in the `chef`, `cheffish`,
+and `chef-provisioning*` gems use `provides` to avoid warnings.  LWRPs will
+automatically do this as well.
 
-```ruby
-class MyPackage::MyResource < Chef::Resource
-  provides :my_resource
-  provides :shazam
-end
-# Now both `shazam` and `my_resource` work in a recipe
-```
+### Remove the `dsl_name` and `resource_name` class methods
 
-#### `does_not_provide`
+When resource DSL becomes explicit, the idea that a class has a single DSL name
+no longer makes sense.  We will remove `self.dsl_name` from `Resource` and `self.resource_name` from `Chef::Resource::LWRPBase`.
 
-If you want to (for example) create a base class that does not have an actual
-DSL method, you can specify `does_not_provide` or `does_not_provide :my_name` in
-the class:
+### `provides_nothing`
+
+If you want to turn off method DSL for a class, you can call `provides_nothing`:
 
 ```ruby
 class MyPackage::MyResourceBase < Chef::Resource
-  does_not_provide
+  provides_nothing
 end
 # my_resource_base will not exist in recipe DSL
 ```
@@ -134,25 +109,11 @@ Removing `method_missing` means several methods of creating resource names canno
 be caught.  These are the cases.  We will issue deprecation warnings for each one
 and remove them (and method_missing) come Chef 13.
 
-### Chef::Resource::MyResource = <resource>
+### Chef::Resource::MyResource without `provides`
 
-When you assign an anonymous class or a resource from another namespace into
-`Chef::Resource`, your new resource will currently be accessible via the DSL.
-Since there is no Ruby method that lets us catch the assignment of a constant,
-there is no way for us to assign an explicit DSL method to handle the resource.
-
-```ruby
-# Deprecated
-Chef::Resource::Blah = Class.new(Chef::Resource)
-```
-
-We will catch this in method_missing, issue a warning letting the user know they
-should use the `provides` directive on their class to get it into the DSL, and remove it come Chef 13.
-
-```ruby
-# Not deprecated (though the assignment becomes unnecessary)
-Chef::Resource::Blah = Class.new(Chef::Resource) { provides :blah }
-```
+When you have a class in the `Chef::Resource` namespace without `provides`, the
+DSL will no longer work in Chef 13.  In 12 we will issue a deprecation warning
+whenever the user tries to use a DSL class without corresponding DSL.
 
 ### Calling `method_missing` directly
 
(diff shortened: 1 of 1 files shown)


jkeiser:
Note: most of the deprecation feedback is based on edges discovered during the implementation at https://github.com/chef/chef/tree/jk/missing_method_missing .  That branch still needs to handle Chef::Provider and add a number of deprecation (and stack trace) tests.

new/missing-method-missing.md
```diff
@@ -0,0 +1,156 @@
...
+Type: Standards Track
+---
+
+# Remove magic and method_missing from Chef::Resource lookup
+
+This proposal:
+
+- Switches to explicit methods for Chef recipe DSL instead of method_missing
+- Removes all special treatment of classes in the Chef::Resource and Chef::Provider namespaces
+- Automatically adds resource DSL for all named descendants of Chef::Resource, no matter what their namespace
```
coderanger: I would rather never create them for descendants per se, just should be any class that calls Chef::Resource.provides, no matter what that class is. We already have enough issues with descendant tracking for providers as is :-( I think that's basically what you mean though.

jkeiser: I'm actually just using `def self.inherited` on `Chef::Resource`, which seems to do the job.  I'd like descendant tracking to go away, but we have to do this in chess moves, or we get RFCs like the last one :)

new/missing-method-missing.md
```diff
@@ -0,0 +1,156 @@
...
+## Specification
+
+### Create explicit DSL methods for resources
+
+When looking up resources, there is no method anywhere that corresponds to the
+actual resource name.  This can lead to a lot of unnecessarily heavy thinking
+when following some very typical stack traces.  It also means that the name of
+the resource being declared is often not on the stack, making debugging harder.
+
+Here we propose that all resource and definition DSL be added to
```
coderanger: Need to specify when this happens exactly and what the life cycle of these methods will be. Presumably `Chef::Resource.provides` will call `define_method` and we'll need to add an explicit method to unregister a resource.

new/missing-method-missing.md
```diff
@@ -0,0 +1,156 @@
...
+namespaces `Chef::Resource` is explicitly searched for
+`Chef::Resource::MyResource`.
+
+This means that if you have seen examples of classes in Chef::Resource, and want
+to do the obvious thing and just move to your own namespace, it doesn't
+immediately work and it is far from apparent why. The result? A massive traffic
+jam of non-core resources in the `Chef::Resource` namespace and puzzling error
+messages for cookbook conflicts such as "superclass mismatch for class MyResource."
+
+Chef will now automatically call `provides` on all new, named resource classes,
```
coderanger: We should support naming the class whatever you want as long as you explicitly call `provides()` yourself, which means it can't be called automatically on subclassing.

new/missing-method-missing.md
jkeiser: As far as the everyone vs. no one decision, here are my thoughts:

1. I agree with you from a cleanliness standpoint: everyone should call `requires`
2. From a practical standpoint this would make a lot of resources stop working
3. I think that, as you have proposed in the past, the real Resource base should be a mixin that we will create later.  That is the point at which we get to do things that really cut backcompat deeply, because it won't affect existing Chef::Resource derived stuff.

new/missing-method-missing.md
```diff
@@ -0,0 +1,156 @@
...
+using `ResourceClass.dsl_name` (a transform of the class name, i.e.
+MyPackage::MyResource -> my_resource) as the provided class.
+
+### Move LWRPs out of the Chef::Resource namespace
+
+With Chef::Resource no longer a special place for class lookup, LWRPs no longer
```
coderanger: To be clear, this will not stop the collisions on the `File` const. :+1: still, but this don't stop those particular errors unless we also rename the `File` classes in Chef.

new/missing-method-missing.md
```diff
@@ -0,0 +1,156 @@
...
+the case of conflicts, and gives us control over how we handle conflicts.
+
+Users who wish to set the class of the resource can assign it directly in the
+LWRP with `Namespace::MyResource = self`.
+
+## Deprecations and Errors
+
+Removing `method_missing` means several methods of creating resource names cannot
+be caught.  These are the cases.  We will issue deprecation warnings for each one
+and remove them (and method_missing) come Chef 13.
```
coderanger: A change of this magnitude probably shouldn't go in until Chef 13 at all, so the full deprecation cycle would extend until Chef 14.

new/missing-method-missing.md
jkeiser: An excellent point, and I'll modify the RFC, but here's the answer:

`Chef::Resource.provides` calls `define_method` (actually, for technical reasons involving making the stack trace actually have the name in it, it does class_eval "def #{method} ...").  To remove a resource, we need to implement an `unprovide` somewhere.

new/missing-method-missing.md
```diff
@@ -0,0 +1,156 @@
...
+Since there is no Ruby method that lets us catch the assignment of a constant,
+there is no way for us to assign an explicit DSL method to handle the resource.
+
+```ruby
+# Deprecated
+Chef::Resource::Blah = Class.new(Chef::Resource)
+```
+
+We will catch this in method_missing, issue a warning letting the user know they
+should use the `provides` directive on their class to get it into the DSL, and remove it come Chef 13.
```
coderanger: As above, should say Chef 14.

new/missing-method-missing.md
```diff
@@ -0,0 +1,156 @@
...
+
+```ruby
+class MyResource < Chef::Resource
+  def provides?(node, name)
+    name == :blah
+  end
+end
+```
+
+The implications of this are that we have to scan every single resource when
```
coderanger: This should be called out as a compat break, and is part of why this can't go in before Chef 13 at a minimum.

new/missing-method-missing.md
```diff
@@ -0,0 +1,156 @@
...
+
+```ruby
+class MyResource < Chef::Resource
+  provides :blah
+end
+```
+
+### Referencing Chef::Resource::MyLWRPResource
+
+When you create an LWRP, it is presently placed in
```
coderanger: Should add an accessor method to take the place of this. Maybe something like `Chef::Resource::LWRP.get('cookbook', 'name')`?

new/missing-method-missing.md
```diff
@@ -0,0 +1,156 @@
...
+## Risks To Specific Software
+
+While we're trying to maintain full compatibility, there's risk here, particularly for people doing metaprogramming of resources.  We will run the tests of the following packages, and make sure current versions
+still work (possibly emitting deprecation warnings):
+
+- chef-sugar
+- chefspec
+- chef-rewind
+- Poise
+- Crazytown
```
coderanger: Add Halite to this list, it's spec helper does a lot of this work already as it uses a very large number of anonymous resource and provider classes.

new/missing-method-missing.md
jkeiser: Agreed; I had to give that up when I agreed not to move the resource classes into a different namespace for backcompat reasons.  `File` will just have to get solved in one of several other ways, at a later time :/

new/missing-method-missing.md
coderanger: Recording the name on `.inherited` is fine, as long as the system recognizes calling provides manually later and removes the very-briefly set incorrect name. As it stands, the NodeMap code doesn't do that.

new/missing-method-missing.md
jkeiser: That's how I initially thought, but then I realized, for classes that *are* providing specific OS implementations, they could easily be relying on the current behavior.  i.e. you could do this:

```ruby
class Chef::Resource::AptPackage < Chef::Resource::Package
  provides :package, os: "linux", platform_family: [ "debian" ]
end
```

It would still be available as both `apt_package` and `package`.

The current `AptPackage` happens to behave nicely, declaring both; but it didn't have to.  If the consensus of this august body is that we are willing to take that particular backcompat hit, I'm willing to make it happen.

new/missing-method-missing.md
coderanger: Also because I have to say it, "This won't be thread safe" but I don't think anyone cares about multi-threading the initial cookbook loading process yet, so as long as we all remember it should be fine :-)

new/missing-method-missing.md
jkeiser: I can see that sort of argument.  I'll happily flip it if that's the direction we go.  Should I just say Chef X and let people decide at the meeting?  (If the rest of this goes through.)

new/missing-method-missing.md
coderanger: All of my resources are in a class named `Something::ResourceName::Resource` so they will all otherwise map to `resource`. Could make that a special case to prevent automatic registration, but that seems icky. Another option is to use `Chef::Mixin::DescendantsTracker` and do all resource registration in a specific pass after code loading happens for each cookbook. Probably less efficient, but at that point you can see if provides was called or not.

new/missing-method-missing.md
jkeiser: Except it's not a compat break, because in Chef 12 it will still work.  That's why we're calling it out and putting a deprecation warning in.  (I have already implemented it.)

jkeiser: Oh, you mean that I should point out that each of the things I just called deprecations are compat breaks?  Isn't that sort of implied?  I can make the change, but just in case this comment was due to me not making myself clear enough, you could read the current verbage like this:

> In Chef 12 this will still work and cause a deprecation warning, and in Chef 13 the behavior will simply fail"?

Again, happy to move deprecation to 13 and break to 14, just wanted to make sure we're on the same page here regarding the nature of this RFC.  The intent is to break nothing at all in version N, but issue deprecation warnings for every behavior that will break in version N+1.

new/missing-method-missing.md
jkeiser: Does get_resource_class('cookbook_name')` work too?  Like, basically, "get me the class of the resource named X."  I like the idea of not treating them differently if we can get away with it.  Also, I think you have to put it on something with a run_context available to it.

new/missing-method-missing.md
jkeiser: Will do!  Also Phil had a suggestion I have to go look up as well.

new/missing-method-missing.md
coderanger: Yep, that [basically what I do](https://github.com/poise/halite/blob/master/lib/halite/spec_helper.rb#L122-L130).

new/missing-method-missing.md
coderanger: So we would continue to support non-name-based `provides?` via `method_missing` in Chef 12 with a warning? That sounds like a thing and would mean it could be done with full compat. I don't know of anything actually using that feature so should be easy to roll forward.

new/missing-method-missing.md
jkeiser: @coderanger having thought about it, I think I'm on board with making the "implicit `apt_package`" case a deprecation, because I really like the "`provides` = override the automatic stuff" idea and I can't imagine it being *super* widespread (or hard to change).  It will require a tiny bit of voodoo, but not that much.  I'll see what I can do.

I'd *really* like to avoid DescendantsTracker if I can, because the more we do that the harder certain customizations get.  But I'll

coderanger: Another slightly wonky option is to have a `Chef::ResourceV2` class or something you can inherit from to get the new behavior while inheriting from `Chef::Resource` gives you some compat hacks like auto-registering the class name. I don't love the name, but structurally it seems workable (prior art: Python's oldclass vs. newclass stuffs).

new/missing-method-missing.md
jkeiser: Believe me when I say I am keenly aware :) I am hoping we can get to the point where separate chef runs can go in separate threads, among other things. I am incredibly uncomfortable with our use of globals to store things that users do in a cookbook in a single chef run. But, baby steps ...

new/missing-method-missing.md
jkeiser: Yep, that's the case.  It would be more of an issue of risk if we decided not to do it in Chef 12; so far everything I've seen or thought of, including crazy things like calling `method_missing`, I've been able to get in as a deprecation (in some cases that was fun :)

jkeiser: To be even ridiculously clearer: `method_missing` doesn't actually go away in Chef 12.  But it is only used for deprecation warnings (everything else funnels through explicit methods) and we should be able to straight delete it in 13.

coderanger: I don't think that is correct, it will exist in Chef 12 and will still run the same behavior it does now in addition to the warnings.

jkeiser: I don't think we are actually saying incompatible things here. Everything you say is true. All I am saying is that after this change in chef 12 method missing should only ever be called in the event that you are using deprecated behavior :)

coderanger: "But it is only used for deprecation warnings" is the false part. It is still the same method (notably a bunch of my subresources stuff will still be calling it until I figure out a better way to handle resource name aliasing/search).

jkeiser:
@coderanger I modified the provides / does_not_provide sections, and provided an overview of what would be affected in Chef 12 (couple of extra features, and deprecation warnings).

new/missing-method-missing.md
jaymzh: it was foodcritic - but you have it in this RFC already :)

new/missing-method-missing.md
lamont-granquist: What is the use case for resource unloading?  We did do LWRP unloading and reloading in Chef 11 and killed that off for Chef 12 and simultaneously took away the ability to run interval runs in a single process which was the use case we had for unloading and reloading resources.

new/missing-method-missing.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Create explicit DSL methods for resources
+
+When looking up resources, there is no method anywhere that corresponds to the
+actual resource name.  This can lead to a lot of unnecessarily heavy thinking
+when following some very typical stack traces.  It also means that the name of
+the resource being declared is often not on the stack, making debugging harder.
+
+Here we propose that all resource and definition DSL be added to
+`Chef::DSL::Recipe` as actual methods, as they are created, so that they can be
```
lamont-granquist: Either you should use a cleanroom style class instead or Chef::DSL::Recipe should be converted to a cleanroom style class.  Probably the former.

new/missing-method-missing.md
lamont-granquist: I actually favor dropping the implicit magical class based stuff entirely and requiring at least one 'provides'.  For LWRPs and other DSLey ways of writing resources this can be done for the ultimate end user, but HWRP users should need to specify a provides line.  Obviously, this is a breaking change, Chef-13, yadda, yadda.

For back-compat we could wire up a define_method on inheriting from Chef::Resource if the resource is under the Chef::Resource namespace.  That method should then spit out a deprecation warning if its run.  Then a subsequent explicit provides line would define_method and overwrite the magically added one and would not spit out a deprecation warning.  Eventually the magic on inheriting from Chef::Resource gets removed.

new/missing-method-missing.md
```diff
@@ -0,0 +1,213 @@
...
+If you want to (for example) create a base class that does not have an actual
+DSL method, you can specify `does_not_provide` or `does_not_provide :my_name` in
+the class:
+
+```ruby
+class MyPackage::MyResourceBase < Chef::Resource
+  does_not_provide
+end
+# my_resource_base will not exist in recipe DSL
+```
```
lamont-granquist: And if you deprecate the implicit magic the need for this goes away.

new/missing-method-missing.md
lamont-granquist: Chef >= 13?

Usually when I write 'Chef 13' I mean 'at least Chef 13 if not later'.

new/missing-method-missing.md
```diff
@@ -0,0 +1,213 @@
...
+Removing `method_missing` means several methods of creating resource names cannot
+be caught.  These are the cases.  We will issue deprecation warnings for each one
+and remove them (and method_missing) come Chef 13.
+
+### Chef::Resource::MyResource = <resource>
+
+When you assign an anonymous class or a resource from another namespace into
+`Chef::Resource`, your new resource will currently be accessible via the DSL.
+Since there is no Ruby method that lets us catch the assignment of a constant,
+there is no way for us to assign an explicit DSL method to handle the resource.
```
lamont-granquist: And if you don't do magic wiring on inheriting and force using provides lines this all goes away as well?

new/missing-method-missing.md
```diff
@@ -0,0 +1,213 @@
...
+
+The implications of this are that we have to scan every single resource when
+you type `blah` in a recipe, just so we can catch this one.  In Chef 12 we will
+issue a warning and tell you to do this instead:
+
+```ruby
+class MyResource < Chef::Resource
+  provides :blah
+end
+```
```
lamont-granquist: So this scanning is required for providers, and the symmetry between resources and providers is **very** deliberate.  Unless there's a compelling argument that this is a high resource consumption problem, I'm :-1: on this optimization.

new/missing-method-missing.md
lamont-granquist: `Chef.resource_class(:yum_package)`?

new/missing-method-missing.md
coderanger: I use it for running tests, anonymous resources are created before the example runs and removed afterwards. Otherwise I would end up declaring resources like `poise_test48` all over my tests to avoid name collisions and that gets annoying.

new/missing-method-missing.md
coderanger: If we want to allow for scanning-driven behavior then we cannot get rid of method_missing because it needs to check all the `provides?` for any method call. I'm totally fine with that but it basically makes this RFC moot.

coderanger: Well I guess we could still make the specific methods for the benefit of nice tracebacks, so not completely moot, but it would defang a lot of the switchover.

new/missing-method-missing.md
coderanger: I think that is basically the same as what we were discussing, but :+1: on that overall.

coderanger:
So I think we have hit a fork in the road. If we want to remove `method_missing` that is doable but will require the eventual removal of the `provides?` scanning feature for resources. If we are not okay with that feature loss, then we can drastically simplify this proposal to basically just make `provides` define a method in the DSL module and nothing more since it is an opt-in feature to get slightly nicer tracebacks.

new/missing-method-missing.md
jonlives: Yeah I think we need to be extremely clear here exactly what will happen instead, as this is one of the more common use cases I'd have thought.

new/missing-method-missing.md
jonlives: I don't like the idea of breaking symmetry between resources and providers, but I like the cleanness of using ```provides :blah``` here...in the resource context, I'm not that familiar with the use cases for defining a custom ```provides?``` method versus using ```provides :blah```, can anyone fill in the blanks there a little?

jonlives: @lamont-granquist (or anybody else) could you clarify what provider behaviour is driven by that scanning please? I may not have had enough coffee, but I feel like I'm missing some details to fully understand this one.

jonlives:
I'm broadly in favour of this RFC, and my initial thinking was that removing ```method_missing``` would provide a neater eventual solution, but after reading @lamont-granquist's points I need to grok a little better how this breaks the overall resource / provider symmetry.

new/missing-method-missing.md
jkeiser: In any case, I added `provides_nothing` for this, so that if your class *does* extend `Chef::Resource`, you have a way to turn it off.

new/missing-method-missing.md
jkeiser: There will be some sort of method.  Right now, you should be able to do `Chef::DSL::Resources.resource_classes['my_resource'].  A getter method makes sense too.

new/missing-method-missing.md
jkeiser: The actual solution I implemented was a module named `Chef::DSL::Resources`, which I include into `Chef::DSL::Recipe`.  But I'm unclear what the purpose of a cleanroom style class is here, and whether that solution addresses what you are talking about?  Can you clarify?

new/missing-method-missing.md
jkeiser: Deprecating the implicit magic would cause a large amount of backcompat woes.  I'd like that to be a separate RFC.

My feeling is, Chef::Resource should always do implicit magic for backcompat, and when we move to some sort of real base mixin that is our true public API for resource, we can just not implement implicitness there.

new/missing-method-missing.md
jkeiser: To be clear, this section of the RFC could *not* go away.  It still has to work in Chef 12, and be marked deprecated.  It would just be a different reason (classes don't get implicit values).

*Also* to be clear, the logic implementing this would go away in Chef 13 in *both* cases.  There is no net gain for removing the implicit gain, as far as this goes.

new/missing-method-missing.md
jkeiser: `provides` handles all of this, to be clear.  provides :x with a *block* performs the same function, without requiring a full scan, and gives us all the information we need to create an explicit DSL.  I don't think we need to abandon explicit methods just because we currently scan.  We just make `provides?` @api private and require you to use `provides` to specify your DSL.

jkeiser:
It's not necessary to remove `provides?` or break resource / provider symmetry around that; it may be necessary to modify the rules around it, however.

jkeiser:
Specifically, the rule could be:

### `provides` required in order for `provides?` to be called

In order to be considered a potential provider or resource for the 'my_resource' DSL, a user *must* call `provides 'dsl_name'`.  This means that anyone overriding `provides?` to provide a finite set of DSL can still do so.

It means, however, that someone who overrides `provides?` to provide an *infinite* set of DSL cannot.  For example, if you say "I provide ALL DSL on the Solaris platform, you can't.  If you say "I provide sql_<database_name>_<table_name>", you cannot.

This has the advantage that, after scanning all resource classes, you know with certainty the *full* set of DSL, the full interface for a cookbook.

jkeiser:
Actually, the RFC is even more relaxed than that, without breaking anything: `provides?` scanning still happens on *both* resources and providers.  However, we don't try to do provides? scanning unless someone has declared that the DSL method exists with `provides`.  While I think we should eventually restrict it   I don't think we should change the RFC on that point in particular; going any further is unnecessary to achieve the aims.

If we are OK with the rule "you must declare the name of a DSL in a `provides` somewhere," but you may use custom rules in any resource and provider class to implement it," then we're OK for the moment.

jkeiser:
The final thing we could do--which would *still* be a big net gain in my opinion--is not deprecate method_missing at all, but still implement explicit methods everywhere.  All of the benefits we get here, we would continue to get.  The system might be hard to understand in cases where you do a hidden `provides?` with a name that was never declared as DSL anywhere, but eh, you kinda asked for that complexity when you did something like that.

I'll admit I'm not a fan of this approach, and would rather not change to it, but I'll definitely take it if we feel that being able to say essentially `provides *` and `provides /sql_.*_.*/` in an implicit method is important.

jkeiser:
OK, here's the full description of the issue we're discussing, for those following.

Right now, the Provides system has a `provides?` method on every single resource and provider.  When you call a DSL method, it goes through *all* Resources and Providers (descendants of Chef::Resource and Chef::Provider) anywhere, and calls provides? on each one.  So you can declare a `super_file` resource like this:

```ruby
class MySuperFileResource < Chef::Resource
  def provides?(node, resource_name)
    resource_name == 'super_file'
  end
end
class MySuperFileProvider < Chef::Resource
  def provides?(node, resource_name)
    resource_name == 'super_file'
  end
end
```

In Chef 13, with this proposal, typing `super_file` in a recipe would no longer work.  You would need to do this instead:

```ruby
class MySuperFileResource < Chef::Resource
  provides 'super_file'
end
class MySuperFileProvider < Chef::Resource
  provides 'super_file'
end
```

This isn't too onerous--`provides` even takes a block that passes in the node, so you can do any arbitrary logic you want to do to decide whether you handle the resource.  Where it gets tricky is this:

```ruby
class MySuperFileProvider < Chef::Resource
  def provides?
    true # provides any and all resources, because it is THAT awesome
  end
end
```

Or other arbitrary criteria that allow for arbitrary names.  provides 'name' doesn't support that.  We would be causing those cases to be impossible.  I'm personally OK with it.  I'm uncomfortable with the way Rails does this with ActiveRecord, and at any rate, I don't think we *intended* it, it just happened as a side effect and no one is actually using it this way.  If we add the feature in later, we should do so intentionally.

cc @jonlives

jaymzh:
I'm with you @jkeiser - providing a regex of DSL entries seems crazytown to me.

new/missing-method-missing.md
jkeiser: So, @coderanger and I talked elsewhere, and I *think* we ended up at "as long as it's actually *true* that it doesn't break backcompat, it's OK to include the deprecation warnings in Chef 12 and then remove in Chef 13."

new/missing-method-missing.md
jkeiser: So, what I implemented was:

- If you do not specify `provides` in your class at all, and inherit from Chef::Resource, it gives you a magical DSL name.
- If you *do* specify `provides` in your class, you don't get any magical DSL name.

I think @lamont-granquist is suggesting that we change it to this:

- If your class descends from `Chef::Resource` *and* is directly under the `Chef::Resource` namespace, *and* you do not specify `provides` in your class, then we give you a magical name and warn you about deprecation.
- If your class doesn't meet all three of those criteria, you don't get a magical name or warning.  If you are outside of `Chef::Resource` and don't specify `provides`, in fact, you don't get a magical name at all.

I'm game to that if @coderanger and @lamont-granquist are both on board.

Last question on that: should we warn the user in that last case?  Like, if the user doesn't specify `provides`, they might well be surprised not to be able to use their resource in the DSL.

coderanger: That seems fair, matches current behavior and doesn't add the new case of "inherits from `Chef::Resource` but does not call `provides` and is not in the `Chef::Resource` namespace". As it stands that would simply do nothing so I vote we just keep that behavior.

jkeiser: Yeah, as long as our eventual aim is to deprecate auto-provides, that is the best way to go.  I'll work that in on the assumption that @lamont-granquist agrees with himself :)

jkeiser: Heh, I see the `provides_nothing` use case.  I'll keep that in.

jkeiser:
@coderanger @lamont-granquist one more question: once we deprecate the implicit magic entirely and only support `provides`, `Resource.dsl_name` and `LWRPBase.resource_name` are in a ... weird place.  It feels like we should deprecate them entirely, and rely on `build_resource` to tell the class what its name is.  What say you?

jkeiser:
@coderanger @lamont-granquist just wanted to check whether we're in good shape on this or if there's anything else we need to hash out before tomorrow.  I've got a working patch I'd like to get in soonish, so that it's cleared out and we can move on to next things.

As to the last question, I modified the RFC to deprecate implicit `provides`.  I did not modify `dsl_name` and `resource_name`, but I'll take a shot at all that in a future RFC (when we start getting super explicit about our public API / possibly a mixin).

new/missing-method-missing.md
```diff
@@ -0,0 +1,174 @@
...
+class MyResource < Chef::Resource
+  def provides?(node, name)
+    name == :blah
+  end
+end
+```
+
+The implications of this are that we have to scan every single resource when
+you type `blah` in a recipe, just so we can catch this one.  In Chef 12 we will
+issue a deprecation warning and tell you to do this instead:
```
coderanger: Might want to be clearer that this will be removed in Chef 13.

coderanger:
@jkeiser I think we should probably just get rid of provides_nothing at this point, it is extra stuff to maintain and that isn't a use case that is currently supported (defining a class under `Chef::Resource` that should not be part of the DSL). If you want to make an abstract base class just put it in your own namespace and it works as expected (no call to provides means no DSL)?

jkeiser:
Yeah.  I'm fine with removing `provides_nothing` now that we're not doing anything automatic.  I'll do that.  And yes, if you make an abstract base class outside of `Chef::Resource`, and don't specify `provides`, you get no DSL.

lamont-granquist:
Yeah, provides? can be less lazy and can be moved somewhere right after we first build the node.  The right solution to this probably does away with descendent tracking completely and requires every provider that uses dynamic resolution to have a provides? line.   We should also be able to refactor into using a single node map at the same time -- although we might need to be careful with handling the case of having multiple providers that can be returned by dynamic resolution with the single node_map (I'll really need to sit down with the code to figure out that, not sure it works in my head).

I'd like to see all of that internal cleanup get done.  @jkeiser if your branch doesn't cover all of that, you should probably complete what you can get done so that I can patch on top of that.

jkeiser:
Yeah, I spiked on getting everything into a single node_map, but it wasn't necessary to get this done so I didn't finish.  We can definitely chat about that.

lamont-granquist:
Yeah, in the shower today I became not sure that works the way that I want.  Generally would be good to push your code and start to more collaboratively hack on stuff.

jkeiser:
@lamont-granquist @coderanger note: I wrote up a new implication of forcing people to use `provides` that wasn't clear before: `Chef::Resource::FooBar` has a *lower priority* than `Chef::Resource::FooBar` with `provides :foo_bar`.  Adding explicit provides bumps it up, putting it on even footing with any other class that declares `provides` (and we resolve conflicts the way I pick restaurants: alphabetically).  the user is supposed to make anyway: set a priority array.

There are ways we could simplify the resolvers by modifying the rules, but I think that merits a whole separate topic and RFC.

jonlives:
@chef/rfc-editors this one's decided and ready to rock too!

nathenharvey:
ping @chef/rfc-editors can we get this merged before tomorrow's meeting, please?

lamont-granquist:
yeah code is merged and this is gonna ship soonish.

PR #119, opened 2015-04-23
update maintainers file url to match current version

trevorghess:
The Markdown file that was previously targeted has been changed to a TOML file, and the url was broken in the RFC.

### DIFF
--- a/rfc030-maintenance-policy.md
+++ b/rfc030-maintenance-policy.md
@@ -169,4 +169,4 @@ If you would like to speak to, or understand the position of, someone at Chef So
 
 # The MAINTAINERS file in Chef
 
-The current [MAINTAINERS](https://github.com/chef/chef/blob/master/MAINTAINERS.md) file resides in the [Chef](https://github.com/chef/chef) repository on GitHub.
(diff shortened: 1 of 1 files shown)


rfc030-maintenance-policy.md
```diff
@@ -169,4 +169,4 @@ If you would like to speak to, or understand the position of, someone at Chef So
 
 # The MAINTAINERS file in Chef
 
-The current [MAINTAINERS](https://github.com/chef/chef/blob/master/MAINTAINERS.md) file resides in the [Chef](https://github.com/chef/chef) repository on GitHub.
+The current [MAINTAINERS](https://github.com/chef/chef/blob/master/MAINTAINERS) file resides in the [Chef](https://github.com/chef/chef) repository on GitHub.
```
nathenharvey: It's now actually a .toml file available at https://github.com/chef/chef/blob/master/MAINTAINERS.toml

trevorghess:
@nathenharvey, I added the .toml extension :)

thommay:
Once chef/chef#3286 is merged we won't need to change the RFC.

thommay:
(Thanks for pointing out that it was incorrect though!)

PR #120, opened 2015-04-24
Telemetry subsystem : Capture and expose chef run performance metrics

ranjib:


new/rfcXXX-telemetry.md
```diff
@@ -0,0 +1,88 @@
...
+Chef telemetry
+
+## Motivation
+
+  As a chef user and developer
+  I want an easy way to get chef run performance data
+  so that I can identify areas that can be optimized
+
+As we automate more things with chef and control more system resources,
+chef run time gets longer, chef client process consume more memory and cpu.
```
thommay: "chef run time increases, and the chef client process consumes more memory and cpu."

new/rfcXXX-telemetry.md
```diff
@@ -0,0 +1,88 @@
...
+As a configuration management system, it is important that chef run
```
thommay: "a chef run"

new/rfcXXX-telemetry.md
```diff
@@ -0,0 +1,88 @@
...
+does not impact the underlying host performance. Over the last few years
+Chef community has shared a number of optimization techniques. Aim of this
+RFC is to facilitate some of those common patterns by providing a robust
+telemetry system in core chef. If configured, the telemetry system will give users fine
+grained metrics about chef run that can be used to find out
+performance hiccups. This will also help users quatify how a chef
```
thommay: "quantify"

new/rfcXXX-telemetry.md
```diff
@@ -0,0 +1,88 @@
...
+run impacts the uderlying system.
```
thommay: "underlying"

new/rfcXXX-telemetry.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Specification
+
+Telemetry module will provide following metrics:
+
+- Time spent on key chef run milestones: At which stage chef run spends how long? Telemetry system
```
thommay: "At which stage ... ?" -> "How long does the chef run take for each stage?"

new/rfcXXX-telemetry.md
```diff
@@ -0,0 +1,88 @@
...
+  will declare  a subset of the current events as major milestones, these are: run_start,
+  ohai_completed, node_load_completed, cookbook_resolution_complete, converge_start,
+  converge_complete, run_completed/failed. Metrics will be captured during these milestones via event handlers.
+  This can be used to estimate time spent on cookbook sync, compilation
+  convergence phase etc.
+
+- Time spent on individual resource & recipe convergence. This information is already available
+  using the API, telemtry system will just consolidate them, along side the other metrics.
```
thommay: "telemetry"

new/rfcXXX-telemetry.md
```diff
@@ -0,0 +1,88 @@
...
+  telemetry.add_metric 'vsize' do
+    Sys::ProcTable.ps(Process.pid).vsize
+  end
+end
+```
+
+In addition to capturing these metrics, telemetry system will also allow publishing
+them via statsd or saving them as node attribute.
+
+Telemtry sysetm will be configured via the standard chef config file and it will be
```
thommay: "Telemetry"

thommay:
Besides the grammar nits, I'm :+1:

jonlives:
:+1: It'd also be good to specify a command line option to enable telemetry for the current run rather than only globally enabling / disabling in the config file

ranjib:
@thommay fixed the grammatical mistakes, typos etc. @jonlives add CLI option

new/rfcXXX-telemetry.md
```diff
@@ -0,0 +1,94 @@
...
+
+Chef.telemetry do |telemetry|
+  telemetry.add_metric 'vsize' do
+    Sys::ProcTable.ps(Process.pid).vsize
+  end
+end
+```
+
+In addition to capturing these metrics, telemetry system will also allow publishing
+them via statsd or saving them as node attribute.
```
seth: I could see how it might be a convenient stepping stone to store metrics as node attributes. I'm concerned that adding to the node data this way -- especially if not given special handling on the server -- will have some downsides (node size increases, more data to transmit/parse/index) and doesn't provide access to the metrics in a way that will be particularly delightful.

ranjib: @seth agree. hence this is optional and the target attribute namespace is also configurable. this will be useful for scenarios where users dont have an existing metric system (like statsd/graphite). We should definitely document the effect of storing metrics as node attribute. But in its current form, this allows users fetching/displaying these metrics in same manner as `knife node show foo -a 'metrics'`.

jonlives: Might we run into an issue here where we end up dealing with requests to add the ability to publish using other formats? Although commonly used, statsd isn't ubiquitous and if implementation of this is too tied to that format it might  hinder future expandability.

ranjib: @jonlives i want to implement the publishing system as a submodule in the telemetry system. and it will come with statsd based publishing out of the box, because i felt its most popular. But we'll definitely be able to add custom publishing mechanism. I was hoping that having a out of the box publishing system will ease writing additional publishing mechanism. Do you think we should explicitly mention that, along with a api spec? I was feeling it will make it bit complicated, as the whole telemetry system is new. 

On the pluggable publishing mechanism, how you feel about this:
```ruby
config_context(:telemetry) do
  resource true # captures per resource execution time taken
  recipe true # captures per recipe execution time taken
  gc true # captures GC stats during main chef events
  process true  # captures process memroy stats from /proc during main chef events
  client_run true # captures time spent on major chef run milstones
  publish_using Chef::Telemetry::Statsd.new( host: '192.168.2.11', port: 7676) # emit data to statsd
end
```

miketheman:
This is interesting as a concept - getting more insight into Chef runs is a great idea.

I wonder how this compares with execution handlers? Maybe instead of adding (yet another) command line flag, we evolve the handlers to collect these details and report them via report handlers?

new/rfcXXX-telemetry.md
```diff
@@ -0,0 +1,94 @@
...
+
+Telemetry sysetm can be configured via the standard chef config file or using a dedicated
```
smith: "system"

new/rfcXXX-telemetry.md
jamesc: I'd love to have described what an endpoint in Chef Server would be for these metrics.  Maybe this is exposing a Statsd port, or a full RESTful endpoint where you can upload and query the data (like reporting does now).  That would feel more like a more integrated end-to-end solution

new/rfcXXX-telemetry.md
```diff
@@ -0,0 +1,94 @@
...
+  ohai_completed, node_load_completed, cookbook_resolution_complete, converge_start,
+  converge_complete, run_completed/failed. Metrics will be captured during these milestones via event handlers.
+  This can be used to estimate time spent on cookbook sync, compilation
+  convergence phase etc.
+
+- Time spent on individual resource & recipe convergence. This information is already available
+  using the API, telemetry system will just consolidate them, along side the other metrics.
+
+- GC related data. Ruby's internal memory usage obtained from
+  GC.stat, captured during the chef run milestones.
```
evan2645: Lots of time metrics by default, but very few memory metrics by default... only GC statistics. Broader memory statistics for chef client processes should also be part of the core implementation alongside GC stat collection, for each milestone... any reason those were left out?

new/rfcXXX-telemetry.md
jonlives: I think mentioning explicitly that we'll support non-statsd systems with a rough implementation idea, doesn't need to be too exact.

adamedx: @jamesc why wouldn't we just use the reporting endpoint and store?

ranjib: @jamesc i have never used the reporting features., i think it will be easier if we get a rough implementation out for review we can decide on a reporting end point .. likely to be a a json blob with top level keys predefined (a rough implementation will help freezing on those keys/metric names)

ranjib:
@miketheman report handlers will not provide some of the data that event handlers can give, hence the second one (like time spent on run phases). Also primary usecase for this feature is diagnostic or performance optimizations, and i expect this more to be on demand, hence the cli flag, it can be configured to be used as report handlers via config (i.e. persistent across chef runs) , to capture trends. but i think that use case is less. This will also help report handlers to consume a standard set of metrics (like dd or others) provided by core chef instead of each of them defining their own.

miketheman:
@ranjib I'll point out that I was looking at execution handlers, not report handlers - as those are the interfaces that have the details.
A report handler can continue to report whatever it wants - having the execution handler provide any details to the report handler is how the flow works today, is it not?

In my mind, the use case would be to collect all metrics and events every Chef run, and not have a CLI flag to emit them only on demand, as if you're not tracking the executions over time, you have less of a comparison to measure against.

If an emitter is to a local UDP socket, a la statsd, does it matter if it's running all the time? Chef Server does this already.

new/rfcXXX-telemetry.md
```diff
@@ -0,0 +1,94 @@
...
+  resource true # captures per resource execution time taken
+  recipe true # captures per recipe execution time taken
+  gc true # captures GC stats during main chef events
+  process true  # captures process memroy stats from /proc during main chef events
+  client_run true # captures time spent on major chef run milstones
+  statsd( host: '192.168.2.11', port: 7676) # emit data to statsd
+  attribute 'chef-metrics' # save all metrics under node['chef-metrics'] attribute
+end
+```
+Eaxmple of enabling telemetry using the CLI flag
```
miketheman: typo: Example

ranjib:
@jonlives @evan2645 i have added the metrics publishing API and specified that statsd and node attribute based metrics publishing will be implemented using that. Let me know what you guys think

jonlives:
That looks great @ranjib!

jonlives:
Approved, by the powers vested etc etc. Take it away @chef/rfc-editors

thommay:
Accepted, thanks Ranjib!

PR #121, opened 2015-04-29
Define actions directly on resources

jkeiser:


new/resource-actions.md
```diff
@@ -0,0 +1,199 @@
...
+
+The action recipes are all added to a `Provider` class which inherits from the
+parent resource's provider class.  You may call `super` to call the parent
+class's version of the current action.
+
+```ruby
+class MyFile < Chef::Resource::File
+  action :create do
+    puts "I am a file yo"
+    super
```
coderanger: Needs to be super() because Ruby is terrible.

new/resource-actions.md
```diff
@@ -0,0 +1,199 @@
...
+  end
+end
+```
+
+### Inline resource output
```
coderanger: Would love to see output improved, but I think this part is out of scope for this RFC.

new/resource-actions.md
```diff
@@ -0,0 +1,199 @@
...
+  end
+end
+```
+
+`converge_by` (the current way of converging / updating) a resource will still
+be available for those who are used to it.
+
+### `Resource.why_run_description`
+
+When why run is enabled, we automatically prepend "Would" to the text of the
```
coderanger: As above, not directly related to the title.

coderanger:
:+1: :+1: :+1:

Implementation will be tricky, but very much in support of this as a simpler option for new folks. [Poise::Helpers::Fused](https://github.com/poise/poise/blob/cfef644b00d9239e9e6a7ce6dcf32625fa224bfb/lib/poise/helpers/fused.rb) was my stab at this concept a few months ago, but I think @jkeiser has worked out a better solution for how to handle the action evaluation context.

Also discussed on IRC but would really like to see this implemented in a cookbook first and then added to core in 12.4 or 12.5, so cookbook authors can start using it immediately without worrying about upgrade rates (see also: chef_version in cookbook metadata :)

new/resource-actions.md
```diff
@@ -0,0 +1,199 @@
...
+run.
+
+- The entire action recipe is compiled (just like a normal recipe) before
+  converging.
+- This compile happens when the action is *run*, not when the parent resource
+  is declared.
+- Delayed and immediate notifications are local to the action.
+
+This is equivalent to `use_inline_providers` in a Provider, and in fact is
+implemented that way.  We just call out the behavior here to make it clear what will happen.
```
lamont-granquist: This seems like a whole lot of words to just describe that we're constructing a normal LWRP provider with use_inline_resources as the default behavior.

lamont-granquist: We should also probably think about another flag like use_inline_resources to use to have resources not execute two phase with compile/converge mode and consider turning that on by default, with the option to turn it off.

jkeiser: I'm actually not sure that's what will happen.  This thing needs to feel like a resource, not a provider, and most of the provider DSL probably shouldn't be in scope.

new/resource-actions.md
```diff
@@ -0,0 +1,199 @@
...
+class's version of the current action.
+
+```ruby
+class MyFile < Chef::Resource::File
+  action :create do
+    puts "I am a file yo"
+    super
+  end
+end
+```
```
lamont-granquist: How does this work in the case of the Service resource where there are multiple possible providers that it can use?  In that case, the superclass to pick as the parent of the constructed provider seems ambiguous.

And I do tend to think we should move to a 1-1 mapping of resources to providers, but we don't have that right now, and to resource resolution with the same delayed behavior that the provider resolver does now we'd need to eliminate the two-phase compile/converge mode, so i think we're stuck with ambiguity for now.

new/resource-actions.md
```diff
@@ -0,0 +1,199 @@
...
+* x[blah] action blah
+  - create new file /Users/jkeiser/x.txt
+  - update content in file /Users/jkeiser/x.txt from none to a591a6
+  --- /Users/jkeiser/x.txt	2015-04-29 08:15:14.000000000 -0700
+  +++ /Users/jkeiser/.x.txt20150429-56731-6xk0lx	2015-04-29 08:15:14.000000000 -0700
+  @@ -1 +1,2 @@
+  +Hello World
+```
+
+The principle is: the user doesn't see resources they don't type.
```
lamont-granquist: I don't like this particular bikeshed at all.  I read what the LWRPs are doing all the time and it helps me make sense of what is going on inside of them, and I don't find it confusing at all.

new/resource-actions.md
jkeiser: Also, I just realized, the last sentence there directly contradicts what I just said, which is what you were actually trying to tell me. Fixing.  :)

new/resource-actions.md
```diff
@@ -0,0 +1,199 @@
...
+
+  directory '/x' do
+  end
+  file '/x/y.txt' do
+  end
+end
+```
+
+`converge_by` (the current way of converging / updating) a resource will still
+be available for those who are used to it.
```
lamont-granquist: I'm not certain I find the use case totally compelling.  And it smells like something which you would probably want to do within the current resource collection and not open up a new resource collection context.   If you're not doing notifies/subscribes to those resources then you could just use an if block to make the resources go away.

It also feels like a use case that would be better handled by wrapping resources with a block of defaults like you wanted to do with user/group/mode to prevent repetitive typing.

new/resource-actions.md
```diff
@@ -0,0 +1,199 @@
...
+
+We add the following public API to `Resource`:
+
+- `self.action(action, class: nil, &block)`: creates a Provider with the given
+  action
+
+And modify the following:
+
+- `allowed_actions`: defaults to all actions in the resource's provider class(es).
+- `default_action`: defaults to `allowed_actions.first`.
```
lamont-granquist: not sure i like the magical default default_action.

new/resource-actions.md
jkeiser: This doesn't work (literally) unless the parent class is defined with a single provider class.

It is *possible* to make it work by declaring the subclass on the fly when we are about to run the action (it would pretty much be the same thing as if we ).  However, that also means we'd need to know the DSL name of the parent, since that's the key we use to determine the provider.  We could call `superclass.dsl_name` to get this.

It is not unreasonable.  Let's mull it over.

new/resource-actions.md
jkeiser: I will remove this and move it to its own RFC, and we can discuss it there.

new/resource-actions.md
jkeiser: The intent is that if you only type `action` once, that gets selected.  I'd also be OK with defining an "official" action named `:converge` or `:create`, and if you define that action block, it gets set as the default.  What I'd rather avoid is forcing the user to type it twice.

new/resource-actions.md
jkeiser: I'll remove for now.  It's not essential, just a series of things I thought would make it easier to do recipes as they become more common.

jkeiser: (And `resource_group` is basically the same, which I'll probably resubmit when I get further towards parallel stuff.)

new/resource-actions.md
jkeiser: Done.  I'll open up a new one about the changes required to make resources a viable alternative to `converge_by`.

new/resource-actions.md
jkeiser: Oh, regarding immediate execution vs. compile/converge mode: totally with you.  Will make a good RFC.

new/resource-actions.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Provider
+
+Resources in LWRP actions will no longer report in output as nested actions
```
jaymzh: wait, so if I make an LWRP that has 5 recources inside of it, those resources never get printed out? That'll make debugging my LWRP or understanding what went wrong VERY hard. :-1:

coderanger: https://github.com/chef/chef-rfc/commit/dd023e525c6e801d9053b22e3d407692eb467adc removed this section from the description, should probably just remove this too in favor of a future RFC discussing changes to the output system.

jkeiser: Yeah, I didn't realize this section was still here.  It's not meant to be.

jkeiser: Fixed.

martinisoft:
I am :+1: on this.

new/resource-actions.md
lamont-granquist: I like the magical official action a lot less.

new/resource-actions.md
```diff
@@ -0,0 +1,142 @@
...
+class's version of the current action.
+
+```ruby
+class MyFile < Chef::Resource::File
+  action :create do
+    puts "I am a file yo"
+    super()
+  end
+end
+```
```
lamont-granquist: Can you include words that this works only for resources which have a single defined provider class?   And we should make sure that for resources that have a 1-1 mapping that this information is available without any magic.  Although I'm not certain how this fits within the dynamic provider resolver.

But some kind of words towards the point that it works when the superclass has a 1-1 relationship with the provider, and it doesn't work for the service provider.

new/resource-actions.md
coderanger: Some prior art on this: https://github.com/poise/poise/blob/master/lib/poise/helpers/lwrp_polyfill.rb#L40-L52

Been working really nicely in Poise.

lamont-granquist: Yeah, if we're supporting inheritance, then actions + default_actions should inherit properly.

thommay:
I very much like this :+1:

new/resource-actions.md
jkeiser: @lamont-granquist actually, we *could* make it work if there are multiple provider classes.  However, I'm not sure it's prudent to guarantee anything other than that it will work if the parent class defined the same action using `action :create`.  I'll update it to mention that it is only guaranteed to work on that.  We can decide whether it oughta work on more things in the PR itself, but I think that's a reasonable place to start.

new/resource-actions.md
jkeiser: They do inherit properly, in this scheme at least.  Specifically, if parent has actions A, B and C, and the child adds D and E, allowed actions will be A, B and C, and default action will still be A (because it is first).

jonlives:
Sold, to the man with the flowery shirt. Also, approved for merge. Have at it @chef/rfc-editors

thommay:
Merged as RFC 50. Thanks, John!

PR #122, opened 2015-04-30
Remove outdated iCal

juliandunn:
This ics file is outdated as it specifies biweekly meetings; I'm not sure if it was ever useful for people?


nathenharvey:
:+1:

PR #125, opened 2015-05-13
Create RFC template for easy copying

jkeiser:


ranjib:
:+1:

coderanger:
We should just make sure to update this in sync with RFC 0 as needed. There has only been two changes to that template so far though, so that seems entirely manageable. :+1:

coderanger:
(as this is a purely procedural thing and not new content, we can probably merge with 2 +1s from the editors team /cc @chef/rfc-editors)

ranjib:
i think we should add reference of this template on rfc#000

jonlives:
:+1: - as that's 2 +1s, gonna merge this.

PR #126, opened 2015-05-14
DRAFT: Add the ability to run resources immediately after declaration

jkeiser:


danielsdeleo:
Grab bag of thoughts:

### Why I like the general principle

I've long seen a lot of user pain around the current 2-pass system we have now. It's not just inline non-declarative code, but stuff like:

```ruby
package "my-internal-application"

template "/etc/my_app.conf" do
  variables "var" => THIS_VALUE_DEPENDS_ON_SOMETHING_IN_THE_PACKAGE
end
```

To make this work, you either need to do some magic in a LWRP, or add a ruby block resource to run the code that fetches the value and assigns it to an attribute or the run state, then use lazy resource attributes to assign the value to the resource (because just `node["my_app_setting"]` would be evaluated at compile time to `nil`).

Anecdotally, a lot of users in this boat seem to be new to CM, porting an existing thing to Chef. I don't like confronting these kinds of users with lots of the edges cases of our system at once.

### Visibility

I don't like that this is in metadata. If my code is gonna behave way differently, I want it to say `totally_crazy_different_behavior_enabled(true)` in that file.

### Ordering

I agree with @adamhjk that having single recipes opt-in to immediate mode while everything else works as before is probably not a good solution.

## Challenges to Implementation/Adoption

### Accumulators

Immediate execution makes any sort of accumulator pattern impossible. For example, let's say we have to enable SW firewalls on our hosts, and we want to annotate each recipe with a list of the ports we need to open, and we then want a resource to configure the firewall at the beginning of the converge phase.

We can fix this if we move recipes into code blocks, though. For example:

```ruby
# Syntax imagined up on the fly
Chef.accumulators(:firewall_ports).add(22)

Chef.recipe do
  package "openssh-server"
end
```

### Notifications

Forward notifications won't work (as is noted in the existing proposal). I think that most notifications could instead be modeled as a specialized only if, e.g.:

```ruby
template "server.conf"

service "server" do
  conditional_action(:restart, depends_on: "template[server.conf]")
end
```

### Rewind

chef-rewind stuff won't work. But we could instead have a "pre-wind", e.g.:

```ruby
include_modified_recipe "foo::bar" do
  template "server.conf" do
    cookbook "my_wrapper"
  end
end
```

## Half-way: Per-Recipe Converge

Going half-way to full immediate mode, we could instead have per-recipe compile and converge. This has implications for cross-recipe forward-looking notifications, and we might not be able to fix it cleanly with the `conditional_action` idea, because you might notify a resource from a 3rd party cookbook from a first party cookbook. Not sure what the optimal solution is there. We might just have to not support cross-recipe forward looking notifications.

## Implementation

So, we could certainly put all the above together and it might be cool, or we might encounter unknown unknowns that doom us to failure. So, even though I think my ideas are good, I think we need some validation that it's worth doing all this before we embark on that road (unless, of course, we want to adopt some of the above ideas because of their own merit). IOWs I want to know what immediate mode feels like, how much stuff it breaks, etc. before committing to a lot of work which IMO will be necessary to really get it right.

jkeiser:
Consensus today at the IRC meeting was two things:
1. It's not a good idea to mix immediate and non-immediate execution in the same run.  Whatever we do, should work across all resources in a given context, or none.
2. We should run an experiment to see how this feels.

And other things that came up:
1. Several people feel notifications to and from resources in immediate mode cannot be made to work as they stand.

Next steps are to run some experiments with the client that:
1. Immediately execute every resource in either the Chef run (cookbooks+recipes) or inline providers, or both.
2. Converge each *recipe* when it finishes compiling.

These could be feature flags in the core client, or they could be separate monkeypatch gems (I think the former will be far easier to manage).

I'll write some code and maybe modify the proposal (though I probably can't do it this week), but I'll leave this proposal open in the meantime for discussion.

lamont-granquist:
Ideally I'd just like to see a flag like use_inline_resources that could be hung off an LWRP or a new-style merged RP and would force immediate execution for the whole sub-run_context/resource_collection.

I'd also like to see new recipe blocks become their own sub-resource_collection that was put on the main resource collection similarly to an LWRP.  Each recipe block would then compile/converge before later recipes are run -- however all recipes would be declared in the main compile mode before any were converged.  This also has the advantage of leaving the main compile/converge mode alone, then cookbooks that come late in the run list could declare things in the main compile phase which would be run to completion before anything was converged.  That way 'late' recipe code in the run_list can still toss forward state to 'early' resources that converge.  But gradually if all the code is moved into LWRPs or recipe blocks in immediate mode then we dramatically reduce the damage that compile/converge mode does while still keeping just enough of the flexibility that we need.

jkeiser:
What do you mean by new recipe blocks?  Are you talking about `action :create do` in resources, or something else?

danielsdeleo:
I think that's referring to my idea that you write recipes like:

```ruby
Chef.recipe do
  # recipe stuff
end
```

I really need to RFC that up. I think it's cool because then you could have syntax changes without breaking:

```ruby
Chef.recipe dsl_version: 5 do
  # crazy new syntax, wow
end
```

Anyway that's a bit off topic for here

jkeiser:
Oh!  I am completely down with it.  Can I add that `recipe` should be included in recipe DSL, so that you can just write `recipe do ... end` in a recipe?

I actually think `recipe` might be the righter word for `resource_group`.

... and yeah.  Off topic for here.

lamont-granquist:
Yep, I was talking about that thing you both just talked about. =)

coderanger:
In addition to the above feedback, the use of the term `:delayed` is confusing as that already means "at the end of the run" for notifications.

jkeiser:
@coderanger re `:delayed`, it's actually a deliberate choice: in `:delayed` mode, things behave as if we place a `:delayed` notification in the queue for your default actions, after each resource is compiled.  In `:immediately` mode, things behave as if we did an immediate notification for the default actions.

jkeiser:
I'm calling this closed but have no issue if someone wants to reopen it. Rather not keep the repo polluted with drafts.

lamont-granquist:
It got implemented more or less, here:  https://github.com/chef/chef/pull/8668

jkeiser:
Hooray!

PR #129, opened 2015-05-15
Ubuntu 10.04 LTS is no longer a supported platform.

mattray:
Ubuntu 10.04 was dropped by Canonical as a supported platform at the end of last month.

### DIFF
--- a/rfc021-platform-support-policy.md
+++ b/rfc021-platform-support-policy.md
@@ -39,7 +39,7 @@ Oracle Enterprise Linux | 5, 6, 7 | i386, x86_64 | rpm | RHEL 5
 Red Hat Enterprise Linux | 5, 6, 7 | i386, x86_64 | rpm | RHEL 5
 Solaris | 10, 11 | sparc, x86 | shar | Solaris 10
 Windows | 7, 8, 8.1, 2003R2, 2008, 2008R2, 2012, 2012R2 | x86, x86_64 | msi | Windows 2008R2
-Ubuntu Linux | 10.04, 12.04, 14.04 | x86, x86_64 | deb | Ubuntu 10.04
+Ubuntu Linux | 12.04, 14.04 | x86, x86_64 | deb | Ubuntu 12.04
 
 ### Tier 2 Support
 
@@ -48,6 +48,7 @@ Tier 2 supported platforms are those on which Omnitruck will serve packages, but
 * SUSE Linux Enterprise Server 10, 11
 * Scientific Linux 5.x, 6.x and 7.x (i386 and x86-64)
 * Debian Linux 6.x and 7.x
+* Ubuntu 10.04 (x86, x86_64)
 * Gentoo Linux (rolling release)
 * Arch Linux (rolling release)
 * Fedora (current non-EOL revisions)
@@ -75,7 +76,7 @@ Includes any of the add-ons (webui2/manage, push, etc.)
 
 ### Supported
 
-* Ubuntu 10.04LTS, 12.04LTS, 14.04LTS
+* Ubuntu 12.04LTS, 14.04LTS
 * RHEL 5.x, 6.x, 7.x
 * CentOS 5.x, 6.x, 7.x
 * Oracle Enterprise Linux 5.x, 6.x, 7.x
@@ -122,6 +123,7 @@ As platforms are added and removed, the timeline of the changes needs to be reco
 Platform | Change | References
 --- | --- | ---
 Solaris 9 | Removed October 2014 | http://www.oracle.com/us/support/library/lifetime-support-hardware-301321.pdf
+Ubuntu 10.04LTS | Moved to Tier 2 May 2015 | https://wiki.ubuntu.com/LTS
 
 ## Copyright
 


lamont-granquist:
Is it causing us that much pain to continue to build and test on 10.04?  We're a platform play and we have customers that rarely upgrade and we can support them better and have a wider audience if we support more operating systems, and we know that customers do not upgrade as often as they should and the reality is that sites often have ancient operating systems that still need support.

I just see a big difference between dropping Solaris 9 which was released in 2002 and finally went off support by Oracle, to dropping Ubuntu 10.10 which was released in 2010 and has gone off support by Canonical.  The latter has only been around for 5 years, which is not a long time in the system administration game, at large sites its perfectly reasonable to have distros kicking around for that long.  And a major point of evaluating configuration management systems and other IT systems is their support for your oldest distros.  The companies that don't support your oldest distros tend to not even get a phone call.

mattray:
It seems we should be consistent in our approach to deprecating platforms, even if some are relatively easy to continue supporting after the vendor has stopped. We could remove it from the supported platform list but continue to provide packages and check the download metrics to evaluate demand. Perhaps we should move it from Tier 1 to Tier 2 to signify this. While it's essentially semantics, it's recognition the platform may have issues we won't expend effort to address. I'd hate for our Support to get into a situation  where an issue arises that might require the distro vendor's assistance and having to tell a customer we can't help them. We clearly don't have to drop the platform immediately, but we should have a process.

thommay:
I think moving to tier 2 is a fair compromise, but I completely agree with Lamont that we shouldn't rush to remove things - it is **hard** to upgrade old OS versions when they're in production and when the cost to us is low we should bend over backwards to help our customers out.

smurawski:
I'm onboard with moving to tier 2 as well.

lamont-granquist:
So Tier 2 is generally for stuff like where we've shipped RedHat binaries to SuSE installs, and for Ubuntu 15.04 on release day using Ubuntu 14.04 -- where we don't build the package on that platform but it seems to work okay.

In this case to support 10.04 at all we have to build on 10.04, shipping 12.04 to 10.04 will not work -- so dropping to Tier 2 seems to effectively no support at all for omnitruck and breaking knife bootstrap.

There's also some question between support for omnitruck and support in cookbooks.  I can certainly see the argument that support in cookbooks can be dropped where it becomes burdensome, new cookbooks shouldn't be expected to accept 10.04 bugs, and its up to the 10.04 users to keep their own cookbooks on life support.  There's at least three or four different axis of support that I can see here:  omnitruck installs, cookbook support, core chef client support and internal chef support desk support.  The answers for all of those may be different on a distro-by-distro basis.

mattray:
I've updated the PR for 10.04 being Tier 2, but that doesn't fully address @lamont-granquist's points. I expect @schisamo's Release Engineering team will decide when and how they want to handle the 10.04 builds vs. 12.04 and 14.04. 

For now, we're basically acknowledging that 10.04 is no longer supported by Canonical so we can't call it Tier 1.

coderanger:
Part of the point of having a clear tiers system is help streamline all of those different facets of "supported". As a cookbook author, for example, I generally adopt the same lists when people ask what platforms I "support" (modulo "except linux-only" for most).

lamont-granquist:
Also we should probably consider 100% completing the process of getting off of 10.04 internally before deprecating the platform:

```
XXX.compute-1.amazonaws.com "fqdn": "solr-rspreprod-XXX.opscode.us",
XXX.compute-1.amazonaws.com "platform_version": "10.04",
XXX.compute-1.amazonaws.com "fqdn": "bifrost-rspreprod-XXX.opscode.us",
XXX.compute-1.amazonaws.com "platform_version": "10.04",
XXX.compute-1.amazonaws.com "platform_version": "10.04",
XXX.compute-1.amazonaws.com "fqdn": "community-rspreprod-XXX.opscode.us",
```

thommay:
We'd like to clarify exactly what Tier 2 entails before we proceed on merging this.

mattray:
Do we need to update the definition of Tier 2? It's in this document already.
https://github.com/mattray/chef-rfc/blob/ubuntu_10_04_removal/rfc021-platform-support-policy.md#tier-2-support

nathenharvey:
Tier 2 has been defined in RFC 21

lamont-granquist:
Okay, I just had a complete 180 degree change of heart.

While I feel empathy for the poor embattled sysadmins out there who may have hundreds or thousands of Ubuntu 10.04 boxes and be unable to upgrade in a timely matter, I finally found my SOX/PCI-DSS auditing hat in the back of the closet and put that on.

To support Ubuntu 10.04 means we have to build on 10.04 which means we'd have to have 10.04 infrastructure and implicitly have server images that were EOL and no longer receiving security updates.  That would not pass any kind of security audit that actually had usable teeth to it, so IMO Chef Software, Inc really cannot support any distro past its EOL date when the vendor is no longer shipping security updates.  Period.  Full Stop.

So, :shipit:

ranjib:
:+1:

adamhjk:
I agree with @lamont-granquist and what appears to be the majority of maintainers. We should not support platforms after the vendor EOLs it. Merge away, @chef/rfc-editors .

PR #137, opened 2015-06-30
Propose to support Debian 8 (Jessie) & move Debian to tier 1

rmoriz:
(updated:)

I'd like to propose Debian 8.x support ([What's new in Debian 8](https://www.debian.org/releases/stable/amd64/release-notes/ch-whats-new.en.html)).

Jessie was released in April 2015 and is the current stable release of Debian. Notable changes are the new `systemd` init system.

Related: 
Debian 8.0 bento boxes: https://github.com/chef/bento/commit/5e3ec117459263e9942b00027337d4fa2c5ae584

By popular request, I've also promoted Debian 6, 7, 8 to tier 1 support.

### DIFF
--- a/rfc021-platform-support-policy.md
+++ b/rfc021-platform-support-policy.md
@@ -33,6 +33,7 @@ Platform | Versions | Architectures | Package Format | Built on
 --- | --- | --- | --- | ---
 AIX | 6.1, 7.1 | ppc64 | bff | AIX 6.1
 CentOS | 5, 6, 7 | i386, x86_64 | rpm | RHEL 5
+Debian Linux | 6, 7, 8 | i386, x86_64 | deb | Debian 6
 FreeBSD | 9, 10 | i386, amd64 | pkg_add pkg | FreeBSD 9
 Mac OS X | 10.8, 10.9, 10.10 | x86_64 | dmg | Mac OS 10.8
 Oracle Enterprise Linux | 5, 6, 7 | i386, x86_64 | rpm | RHEL 5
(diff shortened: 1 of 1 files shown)


lamont-granquist:
:+1: on Tier 1

lamont-granquist:
And we already ship an omnitruck artifact that works on Debian 8.0 so I think we just need to declare it Tier 1 and then start adding it to cookbook testing matricies.  The biggest issue is systemd, but the Chef 12 Provider Resolver stuff already takes care of that in core chef.

schisamo:
We'll need to get some Debian 8 testers deployed also. Also is Debian 8 32-bit a thing people care about. The reason I ask is the project chose not to create Debian 8 32-bit AMIs:
https://wiki.debian.org/Cloud/AmazonEC2Image/Jessie

lamont-granquist:
32-bit is being slowly dropped by everyone, if they chose to make it difficult to get 32-bit images on AWS, then I'd suggest we follow their lead and not build testers.  omnitruck will still offer the 32-bit artifacts to anyone who installs it and i'll bet money it works fine.

jtimberman:
:+1: on making Debian a tier 1 platform.

thommay:
I'm :+1: on adding jessie, but I agree we should upgrade Debian to tier 1. @rmoriz please change the RFC to do this. 
@schisamo screw 32bit until someone asks for it convincingly, IMO.

rmoriz:
Which Debian releases should get tier 1 support? 7.x and 8.x or 6.x too?

schisamo:
Really surprised Debian isn't tier 1. This is probably a typo/mistake as it's already treated as tier 1 in our release infrastructure (same is true for ChefDK).

lamont-granquist:
It looks like we build on Debian 6, so Debian 6/7/8 are tier 1.   Debian 6 and 7 are tested on 32-bit and 64-bit so those arches are all tier 1.  Debian 8 64-bit only needs to get a tester added to be tier 1.

PR #139, opened 2015-07-07
add debian to tier 1

thommay:
Thanks to @rmoriz for the original proposal.


thommay:
Merged based on commentary on the original PR

PR #140, opened 2015-07-13
Clarify N-1 support policy.

lamont-granquist:
This textual change updates policy to how we are actually behaving.
This gets the messaging correct to Customers and Customer Support.
Since Chef 12 was released in Dec 2014 we did do one release of Chef 11 in
Jan 2015 with new DSC features.  In the past 6 months, however, we have only
released security updates to the chef 11.18.x branch (which is now on
11.18.12 and soon to be 11.18.14).  We did bundle an updated CA certs
package to work around the omission of the root cert that signed all of
Amazon AWSs certs and with AWS SSL certs being entirely broken.  It does
not appear that a new minor release of Chef 11 is likely any time soon
with any bugfixes to chef client core, critical or otherwise.

Users who read "critical bug fixes" and believe that the bug which
ruined their weekend is "critical" and should be fixed in 11.x without
them upgrading to 12 will be disappointed unless we change this policy.

While we clearly reserve the right to push a new 11.x version to fix
earthshattering-the-world-is-broken types of bugs, or to address bugs
which have significant partner benefit, I've dropped that language
entirely since for most users it can only be misconstrued.

For the Customers who feel like this drops support for them, if you look
at the actual behavior of Chef 11 releases, it is clear that this
support was not occurring.  The easiest solution for customers would be
to upgrade to Chef 12 where they will get both security and bug fixes.

Alternatively, the policy that we adopted with Chef 10 backports was to
get a major customer adopting the backporting of fixes from 11 to 10 and
they were made external committers.  If there's interest in the
community by Chef 11 customers who feel it is easier for them to support
backporting to Chef 11 than to get their infrastructure onto Chef 12,
then those Customers should step forwards.

### DIFF
--- a/rfc015-chef-12.md
+++ b/rfc015-chef-12.md
@@ -23,7 +23,7 @@ At a given time:
 
 * `Latest` and `Latest - 1` major versions of Chef Client is supported by Chef community.
 * `Latest` version includes all the enhancements and bug fixes.
-* Security fixes only are backported to `Latest - 1`.
+* Security fixes are backported to `Latest - 1`.
 * Contributions to `Latest - 1` version are always welcome.
 * As Chef Community, we use and strongly recommend using omnibus packages for Chef Client. We still help out to folks who are doing custom installations of Chef Client but on a 'best effort' basis.
 
--- a/rfc015-chef-12.md
+++ b/rfc015-chef-12.md
@@ -23,7 +23,7 @@ At a given time:
 
 * `Latest` and `Latest - 1` major versions of Chef Client is supported by Chef community.
 * `Latest` version includes all the enhancements and bug fixes.
-* The major bug fixes and security fixes are backported to `Latest - 1`.
+* Security fixes only are backported to `Latest - 1`.
 * Contributions to `Latest - 1` version are always welcome.
 * As Chef Community, we use and strongly recommend using omnibus packages for Chef Client. We still help out to folks who are doing custom installations of Chef Client but on a 'best effort' basis.
 


coderanger:
I don't think `only` is the right word there, since feature backports will be accepted as you say. Maybe something more like "Chef Software will provide security fixes, major features will be evaluated on a case by case basis with the community."?

lamont-granquist:
The problem is that it still allows for lawyers to read that however they like.  To the vast majority of Chef customers "only security updates" is completely accurate as to how we're actually behaving for whatever bug they have in front of them.

coderanger:
I think we need to find a balance between setting expectations that features are not, in general, backported while also making it clear that people willing to do the work should feel empowered to.

lamont-granquist:
Okay, I can see that as a point to make.  Still Chef Inc initiated work is only for security releases right now.  Anyone is free to submit a PR with 11-stable as a base.  I ignored that because we're just not seeing it, but obviously we'd merge it if anyone cut such a PR.  If that's the concern, then the words can probably be updated to fit that.

lamont-granquist:
Okay I dropped 'only' and the next line is "Contributions to `Latest - 1` version are always welcome."

coderanger:
:+1: Seems clearer. As this is a pretty simple thing does @adamhjk want to fiat-accept before the meeting and save a few minutes?

coderanger:
(or @thommay by way of delegation)

thommay:
I think this is probably worth at least mentioning in the meeting so that the policy is clearly understood.

adamhjk:
:+1: I have /deciderated

PR #143, opened 2015-07-23
DRAFT: RFC for Resource#current_value

jkeiser:
CC @coderanger @danielsdeleo @someara @lamont-granquist @jaym

Until we've worked through the use cases @coderanger has brought up (I haven't had time this week), this is on hold.

### DIFF
--- a/new/resource-current-value.md
+++ b/new/resource-current-value.md
@@ -1,0 +1,98 @@
+---
+RFC: unassigned
+Author: John Keiser <jkeiser@chef.io
+Status: Draft
+Type: Standards Track
+---
+
+# Allow Current Value To Be Read From Resource
+
+Resources get a `current_value` method, which discovers the current value of
+the resource.
+
+## Motivation
+
+    As a Chef user,
+    I want to read the current value of certain resources (reading AWS object ids is a prime example),
+    so that I can do custom things in my recipes that the resource writers didn't think about.
+
+    As a Chef user,
+    I want to read the current value of certain resources,
+    So that I can make decisions in my recipes (such as only_if) based on actual data.
+
+    As a Chef resource developer,
+    I want to be able to expose all the read code I wrote to my users,
+    So I don't have to tell them to copy and paste it.
+
+## Specification
+
+All Resources get a method, `current_resource`, that returns the current value
+as a new resource. It does not cache this value. It returns `nil` when the
+resource does not exist. By default, it is implemented by creating the provider
+for one of the resource's actions, calling `load_current_resource` on that.
+
+This is additive: it is fully backwards-compatible, and even applies to old
+resources (not just ones with the new DSL using `property` and
+`load_current_value`).
+
+An example use:
+
+```ruby
+# Clean up after old versions of the blarghle gem that don't clean up after
+# themselves
+execute 'rm -rf /var/blarghle/data' do
+  only_if { g = gem_package('blarghle').current_resource; g && g.version.to_f < 2.0 }
+end
+
+gem_package 'blarghle' do
+  version '>= 2.0'
+end
+```
+
+### Discussion: *Should* Resource State Be Easy To Get?
+
+There is a central concern in Chef around the compile/converge model. When you
+write this:
+
+```ruby
+package 'httpd'
+
+service 'httpd' do
+  action [ :enable, :start ]
+  only_if File.stat('/var/www/html').owner == 'root'
+end
+```
+
+The httpd package is not installed until converge time (*after* both package and
+service are defined, so the recipe fails to compile (since /var/www/html is
+not yet installed), the `stat` call raises an exception. The same problem will
+occur if the user writes
+`only_if directory('/var/www/html').current_resource.owner == 'jkeiser'`.
+
+This causes the users first true encounter with the compile/converge model, and
+their first encounter with this is generally a frustrating and despair-inducing
+one. The answer, of course, is to put the { } in a block in `only_if`, so that
+it will be executed late, but the user doesn't know that when they write this
+code.
+
+Why is that an issue? The argument could be made that making state *too easy*
+to access will encourage more users down this road, into heartbreak. By making
+state hard to access at all, we have made it less likely they will access it in
+the wrong place.
+
+This RFC's position is that users who want to do this will find a way,
+*plus* have the heartache of having to learn Ruby APIs to rewrite code that
+*already exists inside a function in the Provider for the resource*. So the
+cost being mitigated, the *benefit* is definitely large enough that users want
+it. The Provisioning AWS driver, for example, has an `aws_object` method because
+users were demanding some way to access current state. This resolution,
+compile/convergey though it was, was much less burdensome than forcing them to
+learn how to construct the AWS APIs and replicate the code in Provisioning that
+would look up the objects.
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.
--- a/new/rfc-template.md
+++ b/new/rfc-template.md
@@ -1,11 +1,10 @@
 ---
 RFC: unassigned
-Author: Alan Smithee <asmithee@example.com>
+Author: John Keiser <jkeiser@chef.io
 Status: Draft
-Type: <Standards Track, Informational, Process>
-<Replaces: RFCxxx>
-<Tracking:>
-<  - https://github.com/chef/chef/issues/X>
+Type: Standards Track
+Tracking:
+  - https://github.com/chef/chef/issues/X
 ---
 
 # Title


new/rfc-template.md
```diff
@@ -1,11 +1,10 @@
 ---
 RFC: unassigned
-Author: Alan Smithee <asmithee@example.com>
+Author: John Keiser <jkeiser@chef.io
```
coderanger: Probably not intended to commit this.

coderanger:
I don't know if this should be a global API. I agree the general idea is useful, but I'm not sure this needs to be in core. The examples with core resources feel kind of contrived, this really seems to come up more in super-advanced use cases like chef-provisioning and [poise-ruby/python](https://github.com/poise/poise-ruby/blob/master/lib/poise_ruby/resources/ruby_runtime.rb#L58) (if we extend this to the general case of "resources having output data". For the example using `gem_package` to work correctly it would be relying heavily on cloning since you aren't setting `gem_binary` and that might have been used earlier in a previous definition of the resource. This would basically mean this cannot be used safely in library cookbooks as a resource needs to be defined twice, once for the collection and once for output data.

Basically this feels like a footgun unless we drop down a level and come up with a more general solution for resources having output data, of which current state will be a specific use case.

coderanger:
Copying from IRC for the future:

1. A way to get a reference version of a resource that does not cause it to also converge
2. A way to get the output of multiple actions that occur on a resource
3. A way to handle the degenerate case of #2 with a single action in a simple way (90% case)

coderanger:
Also a list of use cases we should map out:

1. Current value of `version` for a `gem_package` where the package resource might or might not have a `gem_binary` (control property) defined earlier in the recipe.
2. Getting the AWS ID for a newly created `machine` resource.
3. Getting `stdout` for an `execute` resource.
4. Getting the `ruby_binary` path for a `ruby_runtime` resource.

lamont-granquist:
I'd like to be able to say:

```ruby
package package_installable?("git-core") ? "git-core" : "git"
```

`package_installable?` would then be some sugar around using the package resource to see if the arg was an available package or not.  I want to be able to re-use the package resource so that I don't have to think about yum and apt and everything else myself.  I also don't want to trigger the installation and then rescue or `epic_fail true`.  And this is cleaner than platform/platform_family/platform_version case statements and is really the behavior that I want rather than having to think about which debian/ubuntu/linuxmint platform_versions renamed 'git-core' to 'git'.

new/resource-current-value.md
```diff
@@ -0,0 +1,98 @@
+---
+RFC: unassigned
+Author: John Keiser <jkeiser@chef.io
+Status: Draft
+Type: Standards Track
+---
+
+# Allow Current Value To Be Read From Resource
```
martinb3: Did you decide to call this `current_resource` instead of `current_value` after writing this title and the paragraph of text under it? I think these might just need to be updated to:

```
# Allow Current Resource to be Read from Resource Collection

Resources get a `current_resource` method, which discovers an existing resource in the resource collection
```

new/resource-current-value.md
```diff
@@ -0,0 +1,98 @@
...
+    As a Chef resource developer,
+    I want to be able to expose all the read code I wrote to my users,
+    So I don't have to tell them to copy and paste it.
+
+## Specification
+
+All Resources get a method, `current_resource`, that returns the current value
+as a new resource. It does not cache this value. It returns `nil` when the
+resource does not exist. By default, it is implemented by creating the provider
+for one of the resource's actions, calling `load_current_resource` on that.
```
martinb3: I was expecting this would be implemented by returning a reference to the existing resource, not to load a new version of it using the provider. I feel like some providers may have (albeit a bad practice) existing state that `load_current_resource` may not capture. Is there a reason to prefer creating a new resource object over searching the resource collection by name and returning the existing resource object?

martinb3:
Just reading the other comments here, and catching up.

I think my use case is a little less elaborate than @coderanger's described one. I'd simply like to see what parameters were set on a resource, by looking it up. For example, if a resource installs some files at a particular path, I'd love to fetch that path later in a recipe in order to some additional work on it.

Regarding @lamont-granquist's example, I couldn't tell if `package_installable?` happens at converge or compile time. That example feels to me like yet another case I hadn't considered -- drawing on a resource to query some information. Perhaps that would be better suited to something like chef-sugar? Querying state doesn't feel idempotent when used in that example.

Overall, I'd be happy with this RFC being implemented, FWIW. :+1: for writing it @jkeiser!

new/resource-current-value.md
jkeiser: The intent is to read the actual value from the system, not from the resource collection, actually.

new/resource-current-value.md
jkeiser: Because the existing resource in the collection is a statement of what the user wants to be true (a desired value), not what is actually true (current_resource).

martinb3: Ah. I guess I'm sometimes looking for what the desired value is, like in my example below about the path that some files are going to be copied to.

new/resource-current-value.md
martinb3: But you left in `current_value` when the examples further down call the new DSL `current_resource`, right? Or is this proposing both?

new/resource-current-value.md
jkeiser: Yep. There are reasons to want to find out desired state :) This isn't that, but once we've worked through all the issues I suspect we'll have both.

new/resource-current-value.md
jkeiser: Just current_value--which corresponds to the new `load_current_value` API. current_resource might be the right word, but in that case it would still correspond to `Provider.load_current_resource`, an existing API which lets you load the current value of the resource from the system.

new/resource-current-value.md
martinb3: :+1:

lamont-granquist:
So `package_installable?` clearly would need to be compile-time and the method itself would probably be in chef-sugar, but right now there's no good way to implement that in chef-sugar.   This RFC should allow that method to actually get written.   And again the problem is that in order to write that you'd want an abstraction over yum/apt/etc packages and then be able to query state without modifying it.  That is what is accomplished by being able to grab a resource and inspect the load_current_resource state.  Otherwise we have to right some query tools which re-implement a fascade over yum/apt/etc and that's exactly what we already have with the package pseudo-resource, so I want to re-use that.

And querying state like that is perfectly idempotent.  But instead of having awful case statements around platform and platform_version (and having to take into account debian, ubuntu and linuxmint and where exactly the pv on all of those uniquely change) you just duck-type the operating system.   In pseudo-code it has this kind of shape:

```
install_git_core if platform.respond_to?(:install_git_core)
```

For any given platform and platform_version the boolean value it returns is deterministic, and overall its idempotent and convergent, but it allows you to query the operating system to find out how it behaves rather than having to manage a case statement with knowledge about how every platform and platform_version behaves.  Because ultimately I really don't care.  If I have a platform which lets me install 'git-core' then I know I'm on old-debian where 'git' is really 'gnu interactive toolset' and I need to use 'git-core', otherwise I just use 'git' and that works for every case I'm aware of.  If we discover a new platform and platform_versions which have a similar lineage to debian then it'll just work without modification (although the better win is that we already have that platform, which is linuxmint, and i think i'm the only person in the Chef community who ever bothers to convert ubuntu versions to linuxmint versions whenever i find them).

lamont-granquist:
Also -dev vs -devel vs none for header files gets a lot easier:

```ruby
package "foo"
package "foo-dev" if package_installable?("foo-dev")
package "foo-devel" if package_installable?("foo-devel")
```

That'll support fedora, amazon, all the rhels, debian, ubuntu, raspbian, linuxmint and arch at least right out of the box, and has a good shot of solaris, gentoo, etc i think.  All without a stupid case statement around the platform.

lamont-granquist:
There's another use case here, which is minitest-chef-handler.  It already implements this kind of behavior using resources and load_current_resource in the provider in order to generate matchers for minitest.  So if you're writing a rule that a file should have mode 0644 or whatever, internally minitest will create a file resource and call load_current_resource and then pull the mode out.

If we turn this into a public API it will create less breakage for minitest-chef-handler and we should be able to integrate similar chef resources based matchers with rspec.

martinb3:
Maybe idempotent is a poor word choice on my part, but something about that example does not feel declarative about the state of the system. In any case, I withdraw the comment since I can't really justify why it strikes me that way.

My use case would be to look at the desired state of a resource that has already been declared, not the current state of the resource, so I'd love to see a way for us to look up both. I would think that if we looked up the resource in the resource collection, it would contain both the result of load_current_resource data as well as the desired state (new_resource.params).

martinb3:
Perhaps I should just use the existing `#resources` DSL functionality for my use case, and leave this to the other use cases. There's just something cleaner about `package('foo').version` over `resources(:package, 'foo').version`.

lamont-granquist:
Its still also declarative, but it is giving you a much better if-then language to use:  "if you can install git-core then it should be installed" or "if you can install zlib-devel, then it should be installed".  Rather than "if you are on ubuntu older than X, or linuxmint older than Y, or debian older than X, then git-core should be installed", which is tedious.

lamont-granquist:
seems this snuck into the code:

https://github.com/chef/chef/blob/3baa2c1abc07a21acf9dd1430d4b97700c2835a1/lib/chef/resource.rb#L1161-L1176

should probably close this RFC, we could open the RFC again if there were issues with the current_value implementation there.

thommay:
Agreed.

lamont-granquist:
We have load_current_value and current_value, and this RFC is inaccurate as to how they've worked now for years.

PR #145, opened 2015-08-11
Auto-vivify plugin configuration hashes.

mcquin:
We can auto-vivify hashes easily, without using `method_missing`. This change proposes auto-vivifying plugin configuration hashes at all levels, instead of only at the plugin-name level. This also simplifies enforcing Symbol-only keys and look up.

\cc @lamont-granquist @danielsdeleo @jkeiser @coderanger @btm

### DIFF
--- a/rfc053-ohai-config.md
+++ b/rfc053-ohai-config.md
@@ -31,7 +31,7 @@ The existing hints system was designed to provide facts about a system that Ohai
 
 When a 'hint_name.json' file exists in the directory specified by `Ohai::Config[:hints_path]`, `Ohai::Hints.hint?(hint_name)` will return non-nil. If the file contains JSON data, it will be returned as a hash. If the file is empty, an empty hash will be returned.
 
-The hints system should only be used by other tools to assist Ohai in collecting data. It should not otherwise be used to configure the behavior or Ohai or its plugins. That is the purpose of the new configuration system. The hints system may be combined with the configuration system in the future and deprecated for general ease of use of Ohai. 
+The hints system should only be used by other tools to assist Ohai in collecting data. It should not otherwise be used to configure the behavior or Ohai or its plugins. That is the purpose of the new configuration system. The hints system may be combined with the configuration system in the future and deprecated for general ease of use of Ohai.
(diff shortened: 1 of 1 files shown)


coderanger:
This would break anything that checks for a key via code like `if config[:foo]` since accessing the key creates and stores it. `include?` should still work, but can you perhaps including some motivation for this change?

lamont-granquist:
The problem is allowing `Ohai[:plugin][:dmi]` to autovivify to a hash without the user having to create it which seems to be necessary and required by the RFC, while blocking `Ohai[:plugin][:dmi][:foo][:bar]` from autovivifying would additionally require an autovivifying Hash that stopped after 2 levels.

That's clearly awful code and UX, so it should either be all or nothing, so the autovivifying of the top levels is driving making the whole thing autovivify.

Unless there's a way to create and prepopulate the top level hashes without autovivifying?  I didn't look at the code, but that would be the way to eliminate all autovivification....

mcquin:
As far as I can tell, we'll already have to auto-vivify the hash at the plugin-name level. 

```ruby
config_context :ohai do
  default :plugin, {}
end
```

```
pry> ohai.plugin[:foo][:bar] = :baz
NoMethodError: undefined method `[]=' for nil:NilClass
```

If we choose to only auto-vivify at the plugin-name level, we still have the issue of `if Ohai.config[:plugin][:foo]` instantiating the `:foo` key to a new Hash. We could do something gross and modify `[](key)` such that if `{}` then return `nil`, or handle that in a DSL method. I was going to need to likely create a subclass of `Hash` and override `[]=(key, value)` to enforce the keys-as-Symbols rule anyway.

coderanger:
If we did something like `config[:plugin] = Hash.new {|k, h| h[k] = Hash.new}` you get the behavior that `config[:plugin][:dmi]` vivifies and is a hash, but `config[:plugin][:dmi][:foo]` would be nil if unset.

mcquin:
Okay. That's fair.

lamont-granquist:
Guess that works out okay.

PR #146, opened 2015-08-11
Plugin configuration getter DSL.

mcquin:
Add a new DSL method for accessing plugin configuration settings. Since plugins can be configured in two different ways (as an explicit Hash or an underscore-deliminated Symbol), having one DSL method for accessing plugin configuration will simplify how plugin authors author plugins.

\cc @coderanger @lamont-granquist @danielsdeleo @jkeiser @btm

### DIFF
--- a/rfc053-ohai-config.md
+++ b/rfc053-ohai-config.md
@@ -31,7 +31,7 @@ The existing hints system was designed to provide facts about a system that Ohai
 
 When a 'hint_name.json' file exists in the directory specified by `Ohai::Config[:hints_path]`, `Ohai::Hints.hint?(hint_name)` will return non-nil. If the file contains JSON data, it will be returned as a hash. If the file is empty, an empty hash will be returned.
 
-The hints system should only be used by other tools to assist Ohai in collecting data. It should not otherwise be used to configure the behavior or Ohai or its plugins. That is the purpose of the new configuration system. The hints system may be combined with the configuration system in the future and deprecated for general ease of use of Ohai. 
+The hints system should only be used by other tools to assist Ohai in collecting data. It should not otherwise be used to configure the behavior or Ohai or its plugins. That is the purpose of the new configuration system. The hints system may be combined with the configuration system in the future and deprecated for general ease of use of Ohai.
 
 ## Specification
 
@@ -110,10 +110,41 @@ ohai[:plugin][:dmi][:id] = { :cpu => true, :memory => true, :disk => false }
 Ohai.config[:plugin][:dmi][:id][:disk] => false
 ```
 
+### Plugin configuration lookup DSL method
+
+To facilitate lookup of plugin configuration options, a new method will be
+added to the plugin DSL: `configuration`. The `configuration` method
+accepts as parameter the configuration option to fetch (as a Symbol).
+
+For example, a plugin author may write
+
+```ruby
+Ohai.plugin(:Foo) do
+  provides 'foo'
+
+  collect_data do
+    if configuration(:foo_option)
+      # ...
+    end
+  end
+end
+```
+
+A user would configure
+
+```ruby
+ohai.plugin[:foo][:foo_option] = true
+```
+
+The `configuration` method is intended only to be used to access top-level
+configuration options. If the plugin author chooses to require plugin configuration
+to be a Hash, the author must access those configuration options as Hash elements.
+That is, `configuration(:one_two_three)` cannot be used to look up
+`plugin[:foo][:one][:two][:three]`.
+
 ## Copyright
 
 This work is in the public domain. In jurisdictions that do not allow for this,
 this work is available under CC0. To the extent possible under law, the person
 who associated CC0 with this work has waived all copyright and related or
 neighboring rights to this work.
-


coderanger:
Are you saying that `[:id_cpu] = true` and `[:id] = { cpu: true }` would be equivalent? Because that's not a `.` delimited string.

mcquin:
@coderanger They're equivalent configurations. Plugins configured without setting explicit Hashes are made to use `_` to separate keys. This is to get around the annoyance of needing to check for `Ohai.config[:plugin][:foo][:set_this_way]` and `Ohai.config[:plugin][:foo][:set][:this][:way]` in your `:Foo` plugin.

coderanger:
It seems weird to have four different ways to denominate config values. Underscore-separated, nested hash, array of keys, period-separated?

mcquin:
I thought about accessing config values with an underscore, but that won't work in cases where the underscore actually matters. Example: `Ohai.config[:plugin][:foo][:baa_baa] = { :black_sheep => true }` and `Ohai.config[:plugin][:foo][:baa_baa_black_sheep]`. If you tried to access with `configuration("baa_baa_black_sheep")` we wouldn't know which underscore to split on. If the user had configured with the first configuration syntax, it'd take some guesswork to match.

coderanger:
I just still don't understand why this is such a big deal we need so many ways to interact with config data. Ohai plugins are not something usually so complex they likely need this level of DSL sugar I would think.

coderanger:
Where are the cases that have so many deeply nested config values that this sugar is even a big deal? And could we just not do that instead?

mcquin:
I'm trying to provide one method for getting configuration settings.

Here's my logic: If users can set `ohai.plugin[:foo][:key1] = { :key2 => "value" }` or `ohai.plugin[:foo][:key1_key2] = "value"`, then it'd be really nice to have one way to look up either specification (since they're represented differently in a Hash). Otherwise, every time you want to check the configuration you'd have to check both ways of setting it.

Accessing with an array of keys makes the most sense, giving us either a path to follow or to construct a "superkey" from by joining with `_`. Accessing with a string key delimited by `_` doesn't make sense for all cases (see: [baa_baa_black_sheep](https://github.com/chef/chef-rfc/pull/146#issuecomment-130085687)). If supporting a String delimited by `.` is too much sugar, we can drop that functionality.

coderanger:
re: "If users can set...", why do we have two ways to set things? Just pick one or the other?

mcquin:
[That's how it's specified in the RFC](https://github.com/chef/chef-rfc/blob/mcquin/rfc053/plugin-config-lookup/rfc053-ohai-config.md#configuration-hash-nesting)

coderanger:
It doesn't say the that two formats should be interchangable, at least not that I can see.

coderanger:
For any given plugin I would think it would use only one or the other, and the opposite style key(s) would simply not be valid.

mcquin:
In my opinion, being able to configure a plugin either way is A Nice Thing To Do. It's nice from a UX perspective depending on how much there is to configure (probably not a ton, but maybe enough that a hash is helpful), or how much you actually want to configure.

coderanger:
I think that level of sugar could wait until we get more experience with configurable ohai plugins both as more are written and as we see how users interact with the configuration system (via `knife`, `chef-provisioning`, etc). As it stands in the spec, I don't think this would block anything, right? (the original justification to move forward on the config format was it was a blocker on some important functionality)

mcquin:
Additionally, it'll be annoying to have plugins configured with a different syntax depending on who authored them. Allowing users to choose their own configuration syntax in their files means they can be consistent for themselves.

stevendanna:
> Additionally, it'll be annoying to have plugins configured with a different syntax depending on who authored them. Allowing users to choose their own configuration syntax in their files means they can be consistent for themselves.

As a user of this, I *really* don't want to have to constantly figure out which plugins are using hashes and which are using underscores to group their options.  I'd be :+1: on removing the "." syntax for lookups though since I don't mind making plugin authors be as explicit as possible about what they want.

coderanger:
The RFC does already state that single-layer configuration using keys with underscores is preferred, and as the people on this pull request will be writing the vast majority of complex, configurable plugins I'm not sure we need to worry too much about style drift :)

mcquin:
> and as the people on this pull request will be writing the vast majority of complex, configurable plugins I'm not sure we need to worry too much about style drift :)

That's not a valid assumption.

thommay:
I agree with @stevendanna here - we should just remove the hash and dot syntax and go with underscores across the board.

lamont-granquist:
I don't think the original text intended for the two methods to be equivalent.  It was simply steering plugin authors away from using hashes with a style guide recommendation, and then recommended that if plugin authors did use hashes that users should assign the hash -- all due to there being no autoviv.

Simpler solution to this would be to just add a method that took a single arg:  `configuration(:id_cpu)` which steered plugin authors toward using a single setting.

mcquin:
Updated the proposal so that `configuration` only takes an option as a Symbol and only looks up the top-level Symbol in the plugin's configuration (it won't work for nested options/configuration Hashes).

lamont-granquist:
:+1:

thommay:
:+1:

coderanger:
:+1:

jonlives:
:+1:

thommay:
This is deciderated. @chef/rfc-editors, make it so!

PR #147, opened 2015-09-10
Cookbook Quality Metrics

nathenharvey:
As a cookbook author,
    I want to write cookbooks that can be shared with and easily used by the community
    so that it is easier to get going with Chef.

    As a Supermarket user,
    I want to identify the relative quality of a given cookbook,
    so that I can spend less time looking for a good cookbook and more time automating.

We need an objective, automatic way to assess certain qualities of a cookbook.

### DIFF
--- a/new/cookbook_quality_metrics.md
+++ b/new/cookbook_quality_metrics.md
@@ -40,6 +40,12 @@ The lifecycle of a metric will be:
 * *Implemented* - this metric has been implemented and is contributing to the scores shown on the Supermarket.
 * *Closed* - this metric has not been accepted or has been removed from the system.
 
+### Viewing Cookbook Scores
+
+The [Supermarket](https://supermarket.chef.io) will show a total score for each cookbook.
+
+Each cookbook listed on the Supermarket will have a tab or other view showing how it's score was generated.
+
 ## Copyright
 
 This work is in the public domain. In jurisdictions that do not allow for this,
--- a/new/cookbook_quality_metrics.md
+++ b/new/cookbook_quality_metrics.md
@@ -10,9 +10,9 @@ Type: Informational
 
 # Cookbook Quality Metrics
 
-The question of what makes a good cookbook has been asked since the dawn of Chef.  As a community we've struggled with answering this in part because each recipe is unique to the specific requirements it is trying to meet.  Allowing Supermarket users to rate cookbooks is one thing that we've tried in the past.  This did not work because you simply do not have enough knowledge about the quality of a cookbook when you discover and download it from the Supermarket.  Similarly, download counts cannot be trusted because of [the way some older versions of the berkshelf-api server work].
+The question of what makes a good cookbook has been asked since the dawn of Chef.  As a community we've struggled with answering this in part because each recipe is unique to the specific requirements it is trying to meet.  Allowing Supermarket users to rate cookbooks is one thing that we've tried in the past.  This did not work because you simply do not have enough knowledge about the quality of a cookbook when you discover and download it from the Supermarket.  Similarly, download counts cannot be trusted because of [the way some older versions of the berkshelf-api server work](https://www.chef.io/blog/2015/01/21/those-pesky-supermarket-download-counts/).
 
-The Supermarket should make it easy for anyone to find quality cookbooks.  In order to do so, we must agree on some qualities that indicate whether or not a cookbook is considered to have high quality.  Ideally, these qualities are objective and able to be determined automatically.
+The Supermarket should make it easy for anyone to find quality cookbooks.  In order to do so, we must agree on some qualities that indicate whether or not a cookbook is considered to have high quality.  Ideally, these qualities are objective and able to be determined automatically.  These qualities should be defined and agreed on by the community.
 
 ## Motivation
 
@@ -26,56 +26,19 @@ The Supermarket should make it easy for anyone to find quality cookbooks.  In or
 
 ## Specification
 
-The metrics that determine the relative quality of a cookbook are listed below.
+A [Cookbook Quality Metrics](https://github.com/chef-cookbooks/cookbook-quality-metrics) system will be implemented to provide a quality score for each cookbook.  This score will be visible on the Supermarket.
 
-* It converges without error.
+* The community will be able to collaborate on the metrics.
+* Each quality metric will add or remove points from a cookbook's score.
 
-    ```
-    it 'converges successfully' do
-      expect { chef_run }.to_not raise_error
-    end
-    ```
+We will collaborate on quality metrics in the [Cookbook Quality Metrics](https://github.com/chef-cookbooks/cookbook-quality-metrics) repository.
 
-* It converges without error using Fauxhai data for each supported platform.
+The lifecycle of a metric will be:
 
-* It links to the source.
-
-    ```
-    # metadata.rb
-    source_url 'http://github.com/chef-cookbooks/mysql'
-    ```
-
-* It links to the issue tracker.
-
-    ```
-    # metadata.rb
-    issues_url 'https://github.com/chef-cookbooks/mysql/issues'
-    ```
-
-
-* It is updated and released on a regular basis
-  * time since last release
-  * time since last change in source code repository
-  * number of commits not released
-* It includes a README.md with more than the boilerplate copy.
-* It includes a MAINTAINERS.md with contact information for each authorized MAINTAINER and Supermarket Collaborator.
-  * There is more than one maintainer.
-  * What is the optimal range?  5-7 maintainers?
-  * We should not call it the 'bus factor'
-* It includes a version number the conforms to the [SemVer specification](http://semver.org/), e.g., X.Y.Z
-* It includes a .kitchen.yml which includes a platform declaration for each of the platforms listed as a `supports` in the cookbook's metadata.
-* `kitchen test` completes successfully.
-* It includes an open-source license (LICENSE.md).
-* It includes rspec-based unit tests that pass.
-* It passes Foodcritic (`foodcritic --tags correctness,metadata`) and Rubocop (`rubocop --lint`), with allowances for custom rule specifications.
-* The README includes a badge indicating status of Code Climate and TravisCI tests.
-  * Code Climate for Foodcritic and Rubocop
-  * TravisCI for ChefSpec
-* It includes a `name` in the metadata.
-* It includes a `version` in the metadata.
-* It includes a `Berksfile` that lists all dependencies.
-* It includes a CHANGELOG.md that is updated with each release.
-* It is published to the [Supermarket](https://supermarket.chef.io).
+* *Draft* - this is a proposed metric, ready for community discussion and approval.
+* *Accepted* - this metric has been accepted and merged into the master branch.
+* *Implemented* - this metric has been implemented and is contributing to the scores shown on the Supermarket.
+* *Closed* - this metric has not been accepted or has been removed from the system.
 
 ## Copyright
 
--- a/new/cookbook_quality_metrics.md
+++ b/new/cookbook_quality_metrics.md
@@ -59,13 +59,15 @@ The metrics that determine the relative quality of a cookbook are listed below.
   * number of commits not released
 * It includes a README.md with more than the boilerplate copy.
 * It includes a MAINTAINERS.md with contact information for each authorized MAINTAINER and Supermarket Collaborator.
+  * There is more than one maintainer.
+  * What is the optimal range?  5-7 maintainers?
+  * We should not call it the 'bus factor'
 * It includes a version number the conforms to the [SemVer specification](http://semver.org/), e.g., X.Y.Z
 * It includes a .kitchen.yml which includes a platform declaration for each of the platforms listed as a `supports` in the cookbook's metadata.
 * `kitchen test` completes successfully.
 * It includes an open-source license (LICENSE.md).
 * It includes rspec-based unit tests that pass.
-* It passes Foodcritic and Rubocop, with allowances for custom rule specifications.
-  * Should the Foodcritic and Rubocop standards be agreed?
+* It passes Foodcritic (`foodcritic --tags correctness,metadata`) and Rubocop (`rubocop --lint`), with allowances for custom rule specifications.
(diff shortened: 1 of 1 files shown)


new/cookbook_quality_metrics.md
```diff
@@ -0,0 +1,48 @@
...
+    As a Supermarket user,
+    I want to identify the relative quality of a given cookbook,
+    so that I can spend less time looking for a good cookbook and more time automating.
+
+## Specification
+
+A [Cookbook Quality Metrics](https://github.com/chef-cookbooks/cookbook-quality-metrics) system will be implemented to provide a quality score for each cookbook.  This score will be visible on the Supermarket.
+
+* The community will be able to collaborate on the metrics.
+* Each quality metric will add or remove points from a cookbook's score.
```
coderanger: Major :-1: on anything numeric. As a starting point we can have a bunch of flags and individual metrics but going from that to any kind of overall "object" quality score seems like a bridge too far from here.

nathwill:
:clap: super happy to see a move toward defining and surfacing community standards for quality measures, thanks so much @nathenharvey!

coderanger:
Overall :+1: on the idea of defining a format for us to work on what metrics should be tracked. I would probably add a template in the same way that RFC0 has a template for future RFCs. Not thrilled about the concept of using the metrics to create a composite score. Defining stuff like "bad practice X" is _worth_ the same amount as "good practice Y" seems like it will either be super bike-shed-y or won't reflect reality and will push people to game the system.

nathenharvey:
There is a template in that repo - https://github.com/chef-cookbooks/cookbook-quality-metrics/blob/master/quality-metrics/template.md

That repo takes a lot of goodness from this one and implements it for the quality metrics.

new/cookbook_quality_metrics.md
jonlives: I don't have super strong opinions on the numbering thing explicitly, but I do think that we need to have a simple and easy to view quality indicator - people aren't going to want to have to read and understand a bunch of flags and metrics when all they want is to install Apache.

What about something like they do with food packaging, where we calculate an aggregated "traffic light" score of Red, Amber or Green, with a breakdown of what composes that score (eg the nutritional information table) for those who want to dive in deeper?

jonlives:
As mentioned above I'm in favour of having an aggregate score alongside the more detailed breakdown. The good thing is that if this process is open etc, as suggested, then if it turns out someone is gaming the system or we're bikeshedding specific metrics, we can always adjust the process to account for this.

I think we need to be realistic about the fact that this process will not be perfect in its first iteration and that we're going to end up tweaking  and changing it as people start to use it.

I'd also argue that if people are attempting to game the system that's not *entirely* a bad thing, as it at least means they're actively working on their cookbooks :p

jonlives:
One other comment on the existing metrics in the cookbooks quality repo - all of the metrics so far assign a "1" positive score and a much greater negative score if any of the metrics is not met.

My worry is that this will mean a "good" cookbook with a perfect score would have a score of "5" whereas a cookbook failing even one metric would typically have a score in the negative tens. 

I think we need to balance the positive / negative weighting here a little more so that good scores and comparably bad scores don't feel so out of whack. One suggestion would be that the negative weighing can only ever be twice the positive. For example, a perfect score would be 5, and a score failing one metric would then be 3 (the negative weighting being 2 * -1)

new/cookbook_quality_metrics.md
martinb3: I agree that specific scores is probably too much. I think I'd most like a bunch of true/false metrics, like the ones already in that repo, and then maybe:

Green - all metrics passed - this looks good
Yellow - 1 or more metrics did not pass - proceed with caution
Red - 3+ metrics did not pass - you should probably avoid this

martinb3:
> I think we need to balance the positive / negative weighting here a little more so that good scores and comparably bad scores don't feel so out of whack. One suggestion would be that the negative weighing can only ever be twice the positive. For example, a perfect score would be 5, and a score failing one metric would then be 3 (the negative weighting being 2 * -1)

I propose true/false metrics, where the presence and number of failed metrics can be used to indicate overall quality. That way, we're not giving anyone negative scores, just indicating the presence or absence of quality indicators -- and if a certain number of quality indicators are missing, we'd start to show yellow, then if more than, e.g. 3, are missing, show red.

This would allow cookbooks to 'turn yellow' without having them go right to red / 'stop' from green, when they fail a single metric. I also don't think any numeric score is going to be operationally useful (e.g. "Oh, this redis cookbook is 83% and this one is 74%, I should use the higher scoring one") -- I feel that coming out with a yes/maybe/no is a stronger statement than a numeric score, but I would treat the yes/no/maybe as passed quality metrics/failing a small number/failing many quality metrics, and I don't think that's necessarily harsh to apply to community cookbooks like a "Don't use this" label or large negative score would be.

jonlives:
The main issue I see there though is that it doesn't allow you to weight metrics more heavily than others. For example, say we introduce a metric for whether or not the cookbook adheres to rubocop guidelines - we'd score a cookbook which failed that metric equally with one which had no maintainers, or no name.

If we used the "3 fails is a red" yardstick, for example, we could have a cookbook failing 3 minor lint checks ranked as a Red despite it passing every other metric. I think there will inevitably be some metrics we consider more indicative of cookbook quality than others, and we're going to need a way to indicate that.

thommay:
I wonder if using an aggregate score to set up a traffic light rating would be a good compromise? IE, score as proposed, but rather than displaying the score directly, have a set of thresholds that define red/amber/green.

jonlives:
That's kind of what I was aiming for with my food label analogy - the scores would make up the red / amber / green mark but the full results table would be available for those who want to dive in.

thommay:
@jonlives sorry, missed that. Yes, indeed. :)

martinb3:
I think we're going to get into the weeds if we start ranking some metrics differently than others. There's bound to be a _lot_ of false positives and differences of opinion. When I look towards the drama/debate with rubocop & foodcritic rules, as well things like rubocop having a release with a broken rule, I'm not sure I would really want to weight any particular cookbook quality metric that much higher than another. I think the really bad ones will sink to the bottom anyway, as they'll tend to fail multiple metrics.

For example, rubocop [released a broken version yesterday](https://github.com/bbatsov/rubocop/issues/2218). Do we want all cookbooks to show red in that case until someone pins back rubocop?

This is just an example, but my guess is that there's going to be a ton of debate around the ruleset causing perfectly fine cookbooks that, e.g. use the wrong quotes, to get marked down.

martinb3:
:+1: for the food label analogy, regardless :)

jonlives:
I dunno, my gut tells me we're going to want to weight some rules (does this have a maintainer, does it have a name) over "does this have double quoted non interpolated strings". Maybe we could have a catchall style category which is weighted at "1" just like the other major metrics. 

Really I mainly want to avoid style bikeshedding crushing cookbook scores while still recognising that people want to know cookbooks follow established practices.

martinb3:
> Maybe we could have a catchall style category which is weighted at "1" just like the other major metrics. 

:+1: That works for me.

I'm excited about the possibilities this will open up, just so I'm clear -- overall this is a huge win :)

jonlives:
Absolutely, no argument from me there :D

someara:
Is there an example of prior art where something like this has proven useful?
Another F/OSS community?

martinisoft:
I like the standards proposal by @nathenharvey and I'm in favor of a more basic initial indicator system to show what is present in the cookbook to indicate completeness and potential quality versus jumping directly at metrics. When inspecting cookbooks for use, I typically ask the following questions:

* Is there a README and CHANGELOG?
* Is the metadata complete?
* Does it work with my platform? What about my version of Chef?
* Are there tests? If so, do they run and/or pass?
* How often are releases happening?

nathenharvey:
@someara I know of http://cpants.cpanauthors.org/kwalitee but can't speak to its usefulness.

Foodcritic is similar in that it has:
* rules that are debated and agreed by the community
* runs on the supermarket displaying pass/fail and rules that have failed

nathenharvey:
While we work through approving this, work will continue on the [chef-cookbooks/cookbook-quality-metrics repo](https://github.com/chef-cookbooks/cookbook-quality-metrics/).  I encourage you to propose metrics, open issues, and nominate yourself as a maintainer and editor of that repo.

miketheman:
Something I've seen used before is the concept of "I use this" on a package. Thereby driving a list & count of "People who use this <thing> in production".
This is a decent measure of quality, as it both has a notion of "look! @miketheman uses X in production,  and I've spoken to him in a hallway at a conference for 2 minutes, that's definitely the right thing to use." as well as a simple metric for use.

Not rating quality, not rating "passes a set of preconceived tests", not "has X downloads, must be good" - more "I use this, for reals, yo.".

martinb3:
> the concept of "I use this" on a package

hey @miketheman! How do other projects keep this up to date and accurate? I worry it would get stale quickly.

miketheman:
@martinb3 It's a personal responsibility, really. If you're no longer using "cookbook awesomesauce" in prod, then it's on you to show that. I consider it more of a social driver - "i vouch that I've used this in prod". Is the cookbook good? Maybe. Is it bad? Also maybe. But I use it.
If I stop using it, I wouldn't want someone to ask me, "Hey Mike, I used the awseomsauce cookbook, and it failed - what did you do that made it work?" and then tell them they have happened on bad recommendations that I stood behind.

I can conceive of a periodic email (quarterly?) - asking a user to update their usage.

All of this is in the name of people wanting to contribute to the community - so the fact that there is a pre-exisiting desire to improve the landscape might be more motivation.

You could also feasible age-out the "I use this" over time if not re-confirmed, as code changes. Might be interesting to tie the "I use this" to a particular version release. This idea comes from where Wordpress plugins are voted on by the community:
screen shot 2015-09-11 at 12 09 24

We all can game a test system (I bet I could write absolutely terrible code that would pass all rubocop tests).
There's a few ways of doing this, and I consider socialization of the feedback vs applying direct metrics is more likely to provide better results.

martinb3:
@miketheman thank you for the thoughtful reply. Your elaboration has gotten my mind running on what a system like this might look like. One particular thing stood out -- if we were using a social endorsement system, we would have to be sure to attach it to the specific version of the cookbook that was endorsed.

MikeTLive:
I find this metric as most valuable. 
"Is this really used" in situations similar to mine.
Even a book with poor structure and zero tests might be best as shown by actual usage vs downloads or even food critic results.

MikeTLive:
This really hits it. 
Affirming usage of a version and aging out that flag over time. - a histogram may be better so that a book which is solid and needs little change would not appear "worse" than a new book. 
If there were a way to track in supermarket convergence data. Say a chef-server plugin that could post back to supermarket that a cookbook is used. 
Would require using a sha vs name-version.

nathenharvey:
I love the idea of having an "I use this" style metric.  I'm concerned about making sure it receives more participation than the rating system we used in the past.  

However, I think this conversation is best had over in the [cookbook-quality-metrics](https://github.com/chef-cookbooks/cookbook-quality-metrics/) repo unless you're suggesting that we should *not* implement some sort of system.

thommay:
this rfc is accepted! @chef/rfc-editors take it away, please :)

miketheman:
> ... unless you're suggesting that we should not implement some sort of system. 
@nathenharvey I think "some system" is a good idea, but applying a "metrics" term to it implies that there's some degree of arbitrary measurement. That's what I'm against.

nathenharvey:
@miketheman would "cookbook qualities" sit better with you?

The qualities of this cookbook include:
* It has a name
* It has a version that is not 0.0.0
* @miketheman says he uses this cookbook
* ...

I'm still a bit unclear if it's the word "metrics", if it's the fact that qualities that can be automatically measured are of little value, or some combination of the two.

miketheman:
@nathenharvey I think that if I were to pop off and build a supermarket plugin (let's assume a world where this exists), the only metric I'd be prone to display is "how many people actually say they use this cookbook".
Maybe I'd add some social connector - "how many people do I know that use this cookbook".
I'f I'm feeling frisky, "how many external cookbook dependencies does this one have".

Then this becomes a reference that anyone can use, and it's not a scoring system for me to game, it's not a download count that I can run up with curl, it's very simple and approachable - "Mike uses this cookbook in production. So do X other people."

It has nothing to do with whether the cookbook has a name or a version string of 0.1 or not - it's whether it's used or not.

(gonna get a bit preachy here for a minute, please forgive me)
Attempting to get 100% code coverage with testing, 100% compliance with a style guide, amongst some measures of quality, make little sense if the code is not useful, unwanted.
Just because I spent the time attaining coverage doesn't mean I've improved anything.

Should these be called qualities? I don't know. Providing a checklist to match against is a good idea (which is why we have linters and foodcritics etc), but by no means is it any measure of quality nor usefulness.

nathenharvey:
The system we launch with will not be perfect and will be subject to some gaming.  'quality' and 'usefulness' are in the eye of the beholder.  As we've seen from previous experiments, asking people to rate cookbooks yields little data and was not helpful.

We will not ever be able to replace a human needing to read a cookbook to determine if it would be helpful or useful to meet that human's needs.

Anything that is measured as part of this system will be exposed.

* Does the cookbook with fewer pull requests open suit my needs better than the one with more?
* Does the cookbook Mike uses suit my needs better that the one he's not using?
* Does the cookbook with a maintainer's email address suit my needs better?
* Does the cookbook with an open source license suit my needs better?

I do not know the answer to any of these questions but do think that they all may predict my happiness with using that cookbook.  Some might be better predictors than other.  I may only care about one or two of them.

A common way to view these qualities and others for each cookbook on the supermarket may make it easier for me to *find* and *write* better cookbooks.

The only way we'll know is to gather and analyze the data.  To do that, we need to put a system in place.

coderanger:
This is definitely the kind of thing we "Math the shit out of". Follow the process this RFC uses to define a whole bunch of metrics, ask a bunch of community members to subjectively rank a bunch of cookbooks, do Math™, profit?

nathenharvey:
@coderanger yep. that's definitely the spirit of the RFC.

nathenharvey:
@chef/rfc-editors although the conversation continued, I believe this is ready for merge.

PR #148, opened 2015-09-14
requesting comments about namspace governance

someara:


coderanger:
Another topic to throw on this pile is prefix delegation. I've started putting `poise-` in front of all my cookbook names (or `application_` for that family for backwards compat) and it would be nice to formalize that so it can be a strong indicator that anything starting with `poise-` is either from me or authorized by me. Could help with the "something in-between" to allow non-Chef-Software-managed "distributions".

new/supermarket_namespace_governance.md
```diff
@@ -0,0 +1,167 @@
...
+have quazi-official repositories where packages outside the "core" set
+are maintained, and usually still follow the packaging policy for the
+distribution. The RHEL core/EPEL/SCL systems are examples of this, as
+are the Ubuntu universe / PPA systems.
+
+Governing the Supermarket like a Linux distribution might ultimately
+provide the better experience for consumers, but it requires, by an
+order of magnitude, more effort and thought. Interestingly, the
+current flat namespace and dependency tooling fit this model well.
+
```
bitmonk: It could be really valuable to have something like PPAs so that there doesn't absolutely have to be consensus on how particular problems are solved, but much like a Linux distribution, there can be a default solution that works for most, and easily surfaced alternate implementations.  I think this would also address @coderanger's comment in the main discussion, so that if someone wants to build an opinionated or flavored namespace which depend on each other to provide functionality otherwise typically provided by a 'core' cookbook, authorship and control are as easily established as for a github repository or Launchpad PPA.  The PPA model is also notable because often packages start there and are promoted to the main distribution as additions or replacement for deprecated packages.

reset:
@someara first off, thank you for taking the time to get this ball rolling by putting these ideas to disk. You offer a couple of very valid solutions to our problem and clearly outline the frustrations of a supermarket user.

I'm concerned with classifying Chef, or it's cookbooks, as a distribution. I believe that we should move away from OS distributions in the same way that many of us have moved away from Big Cable Company packages for delivering our entertainment. It's not that a handful of the channels aren't awesome, it's just that you're paying quite a lot for the slice of entertainment that you actually would be using and you are locked into the choices that somebody else has made for you. We can see this in the way Ruby was packaged in Red Hat. Nobody who **uses Ruby** would package Rake/RubyGems or any other part of the core library into separate packages to be installed - only somebody who was removed from the ecosystem would have packaged things in this way.

I think the modern approach here is to treat Chef as a language with it's own artifacts (big surprise, huh?) and built the proper curation process/tools to set people up for success. We can hash out what these details look like further if this is of interest to people.

new/supermarket_namespace_governance.md
```diff
@@ -0,0 +1,167 @@
...
+   
+Namespace governance is typically optimized for the preservation of
+backwards compatibility of individual components consumed by build
+systems.
+    
+When two libraries are concerned with the same topic, but work in
+different ways, they play semantic games to distinguish themselves.
+
+Some languages, by convention, use a special character for this
+purpose. 'alice/json' vs 'bob/json' is a popular method. Enabling this
```
acaiafa: I like this approach tbh. If we had the ability to have something similar to github where it was supermarket.chef.io/namespace/cookbookname. It would give everyone the ability to offer up their version and whoever wanted to use it could. Companies could also have the supermarket.chef.io/company and in their Berksfile to use only those cookbooks. I see it as a very similar to how go does it https://golang.org/pkg/#stdlib. The core libs are offered and they offer a curated list of external libs that you can choose from
I also feel like if it's going to be on the supermarket it needs to pass some basic linting a spec rules that are globally agreed on. Without some sort of basic testing things can quickly fall apart and we will be back at square one just with way more junk to clean up.

coderanger: So the problem here becomes how to handle dependencies. If deps are against concrete, namespaced cookbooks then you get the distribution problem that @reset mentioned, you are beholden to take 100% of the opinions of a distro when you use any piece of it. The other solution is to have some kind of symbolic-y dependencies but I suspect our APIs are all under way too much flux to pull that off.

acaiafa: Sure but isn't that any language or OS? You always are beholden to 100% of the opinions made by another person or group of people. That's essentially the foundation of everything we have today. Now if we plan on changing that with chef it would take a bunch of effort from all of us. However it's going to be a bit difficult to get consensus from an entire community of people.

acaiafa: Let me give an example. Lets use the nginx cookbook. The current state of that cookbook is obviously not good at all. We couldn't just instantly replace /nginx with something new because it's going to break a bunch of people's infrastructures.

However if we had the ability to do namespace/nginx those people could be told the top level /nginx that is stamped of approval by ops code is being changed and if they want to use the old one they can point to the namespace. That would give us a clean way to clear out cookbooks that have been abandoned/ poorly written / left behind. 

This would also allow us the community collectively vote and agree on what namespace cookbook gets linked to  /cookbook on supermarket and that will get the stamp of approval by everyone. Now of course not everyone will agree the same way and that's why we will have the namespace/cookbook for them to use as they please.

coderanger: So there are a few options to get around this. The nicest is how NPM does dependencies where every app gets its own deps managed independently. This means You can pull in a dependency without worrying (as much) about the recursive stuffs. This is unlikely to work well for us as the host system being configured acts as a huge global, mutable state.

The more common approach in packaging systems is what we have with Berks (and Bundler like it) where you always depend on concrete dependencies but can use a `:git` source to hack it so that it doesn't use the canonical source for that cookbook.

An extension of this is what DevPI does for Python with layered repositories. Each repository can optionally have a parent link, meaning that when searching for which package (cookbook) to install it checks the specific repo you requested, then up the parent chain. This allows for things like a totally isolated distribution, a distro with one or two cookbooks overlaying (replacing) matching names from the top-level, global repo, or more complex scenarios. The downside is it has to be mediated by a mildly complex webapp, though putting that logic in Supermarket seems like it could be a good fit.

martinb3: I like the idea of a namespace or package-like String that goes before a cookbook name.

> people could be told the top level /nginx that is stamped of approval by ops code is being changed and if they want to use the old one they can point to the namespace

I would strongly suggest we don't have a missing namespace mean that a cookbook's source could change. I would prefer we make namespaces required and assign some indicator value like 'none' to the existing cookbook names. All new cookbooks could use `none/foo` or `martinb3/foo`, but I don't think we should ever change `none/foo` to point to a different cookbook than `foo` does today. This would be nice and backwards compatible because we could always assume in any older code that a missing delimiter means `none/` for the namespace.

new/supermarket_namespace_governance.md
```diff
@@ -0,0 +1,167 @@
...
+This model is often used by sophisticated Chef shops, who often write
+everything themselves and rarely consume cookbooks from the public
+Supermarket.
+   
+In Linux distributions, two packages are rarely, if ever, concerned
+with the same software. Rather than having 'alice/apache' vs
+'bob/apache', the distribution will pick a preferred package and
+declare if "the way of the system". Often these packages vary wildly
+between distributions. The configuration conventions around the Apache
+httpd server is a great example of this.
```
martinb3: I feel like I've seen a lot of counter examples to this, e.g. supporting multiple PHP versions that have been namespaced using package names, as well as sun-java/openjdk and various latex distributions. I'm not disagreeing with your sentiment, but we're likely to have similar outcomes if we declare a 'one true way' model. We've been rather lucky that, for example, the java cookbook has supported both flavors (sun/openjdk). I'm not sure every cookbook would end up with that much unity.

new/supermarket_namespace_governance.md
martinb3: I feel like the model proposed this section will quickly become a 'one true way' debate, and we'll basically be saying there's only one way to do things. This would go against previous sentiments about having everyone do it the way that fits them best, e.g. the berkshelf way, or the RFC regarding 'supported' workflows.

new/supermarket_namespace_governance.md
lamont-granquist: The current thinking now is that policyfiles and multiple sources may address this better.  This will let sites either check in their own version of the build-essential cookbook into their git chef-repo, or run their own internal supermarket with their own version of build-essential and let them pull from supermarket and resolve that conflict, selecting the internal one.  That would also allow someone to setup a public supermarket site that was a collection of cookbooks that overrode public supermarket ones to satisfy some purpose -- like a set of cookbooks that had the same API as the core public ones, but targeted only binary installs on RHEL or something.

Lots of people fairly violently hate the idea of simple namespacing because it will pollute the public supermarket even more with useless unmaintained forks of common cookbooks.

I suspect there will be lots of discussion on this topic at the community summit.

FWIW, I've been a supporter of namespacing.  Although its probably going to be less intrusive to support another parameter in the metadata that implements the namespace rather than adding it to the name.

new/supermarket_namespace_governance.md
bitmonk: I think avoiding that is @someara's goal in engaging the community with this RFC, and that the full text adequately expounds upon that risk.

As a consumer of cookbooks, when I need to install, say, Java and Maven, I want straightforward cookbooks within my reach which are fairly cross-platform, reasonably tested, and maintained by someone making an effort not to break my usage in the future.

As a contributor to cookbooks whose goals sometimes are or are not shared by what one might call 'mainline' cookbook maintainers, perhaps toward ensuring the above, I may want an easy way to published a forked or completely alternative cookbook without pointing at a github repository.  I'm just saying that I think the Ubuntu PPA model works well for that, and I think might address the sort of namespacing that e.g. @coderanger wants to see with 'poise' cookbooks'

There doesn't necessarily have to be 'one true way', but there can be a default way and facilitation of alternatives.

new/supermarket_namespace_governance.md
acaiafa: @martinb3 so yeah i disagree with you on this part. I think that there should be a primary namespace for the cookbook lets use nginx for example again. supermarket.chef.io/nginx should be the primary cookbook that points to someone elses namespace. This cookbook can be selected/voted/crowdsourced by the community that this is the best choice and everyone should be able to use it. If this person decides to stop their project and not continue to work on it a new cookbook will be chosen.

coderanger: That assumes 1) there is a new cookbook to be chosen and 2) 99% of "crowdsourced" work won't be a hot mess. Would a global namespace be useful if there was almost nothing in it because we restricted it to cookbooks that hit very stringent reusability guidelines?

martinb3:
Whatever solution is chosen, I would like it to behave in such a way that the current flat namespace is deprecated, and we don't use it as a way to steer people to a particular namespace depending on $cookbook and $current_timestamp. In other words, $cookbook would always default to $default_deprecated_namespace/$cookbook, and we wouldn't use default namespaces to steer people to particular blessed namespaces or cookbooks.

TL;DR I would like a single inferred namespace for all cookbooks, and I want it to stay constant.

nathenharvey:
@martinb3 are you saying that you would like to see **all** cookbooks currently published on the Supermarket deprecated?  Are you also saying that publishing new cookbooks requires you to go through some sort of vetting process?

nathenharvey:
I think this policy should explicitly state:

* The public Supermarket (supermarket.chef.io) will have one flat namespace.
* The Supermarket application (http://github.com/chef/supermarket) will only support one flat namespace.
* Anyone with a login to the public Supermarket may share any cookbook they'd like with any name that is available at the time.
* Anyone who has published a cookbook to the public Supermarket may place that cookbook up for adoption.
* Anyone with a login to the public Supermarket may request to adopt a cookbook that is up for adoption.

The governance policy should cover:
* Arbitration process for transfer of ownership of a cookbook when both parties do not agree to the transfer or where one party does not participate in the conversation.
* Guidance on what to do when transferring ownership.  This may include things like a post to the mailing list, bumping the major version of the published cookbook, etc.

Outcomes this process:
* Chef is a welcoming and inclusive community that celebrates the contributions of all members.
* It is possible to obtain ownership of a seemingly abandoned cookbook.

acaiafa:
@nathanharvey so wait you are saying that there will never be supermarket.chef.io/namespace/cookbookname?

coderanger:
@acaiafa This is (I think) Nathen in his "community member having opinions" hat, not "decider" hat :)

acaiafa:
@coderanger ah ok because asking politely for people to give up work they have done so you can take it over doesn't work so well.

nathenharvey:
@acaiafa yes, that is what I am saying.  And, as @coderanger correctly stated, I'm saying so as a member of the Chef community.  I'm not the 'decider'.

I do think it's important that we make explicit our intention, as a community, whether or not "supermarket.chef.io/namespace/cookbookname" is a viable path forward.  I do not think it is.

I think we should focus on better management of the namespace we do have.  As an example, there is no arbitration process in place that you can use when you'd like to take over a cookbook that exists.  There should be one but I don't fully understand what that would look like.  I'd like to see that process explored in this RFC.

nathenharvey:
From the Chef Developers' IRC meeting earlier today:

12:24 <@someara> the purpose of the RFC was to try and establish a scope that a governance policy would operate within
12:25 <@someara> define the goals, etc
12:26 <@holoway> another option is something akin to a 'derivation'
12:26 <@someara> from there we can decide if there is any engineering work to do, or if this can be %100 process
12:26 <@jtimberman> i've seen namespacing mentioned in this within context, but do we have a clear statement about whether or not namespacing on supermarket is a thing that could happen? is this to address "someone has the namespace for something but aren't responsive to requests to fix/make changes/etc on a broken cookbook"
12:26 <@holoway> where a cookbook is considered "the same thing" if it has the same name and version number
12:26 <@holoway> but you can have preferences for derivation
12:26 <@holoway> ie: I prefer "someara" cookbooks if they exist, otherwise use "chef"
12:27 <@holoway> it'll still be completely up to the author to have it, well, actually work :)
12:27 <@holoway> I'm not sure this makes things better, but it's another way to deal with it
12:27 <@holoway> (essentially a list of derivations, in preference order0
12:27 <@coderanger> holoway: That is similar to the DevPI model
12:27 < lamont_oc> yeah, i've been a proponent of namespacing, but everyone that i've talked to internally about namespacing makes it pretty clear to me that namespacing is dead
12:28  * holoway nods
12:28 <@someara> well we DO have a namespace, its just a flat one
12:28 <@someara> so are we going to just govern that,  or explore making it not flat
12:28 <@coderanger> Multiple "repositories" with some kind of tiered inheritance scheme seems like the least disruptive way forward
12:28 <@coderanger> Possibly with prefix reservations on the global namespace
12:28 <@holoway> coderanger: is that essentially a +1 on what I just called a "derivation"?
12:29 <@coderanger> holoway: Yeah, same idea.
12:29 < lamont_oc> yeah multiple repos with conflict resolution is what is being done in policyfiles now
12:29 <@holoway> I'm pretty sure that could be implemented in a back-compat way
12:29 <@holoway> ie: everything without a derivation just gets some catch-all default derivation
12:30 <@coderanger> holoway: Or phrased another way, by default all new repos on Supermarket inherit from "root"
12:30 < nathenharvey> someara:  I'll say whatever I need to get everyone to stop spending time thinking about having something other than a flat namespace on the supermarket.  I welcome the discussion about how to manage/govern that namespace better.
12:30 <@holoway> right
12:30 < lamont_oc> https://github.com/chef/chef-dk/issues/430 and https://github.com/chef/chef-dk/pull/450 are minimum viable product
12:31 < lamont_oc> already in chef-dk and policyfiles
12:31 < kallistec> just to be clear, what exists in policyfiles now will conflict if the names are the same, doesn't care about versions
12:31 < kallistec> I could perhaps be convinced otherwise, but am taking the most conservative path to start
12:31 <@holoway> right - but we could easily move it forward
12:32 <@coderanger> The downside of the repo/fallback model is it really needs a webapp to mediate the lookups
12:32 < kallistec> the risk is that you have two cookbooks for the same thing (mysql or whatever) but they don't have the same API
12:32 <@holoway> right
12:32 < lamont_oc> yeah if "build-essential" comes from more than one source (has versions in two sources, even if the versions aren't the same) then you have to override that with an explicit cookbook source line to resolve the conflict
12:32 <@coderanger> So this would have to be coded into Supermarket most likely.
12:32 <@holoway> that's totally going to happen
12:32 <@holoway> but it's going to be on you
12:32 <@holoway> (where you is the user)
12:32 <@holoway> there really isn't another way
12:32 <@someara> the other thing is, prefixes are effectively namespaces
12:32 <@holoway> right
12:33 <@holoway> if you can reserve a prefix..
12:33 <@someara> that use a - instead of a /
12:33 <@holoway> a spade is a spade
12:33 <@coderanger> someara: No, they aren't, unless you can globally say "use my thing instead of the global thing"
12:33 <@holoway> right
12:33 < ranjibd> right
12:33 <@jtimberman> metadata has a 'provides' keyword.
12:33 <@jtimberman> never implemented, but it exists.
12:34 <@holoway> one thing a model like this probably needs
12:34 <@holoway> is a "release" number for a cookbook, seperate from the version
12:34 <@coderanger> someara: You _could_ implement that feature on top of simple name prefixes, but why?
12:34 <@thom> we're almost into API versioning
12:35 <@holoway> ie: "cookbook default/a1 version 1.2.3 and cookbook coderanger/a1 version 1.2.3" are equivalent, but I prefer coderanger
12:35 <@coderanger> meh, if down this path also lies formalized interfaces
12:35 <@holoway> but if coderanger wants to move that forward and remain equivalent, he needs a number he can change that doesn't fuck with the equivilancy acros
12:35 < lamont_oc> right now we have the ability to check a build-essential cookbook into your local git repo or local supermarket instance and override the public supermarket one while still pulling other cookbooks from the public supermarket with dan's patches to chef-dk
12:35 <@coderanger> And I don't think we know what we are doing enough to commit to that kind of thing :)
12:35 <@holoway> ie: "a1 v 1.2.3 r xyz"
12:36 <@holoway> I'm not sure this is good for chef, but it's what you would need to do, I think
12:36 <@coderanger> lamont_oc: Yeah, but its a pain in the ass
12:36 <@coderanger> lamont_oc: Idea is to make that one click on Supermarket instead of a bunch of "download here, upload there, reconfigure everywhere"
12:37 <@coderanger> This is common enough that it also probably shouldn't require a private supermarket (though it should certainly be an option)
12:39 < nathenharvey> we have about 12 minutes left and this RFC is meant to gather feedback.  I suggest we continue the discussion in the PR for the RFC and use the rest of our time to discuss foodcritic release
12:39 < lamont_oc> if the server supported /universe it could also be used as a source (although there's probably a bit of inception in CI pipelines there)
12:39 < martinb3> OT, but my feedback: Whatever solution is chosen, I would like it to behave in such a way that the current flat namespace is deprecated, and we don't use it as a way to steer people to a particular namespace depending on $cookbook and $current_timestamp. In other words, $cookbook would always default to $default_deprecated_namespace/$cookbook, and we wouldn't use default namespaces to steer people to particular blessed namespaces or cookbooks.
12:39 <@coderanger> holoway: In the long term I agree, in the short term I think that is a hole we don't want to go down yet
12:39 <@holoway> y
12:40 <@holoway> lets move on
12:40 <@holoway> and take it to the PR

martinb3:
> @martinb3 are you saying that you would like to see all cookbooks currently published on the Supermarket deprecated?

If we adopt the idea of namespaces for cookbooks, I would like to deprecate references to cookbooks without a namespace. I wouldn't deprecate the cookbooks -- just the syntax that omits a namespace.

> Are you also saying that publishing new cookbooks requires you to go through some sort of vetting process?

No, I think there was negative missed -- I explicitly don't want any blessed namespaces. I think we should defer to the cookbook quality RFCs/metrics for that.

My thoughts on moving forward:

- we should use a delimiter like `/` to indicate namespaces, e.g. `martinb3/redisio` or `martinb3/nginx`. I don't have a strong opinion about whether the left-hand side is attached to a user, organization, or other arbitrary string

- we should create a default namespace, but only for backwards compatibility. we should emit a warning, for example, if a chef-client capable of namespacing has to infer the default namespace because the delimiter was omitted

- the default namespace should not change over time. i.e. it's used for backwards-compat only, not for indicating which $database or $webserver cookbook is the current community favorite.

stevendanna:
Since this RFC was posed as more of a request for comments.  I'd like to discuss the assumption that the *place to start* is some rigid formalism.  

> It is clear that we need some sort of governance model to help things run more smoothly. The scope is the hard part. There are many examples of abandoned cookbooks, namespace squatting, and low quality cookbooks occupying important names.

This, is not clear to me.  Here are a few of my experiences:

- I wrote a xinetd cookbook without checking the community site first because I love superservers/process management.  It was, imho, much more complete the the existing xinetd cookbook.  I emailed the maintainer; he agreed that my cookbook should be the default, and we became co-maintainers of xinetd.

- By an accident of history [0], I ended up with the systemd cookbook attached to my account.  This was very bad because (1) I didn't have big plans/time to do anything with it, and (2) I couldn't easily get rid of it.  During a cookbook maintenance discussion on the mailing list, someone point out this abandoned cookbook, I followed up with a quick message saying I'd be happy to give it away.  @nathwill, @ranjib and a few others immediately emailed me with great ideas.  That cookbook is now in stellar shape because of @nathwill and a couple of emails.

- I maintain the r-project cookbook.  It is getting a bit old now (I need to update it), but for a while it was clearly superior than the R cookbook on supermarket.  I've gotten emails and bug reports from a number of users of my r-project cookbook and none of them seem to have had a problem comparing the two and deciding to use mine (I imagine many people just write their own cookbook for R, it is pretty simple).

I'm certainly not against a process, I just wonder how many cases we actually have of maintainers not cooperating with efforts to improve their cookbooks?  

It seems to be the biggest problem might just be maintainers who have disappeared completely.  What if supermarket had a feature where ownership of a cookbook reverted to some defined user after some expiration period.  That is, if a maintainer falls of the face of the planet some community control account eventually gets ownership of that cookbook.  A group of volunteers could run that account, watch for new cookbooks that got reparented, find people who might be interested in adopting them, etc.

stevendanna:
That was a lot of words, but FWIW, I'm also fine with namespaces :D

lamont-granquist:
What I'd actually like to see are more cookbooks moved back to the chef-cookbooks organization now that we have an internal cookbooks team.  We could then add external teams to the commiters on those and get actual cooperation in maintaining a cookbook (an actual 'community' cookbook) rather than isolated islands of owners being single-points-of-failure for all the cookbooks.  Would also like to see chef_brigade and other teams added as committers to cookbooks they would like to maintain in chef-cookbooks.

As an example, the 'locale' cookbook is pretty decent, but there's a PR that has been open for a year to add RHEL7 support that hasn't been merged:

https://github.com/hw-cookbooks/locale/pull/14

I'd like to be able to whack merge on that PR, and like to have a lot more people than just HW able to whack merge on that PR since it seems to be far enough down their priority list that its not getting a lot of attention.

Kind of similarly I'm not sure its wise to have a cookbook like 'runit' that is close the center of the cookbooks universe managed by only one team.

And that's not to be overly critical of HW as cookbook maintainers, but they own a lot of them so its easy to find examples...

stevendanna:
@lamont-granquist  I would happily give the cookbooks I took over back and contribute to them via a larger community.  They are the type of projects where commits often come in infrequently enough that taking simple PRs means catching up on all the test ecosystem changes, etc.

lamont-granquist:
Yeah that's where its useful to have @tas50 who will just do megapasses through everything updating all the cookbooks for consistency and compliance.

someara:
Closing due to in-actionability. Will reference in a future "here's how to fix it" RFC.
Making a note that this the chef-dev list was probably a more appropriate place to have this discussion.

thommay:
Thanks to everyone who contributed to this RFC.

PR #150, opened 2015-10-14
Add subscribes block RFC

jkeiser:


nathwill:
would this permit a syntax that enabled a resource to subscribe to all resources of a particular type?

something like:

```ruby
service 'my-service' do
  action [:enable, :start]
  subscribes :reload, do
    run_context.resource_collection.select { |r| r.is_a?(Chef::Resource::MyResource }.any? do |r|
      r.updated_by_last_action?
    end
  end
end
```

it looks kind of close, but maybe not?

lamont-granquist:
so for accumulators, you'd generally use delayed notification with an `action :nothing` on the resource. so that later cookbooks could also send the resource information.  and what we're doing is accumulating blocks in a list that we apply to the resource before triggering it?

this also doesn't kill off the use of definitions for doing accumulator patterns, since generally when i'm writing one of those it is desirable for later cookbooks to declare state on the resource, but i want the resource to run early in the run to setup state for resources in later cookbooks, so i'm leveraging the fact that definitions are compile-time macros (i.e. if i'm setting up firewall rules, i want those accumulated across all cookbooks, but applied very early in the converge).

miketheman:
I think we discussed this briefly at summit, but it is still unclear what the current problem is in real-world terms.
Bringing something into desired state by the end of a Chef run is what Chef is supposed to do, and at the successful end, things should be restarted/started/put into place as needed.
Can you try to explain that please?

new/subscribes-block.md
```diff
@@ -0,0 +1,73 @@
...
+
+```ruby
+file '/etc/config-file.txt' do
+  content 'config!'
+end
+file '/etc/init.d/x/start.sh' do
+end
+# The service will only be processed once; if the configuration is updated, we
+# upgrade the action to :reload; if the configuration is started, we upgrade
+# the action to :restart.
```
btm: What decides that reload gets upgraded to restart? 

Is it the order of the subscribes entries in the resource below? Seems magic. You could end up with something similar to "before" in #154 if that was true:
```
service 'x' do
  subscribes('execute[do we need to make a change]', :immediately) { action :stop }
  subscribes('execute[make the change]', :immediately) { action :start }
do
```

Or are we putting that in the code? If that, can we discuss and possibly document that matrix of possibilities?

coderanger:
I am overall unconvinced that this is a common enough problem to warrant new syntax. I don't think it allows anything that notifying a ruby_block resource doesn't already allow.

someara:
Mike, the general problem is long-standing and looks like this:

https://gist.github.com/someara/f1506e9257da611eaabf

^ In that gist, a a template notifies a service to restart when repaired.

There is no way to prevent the unnecessary restart at the end of the
chef-client run while retaining the benefits of the notify/subscribes on
subsequent runs.

The system is at the desired state by the end of it all, but the situation
is Super Freaking Annoying.

If the notifies/subscribed resource takes a long time...
- 'service[enterprise java thing]',

or makes a lot of noise in logs/servicediscovery -
- 'doker_container[noisy initializer]'

of costs real money -
- 'expensive_aws_instance['billed_for_an_hour']'

... this can actually be a problem.

That being said, I'm unable to grok exactly how the subscribes block would
help here...

-s

btm:
@miketheman Looking at existing cookbooks that provide accumulation may help understand the general problem: https://github.com/lamont-cookbooks/multipackage.

Consider a situation where you have a resource that adds something to a file, like a mail alias or a cron job. You may want to accumulate all of those changes and do the action once rather than have the same resource converge many times.

@someara I think in your example you could check to see if the service is currently started in the subscribes block, and choose the appropriate action for the notification, i.e. nothing if not running or restart if it is.

coderanger:
@btm With a delayed notification it would always be running either way, the question is "did it start after the file was changed?" or "was this the first run when it was created?", neither of which are easy to write code to identify even with this new syntax.

someara:
@btm Not exactly... . This has been explored ad-nauseum (I can dig up a few
threads).

The service would already be running by the time it gets there.
The real logic would look like "subscribe :restart, but only if not updated
by :start"

-s

btm:
I think enough people think this needs real-world examples added to the RFC that this would solve to continue, so let's dig those up and give them to @jkeiser.

someara:
1) https://gist.github.com/someara/f1506e9257da611eaabf

2) https://gist.github.com/someara/8b38dcdc9a5040b3d101

3) https://gist.github.com/someara/b2b42d734c447f19157a

4) https://github.com/someara/chef-docker/issues/490

coderanger:
In all of those cases, this doesn't seem to provide much benefit over notifying a ruby_block, the hard problem of "should we ignore this notification" is still present and this proposal doesn't actually help with that.

jkeiser:
I agree we should table this until we have better real world examples.

lamont-granquist:
notifying_block in chef/chef should be able to deal with this:

https://github.com/chef/chef/pull/4818

```ruby
notifying_block do
  file "/etc/thing/conf1" do
    notifies :restart, 'service[thing]', :delayed
  end
  file "/etc/thing/conf2" do
    notifies :restart, 'service[thing]', :delayed
  end
  service "thing" do
   action :start
  end
end
```

- if neither file resource triggers you will get one :start service call at the end of the block which will be fully idempotent
- if either or both of the file resources triggers you will get a :restart that runs delayed to the end of the notifying block (not to the end of the chef-client-run or the resource).
- it will not trigger a restart with partially managed state (like using :immediately would)
- it will de-dup the :restart and only run it once
- you will get a :start followed by a :restart but proper idempotency in the service resource should avoid that being an issue

i think it'll close this issue.

as of 12.9.x we will also support notifications bubbling up to wrapping run_contexts as well, and delayed notifications will fire in the notification context of the targetted resource.  those should give us the tools to work around a lot of these kinds of issues.

lamont-granquist:
^ so i was a bit wrong about that notifying_block runs at compile time and doesn't quite give the correct grouping behavior API unless you wrap it manually in a ruby_block to lazy it.

adding a recipe block resource around a notifying_block would solve that a bit more elegantly tho:  chef/chef#4982

jkeiser:
Should we close this? No word for 4.5 years makes me think we've moved on for now :)

jkeiser:
Calling this closed, but please absolutely reopen if you see value in having it open!

PR #154, opened 2015-10-15
Add RFC for before notification timing

jkeiser:


lamont-granquist:
Another way to do this would be to put an always_why_run flag on to all resources and then if its true its always why-run but then you can chain a notify to a service restart and then to another package upgrade resource.  The problem there is that then the user must duplicate two resources exactly other than the first one does the why-run mode, along with the ceremony of setting up the chain.  That seems more error prone and more ceremonial.

We could also solve this in the DSL with helpers like `package_installable?` and `package_upgradable?` (which i want anyway) that would look something like this:

```
def package_installable?(name)
  !build_resource(:package, name).current_value.candidate_version.nil?
end

def package_upgradable?(name)
  cv = build_resource(:package, name).current_value
  cv.installed_version != cv.candidate_version
end

if package_upgradable?('foo')
  service "foo" do
    action :stop
  end

  package "foo" do
    action :upgrade
  end
end
```

(or with guards)

lamont-granquist:
I'm :+1:  on the idea though

coderanger:
I would need to see really compelling evidence that the why run integration is actually correct enough in existing cookbooks to allow this. I would be worried about this being silently broken in a lot of cases (due to cookbook code, not Chef) and that leading to confusion.

jkeiser:
@coderanger I will add a note here that :immediately_before will raise an error for resources that do not support why run. Cookbooks have to explicitly enable this at present, so it's a reasonable bet that most of those who have disabled it are in good shape.

coderanger:
@jkeiser A lot of cookbooks have resources that claim to support why run, but have never tested this code. I don't even know how you would test it? In light of that, I have to assume most of it is broken and thus this feature would be broken too.

jkeiser:
Is it your intent to bring up a conversation about whether Chef should
have/support the whyrun feature? If so, can you have that conversation in a
separate RFC? I'm happy to wait on this until we've had that conversation,
as long as we decide to have it soon. I'm less happy to be blocked on a
nebulous conversation about whether a major feature of Chef is just
Terrible And Should Not Be Used.

jaymzh:
I'm :+1: on this. I started coding up something similar a while ago without whyrun... if the idempotency check didn't pass I called the notification. It involved a lot of invasive changes to every provider ... while it's nice not to have to depend on whyrun, this may be a somewhat less complicated approach. This feature in some form is sorely needed.

lamont-granquist:
Yeah why-run being broken is broken and should be fixed.  The fact that it might be broken shouldn't block this feature (why-running everything correctly is also a never-ending yak shave because cookbook resources grow unbounded).   The alternative is also to restructure providers completely with test-then-set semantics which is just another why-run-like never-ending yak shave.

coderanger:
I don't think it should block the feature per se, but it should be strongly noted in the docs that it might not work with every community resource out there. My concern is mostly that this feature would be tied internally to the "never-ending yak shave" and so many people that have just totally ignored why run might suddenly be exposed to how gross it is, probably without realizing.

btm:
@zuazo look, the return of [CHEF-2421](https://tickets.opscode.com/browse/CHEF-2421).

thommay:
I'm not sure about this as a feature; it makes cookbooks (I think) harder to reason about - but maybe I'm scarred by the Puppet DAG and having to cover modules in `before` and `after`.

btm:
@thommay I think the key here is you use it very sparingly, only when you have to and doing something without it is pretty complicated (probably you're forced to write it outside of Chef and execute that). My memory of some cases from way back follow. Maybe @knightorc had one in mind too that he discussed with @jkeiser.

As a user with a database that I ship as files, I need to stop the database before I replace those files:
```
remote_file "database.db" do
  source somewhere
  path ondisk
  notifies :stop, "service[mydb]", :before
  notifies :start, "service[mydb]", :immediately
end

service "mydb" do
  action [ :enable, :start ]
end
```

Also the perennial bad apache template crashes apache thing
```
config_file = "/etc/apache2/httpd.conf"

execute "test apache config" do
  command "/usr/local/apache/bin/httpd -t -f #{config_file}"
  action :nothing
end

template "apache config" do
  source "httpd.conf.erb"
  path config_file
  notifies :run, "execute[test apache config]", :before
  notifies :reload, "service[apache2]", :immediately
end

service "apache2" do
  action [ :start, :enable ]
end
```

someara:
My alarms go off when I see something I've never personally had a need
for...

I've always found it more effective to use a single atomic script resource
to do these kind of things, rather than trying to chain together a pile of
executes.

Re, "the idea is that you use the very sparingly" eh.... If the potential
for abuse exists, people will abuse it to its logical conclusion.
Compilation phase decision logic was supposed to be used very sparingly. We
see how that turned out.


-s
​

jaymzh:
This is a thing I'm constantly working around in Chef with fake ruby_block resource that generate a template in another place, then diff them, then notify :stop, then notify the real template, then notify :start... there is a very clear need for this, come on. Or package upgrades instead of templates. Anything that needs something done _before_ you can make a change. Notify a software load balancer. Whatever. Tying this in an opaque script is stupid, then Chef doesn't knwo it's upgrading a package or updating a template or whatever.

coderanger:
To summarize my position from IRC, we should make sure to explore maximum footgun mitigation but not block moving forward.

jaymzh:
@coderanger :+1:

knightorc:
Sorry for the delayed response.  This use case came up on a Windows system. We have a third party product that has a slightly borked msi.  When we need to upgrade the msi to a newer version it fails to stop its service (which is should be doing) prior to upgrading and the chef-client run hangs/blocks due to windows file lock contention.

Obviously getting the msi properly fixed is one solution. We worked around this by manually fix the msi ourselves, however the use case of notifying something to do something before an action is done could be handy.

lamont-granquist:
Still think this should be renamed to `:before` for simplicity since there's no useful delayed concept to use for before, and we should probably look at `:before_once` where the target resource is notified only if its not already been updated by something else.  See the discussion in CHEF-2421 for more information on `:before_once` and the use cases.

It feels like CHEF-2421 got derailed by issues like what to do on first run before the package is installed and you can't stop the service because the files aren't there?  Just dropping a guard in there solves the problem though:

```ruby
service "whatever" do
  action :nothing
  only_if { 'File.exist?("/etc/init.d/whatever") }
end

package "whatever" do
  notifies :stop, 'service[whatever]', :before
  notifies :start, 'service[whatever]', :immediately
end
```

Makes the user responsible for solving the problem, but solving it within Chef itself, without the user having to explicitly guard, seemed to be what derailed CHEF-2421 down the path of trying to have Chef ultimately need to solve the Halting problem...

So, I'm :+1: that we can keep this simple, produce something useful, and avoid the philosophical issues that CHEF-2421 got mired in...

zuazo:
Sorry for the delay. I’m really happy to see this issue again. A huge :+1:

IIRC these were the main reasons that caused the **derailment of CHEF-2124**:

- There was a **pre-whyrun** and a post-whyrun PR. The second, instead of being a new proposal, was a rebase of the first and that caused the commits to be a bit messy and difficult to follow and review. This was an unfortunate coincidence in time.
- The complexity: **Including the :depends** implementation in the same PR was too much.
- There was also some controversy about **how I implemented it**. But I never got to concrete conclusions about what exactly.

We all spent much time on this matter, there were doubts and discussions, spread between GitHub, JIRA and the mailing list, the ticket became complicated to follow and I think we just lose the way. Probably I didn't have the necessary knowledge to handle it properly.

Anyway, that doesn't matter much today. **Focusing on the subject**, some comments:

- I didn't find a good way to implement it until the arrival of whyrun. I don't know if it would be possible if one day it is "removed".
- To support before notifications in resources without whyrun, I used the only_if/not_if properties. If the resource doesn't support whyrun but has an only_if or a not_if condition, use it to simulate whyrun printing a warning about that. It was a little tricky idea and today probably makes less sense when whyrun has been a while among us, but just in case.
- @btm your **apache template example** would not work. The check will run before generating the template, so it will check the old file (if existed). To cover this use case `template#verify` can be used now.

To summarize, the use cases for `:before` notifications:

- To mount /tmp with exec before compiling nginx (the original use case).
- Stopping the database before an upgrade.
- Some services don't restart correctly if you don’t stop them before upgrading.
- Remove instances from a load balancer before upgrading.

Some use cases for `:before_once`/`:depends`:

- apt-get update
- Creating the parent directory of a directory?? OK, I think apt-get update is the most prominent example here.

Well, I think that's enough for now. As you may have noticed, I can spend hours talking about this  :smile:

btm:
@zuazo thanks for your stories of yore.

I'm :+1: on renaming `:immediately_before` to just `:before`. Without a `:before_once` to differentiate it, it might not be clear which use case `:before` actually handles. Either way, I'm :+1: on accepting this with `:before` or both `:before` and `:before_once`.

Clearly we've wanted this for a long time, and I want to be sure the scoping conversation doesn't grow too much and derail us again.

jaymzh:
:+1: on renaming to `:before:` and :+1: in general

lamont-granquist:
Yep :+1: on :before, and :+1: on :before_once:

adamhjk:
I'm the decider and I approve this RFC. Lets make sure we do everything we can to guard against footgunnery. 

Love,
Adam

jkeiser:
Renaming to before in accordance with the will of everybody :) (I'm cool with it.)

thommay:
Merged, thanks!

PR #155, opened 2015-10-15
Fix link to DEP-1 in django/deps.

rhass-r7:
Django seems to have refactored their DEP structure which broke the link in the attributions section. This is just a quick fix to the link.


coderanger:
Thanky!

PR #156, opened 2015-10-15
Add RFC for installing ohai plugins at bootstrap

martinisoft:


new/install-ohai-plugins-at-bootstrap.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Specification
+
+We add a new flag to the knife boostrap subcommand which allows you to specify
+a single file or directory of ohai plugins to be placed in the newly created
+`/etc/client.d` directory upon a bootstrap.
+
+An example command would look like for a single plugin:
+
+`knife bootstrap node.example.com --ohai-plugins="my_custom_plugin.rb" ...`
```
mcquin: Will there be support/is there a use case for a comma separated list? E.g., `--ohai-plugins="my_custom_plugin1.rb,my_custom_plugin2.rb,...,my_custom_pluginN.rb"`

lamont-granquist:
How does this interact with ohai plugins in your cookbooks, though?

I don't really want to copypasta ohai plugin files from cookbooks into my knife bootstrap config under my .chef dir.

After the summit I was actually considering RFC'ing making ohai plugins a top-level segment in cookbooks and then doing ohai loading in two phases where after libraries load then all the cookbook ohai plugins would run.  That would make it easy to find ohai plugins (rather than the way it is now where you can't find them until you compile+coverge recipes and one of the file or template resources happens to drop in a plugin), and it would essentially auto-reload all the plugins without requiring a lot of ceremony, and would get it done early so that the results would be available in attributes, compile phase, etc.

tbunnyman:
@lamont-granquist I think your idea would be even better.

lamont-granquist:
I just cut #159 to discuss that idea.

martinisoft:
Really like #159 so we could close this in favor of that one as well since IMO Ohai plugins should be first class citizens. :heart:

lamont-granquist:
Heh, its your PR @martinisoft, feel free to close it...

new/install-ohai-plugins-at-bootstrap.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Motivation
+
+    As a chef user,
+    I want to install ohai plugins during bootstrap,
+    so that I have my ohai facts on my initial chef run.
+
+## Specification
+
+We add a new flag to the knife boostrap subcommand which allows you to specify
```
miketheman: typo: bootstrap. Unless it's in the Halloween vein? :ghost:

thommay:
Closing this in favour of #159

PR #157, opened 2015-10-19
Chef-Client Exit status

carpnick:
@adamedx @fnichol @mwrock 

At Chef Summit we discussed using Exit codes to determine what was going on inside a Chef run.  Primary use case at the time was for Windows nodes needing to reboot and testing it within Test-Kitchen.

There is an [open issue](https://github.com/chef/chef/issues/3916) that we plan on submitting a PR for that will handle the default --fork use case.  But before making the change, would like to get it discussed here.

If there are more use cases, please list and I will add them to the table.

### DIFF
--- a/exit-status.md
+++ b/exit-status.md
@@ -15,15 +15,16 @@ Signal outside tools of specific Chef-Client run status.  Ability to determine r
     so that Test-Kitchen/Vagrant/any outside tool can wait for node to reboot, and continue converging.
     
 ## Specification
-* Chef-apply, Chef-client, Chef-Solo should honor the below exit chef run exit codes
+* Chef applications (e.g. chef-client) that interpret recipes should use the specified exit codes
+* Chef tools (e.g. knife) should behave appropriately for the exit code, or pass it to the user
 
-### Exit codes reserved across platforms
+### Exit codes Reserved by Operating System
 * Windows- [Link](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681381(v=vs.85).aspx)
 * Linux - [Sysexits](http://www.freebsd.org/cgi/man.cgi?query=sysexits&apropos=0&sektion=0&manpath=FreeBSD+4.3-RELEASE&format=html), [Bash Scripting](http://tldp.org/LDP/abs/html/exitcodes.html)
  
 
-### Exit Codes usable across platforms
-All exit codes defined should be usable on all supported Chef Platforms.  Also the exit codes used should be idential across platforms.  That limits the total range from 1-255.  Exit codes not explicitly used by Linux/Windows are listed below.  There are a total of 59 exit codes that overlap between the two platforms.
+### Remaining Available Exit Codes
+All exit codes defined should be usable on all supported Chef Platforms.  Also the exit codes used should be idential across platforms.  That limits the total range from 1-255.  Exit codes not explicitly used by Linux/Windows are listed below.  There are 59 exit codes that are available on both platforms.
  
  * Exit Codes Available for Chef use:
      * ~~35,37,40,41,42~~,43,44,45,46,47,48,49,79,81,90,91,92,93,94,95,96,97
@@ -41,7 +42,7 @@ All exit codes defined should be usable on all supported Chef Platforms.  Also t
 Exit Code        | Reason            | Details
 -------------    | -------------     |-----
 35               | Reboot Scheduled  | Reboot has been scheduled in the run state
-37               | Reboot Pending    | Reboot needs to be completed 
+37               | Reboot Needed     | Reboot needs to be completed 
 40               | Reboot Now        | Reboot being scheduled means it might run eventually.  Forced means its rebooting now
 41               | Reboot Failed     | Initiated Reboot failed - due to permissions or any other reason
 
@@ -52,12 +53,13 @@ Exit Code        | Reason             | Details
 0                | Successful run     | Any successful execution of a Chef utility should return this exit code
 42               | Audit Mode Failure |  Audit mode failed, but chef converged successfully.
 1                | Failed execution   | Generic error during Chef execution.  
--1               | Failed execution   | Generic error during Chef execution.  On Linux this will show up as 255, on Windows as -1  
+-1               | Failed execution*   | Generic error during Chef execution.  On Linux this will show up as 255, on Windows as -1  
 
+* \*Next release should deprecate any use of this exit code.
 
 
 ## Extend
-This RFC should be able to be ammended to include additional exit code functionality at a later date
+This RFC should be able to be amended to include additional exit code functionality at a later date.  Additional exit codes are assigned by pull request against this RFC as detailed in [RFC000](https://github.com/chef/chef-rfc/blob/master/rfc000-rfc-process.md#changing-an-accepted-rfc)
 
 ## Copyright
 
--- a/exit-status.md
+++ b/exit-status.md
@@ -7,7 +7,7 @@ Type: Standards Track
 
 # Title
 
-Signal outside tools of specific Chef-Client run status.  Ability to determine different results of a Chef-Client run.
+Signal outside tools of specific Chef-Client exit status. 
 
 ## Motivation
 
@@ -25,7 +25,7 @@ Signal outside tools of specific Chef-Client run status.  Ability to determine d
 
 ## Specification
 
-Chef-apply, Chef-client, Chef-Solo should honor the below exit codes
+Chef-apply, Chef-client, Chef-Solo should honor the below exit codes.  
 ### Use Exit codes
 Enumeration      | Exit Code    |Description
 -------------    | -------------| -----
--- a/exit-status.md
+++ b/exit-status.md
@@ -67,12 +67,16 @@ Exit Code           | Phase                             |Details
 Exit Code           | Phase                 |Details
 -------------       | -------------|        -----
 24001               | Reboot Scheduled      | Reboot has been scheduled in the run state
-20002               | Reboot Pending        | Reboot needs to be completed 
-20003               | Reboot Now            | Reboot being scheduled means it might run eventually.  Forced means its rebooting now
-20004               | Reboot Failed         | Initiated Reboot failed - due to permissions or any other reason
+24002               | Reboot Pending        | Reboot needs to be completed 
+24003               | Reboot Now            | Reboot being scheduled means it might run eventually.  Forced means its rebooting now
+24004               | Reboot Failed         | Initiated Reboot failed - due to permissions or any other reason
 
-
-
+#### Bootstrap Failures
+Exit Code           | Phase                         |Details
+-------------       | -------------|                -----
+25001               | chef-client download failure  | 
+25002               | Authentication failure        | 
+25003               | Environment doesn't exist     | 
 
 ## Copyright
 
--- a/exit-status.md
+++ b/exit-status.md
@@ -31,8 +31,17 @@ All exit codes defined should be usable on all supported Chef Platforms.  Also t
      * 213,219,227,228,235,236,237,238,239,241,242,243,244,245
 
 ## Exit Codes in Use
+
+#### Chef Run Generic
+Exit Code        | Reason            | Details
+-------------    | -------------     |-----
+0                | Successful run    | Any successful execution of a Chef utility should return this exit code
+1                | Failed execution  | Generic error during Chef execution.  
+-1               | Failed execution  | Generic error during Chef execution.  
+
+
 #### Reboot Requirement
-Exit Code        | Phase             | Details
+Exit Code        | Reason            | Details
 -------------    | -------------     |-----
 35               | Reboot Scheduled  | Reboot has been scheduled in the run state
 37               | Reboot Pending    | Reboot needs to be completed 
--- a/exit-status.md
+++ b/exit-status.md
@@ -52,7 +52,7 @@ Exit Code        | Reason             | Details
 0                | Successful run     | Any successful execution of a Chef utility should return this exit code
 42               | Audit Mode Failure |  Audit mode failed, but chef converged successfully.
 1                | Failed execution   | Generic error during Chef execution.  
--1               | Failed execution   | Generic error during Chef execution.  
+-1               | Failed execution   | Generic error during Chef execution.  On Linux this will show up as 255, on Windows as -1  
 
 
 
--- a/exit-status.md
+++ b/exit-status.md
@@ -7,7 +7,7 @@ Type: Standards Track
 
 # Title
 
-Signal outside tools of specific Chef-Client exit status. 
+Signal outside tools of specific Chef-Client run status.  Ability to determine different results of a Chef-Client run.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


carpnick:
If there are more use cases, please list and I will add them to the table.

carpnick:
FYI - @rschiefer, @danielvanhouten

mwrock:
This looks like a good start. One additional detail I'd like to see added is an exit code stating that a reboot is needed rather than scheduled. There is probably a better way to word this but the gist of things is to have 2 separate exit codes. One states that the node has initiated the reboot (scheduled) and the other to state that the node needs rebooting but has not actually taken action to reboot itself.

The latter would be used in cases where the local CLI tool wants to drive the reboot.

btm:
Chef::Application#exit! and #fatal! exit with -1 by default. We should try to clarify if -1 and 1 are different.

Looking at calls to `Chef::Application.fatal!` it looks like 2 is commonly used for configuration errors, or when SIGINT is received. We should probably change that. 3 is used for SIGTERM, but we typically squash that during a run and don't appear to modulate the exit code in any way to indicate that.

Lacking any real standard, we could consider using the same codes as [sysexists.h](http://www.opensource.apple.com/source/Libc/Libc-320/include/sysexits.h) for some things.

I would want to separate the exit codes that happen because something went wrong in Chef, e.g. compile failure, and something was triggered, e.g. reboot. Start the user caused ones at 80. Or be super smart and figure out how to make it easy to differentiate class of error using binary, e.g.

```
1000000 (64) - 1011111 (95): Chef raised error
1100000 (96) - 1101111 (111): Standard user errors (reboots)
1110000 (112) - 1111111 (127): Custom user errors
```

What "reboot pending" means is vague if you don't already know exactly what that means on Windows, and we should be clear here. I'd rather say "reboot needed."

mwrock:
Might be a good idea to have several reboot_pending codes to signify why (CBS Servicing, win update, pending file rename, etc).

exit-status.md
```diff
@@ -0,0 +1,46 @@
...
+-------------    | -------------| -----
+Success          | 0            | When Chef executes with a successful convergence and Audit Success
+Generic Failure  | 1            | When Chef executes and fails at convergence
+Compile Failure  | 50           | When Chef executes and Compile time phase fails
+Reboot           | 51           | When Chef executes and reboot is scheduled
+Reboot (Pending) | 52           | When Chef executes and reboot is pending. 
+Audit Failure    | 53           | When Chef executes and chef succeeds but Audit fails
+
+This list should be able to be expanded.  We should be conscious of typical exit codes that are used.  Example, Windows exit code 5 is commonly used for access denied.
+
```
thommay: we should look at the `sysexits(3)` manpage on os x/bsd for exit codes to avoid

thommay:
I'm generally :+1: on this.

btm:
Yeah, I'm :+1 with it generally:, but I don't want to accept it until we're sure what codes we want in what ranges. We need to expand that. Let's think about those now so we don't wish we had when they're near impossible to change later.

adamedx:
@chris-rock, audit failures are part of this proposal, and this would affect how we integrate with test-kitchen. Can you take a look?

jkeiser:
It'd be nice if we defined error ranges that we can extend later--i.e. 1000...1999 are converge failures; 2000...2999 are compile failures ... etc.

That way we can refine them later, for example if we decide to separate audit from converge failures we can set 1500..1599 as audit failures, and distinguish as deeply as we want.

adamedx:
So one question: what if a reboot is required / pending AND there's an audit failure? Maybe we just say one thing takes precedence over the other (i.e. you could assume that reboot  / reboot_pending is like an audit failure, so return that if both occur). Or we could combine the statuses using a bitwise flag-like or radix approach.

lamont-granquist:
255 (aka -1) should also probably be reserved for 'generic failure' as well, and most likely -1 to -10 (246 to 255 or so) should likely be avoided since they're common exit statuses of command line programs.  I'm not sure that we ever return those, but its probably a good idea to stay away from them.

btm:
@adamedx the core "reboot" use case is telling our parent we need to reboot, and presumably run again. In that case, I think we care about audit next time since we're not fully converged.

255 codes isn't enough bits for all codes to be bitwise. That's like 8 bits to play with, and a couple are going to be reserved.

carpnick:
Missed the IRC today.  But filled this out a bit more based on discussion.  Seems to go the way I thought where usually there are different ranges that mean different things.

I also added in some bootstrap items as well, separate PR?  When I think of things  that can fail on a node:
* Bootstrapping
* Running of Chef-Client

If we can exit correctly out of bootstrap, we might be able to build in logic to the knife* tools to have retries, or to hit different CDN's etc.  Example downloading chef-client installer.

btm:
@carpnick Unfortunately 0-255 isn't the range to exclude on Linux, it's the entire range that's available to us. Exiting anything above 255 returns the corresponding number in that range, e.g. 256 => 0, 1024 => 0, 1025 =>1.

We want most, if not all, of these codes the same on Linux and Windows, so we have to stay in 0-255. It might be nice some of our codes were similar to the sysexit or Windows codes in that range, but I think it's more important that we choose appropriate ranges for our sets of codes inside that range.

carpnick:
So that seems problematic.  If we start thinking in this use case, just read the documents in the PR to see which exit codes in the group are free:

### Windows Free exit codes
`
35,37,40-49,73-79,81,90-99,115-116,168-169,172,175-179,181,184,185,204,211,213,219,227-228,235-239,241-253
`
### Linux Free Exit Codes
`
3-63,79-125,131-254*
`
* Linux free exit codes are only free if we dont consider the kill -n use case.

So common exit codes that are left assuming we only use ones not documented in the couple of pages in the PR:

### Free for both Linux and Windows:
```
35
37
40-49
79
81
90-99
115
116
168
169
172
175-179
181
184
185
204
211
213
219
227
228
235 -239
241-253
```

Total common exit codes to work with:  59 across both platforms

btm:
@carpnick yeah, looking through the links you have the current doc, there's nearly zero consistency there. I'd say we consider 0,1,255 as predefined success, failure, failure, respectively. Lamont indicated 246-254 might be good to avoid too. So that gives us the range of 2-246 to figure out a good layout or set of ranges for different types of failures, then assigning specific failures we already know we care about.

carpnick:
Others are special to (Linux side only):
* 2	Misuse of shell builtins (according to Bash documentation)	empty_function() {}	Missing keyword or command, or permission problem (and diff return code on a failed binary file comparison).
* 126	Command invoked cannot execute	/dev/null
* 127	"command not found"

We dont want to accidentally run into either operating system exit codes, otherwise a simple user mistake, we could think actually means something to chef.

carpnick:
Add that Linux and Windows should overlap
Provide the range of overlap that can be used
Get rid of additional phases.  Only include the reboot section

carpnick:
Also I remember at chef-summit Seattle, that @adamedx thought it would be a good idea to have this for audit mode failures vs chef failures.  Just including that here since latest IRC meeting the scope was reduced to reboot support which does not take that into account

lamont-granquist:
Yeah @adamedx mentioned earlier in this thread about audit-vs-converge failures.   I think adding one more state for those would be pretty simple to both specify and implement.  The critical feedback in the meeting was more about the fine-grained instrumentation of chef client exit codes based on what was blowing up inside of a chef client converge phase.

exit-status.md
```diff
@@ -0,0 +1,67 @@
...
+41               | Reboot Failed     | Initiated Reboot failed - due to permissions or any other reason
+
+
+#### Chef Run State
+Exit Code        | Reason             | Details
+-------------    | -------------      |-----
+0                | Successful run     | Any successful execution of a Chef utility should return this exit code
+42               | Audit Mode Failure |  Audit mode failed, but chef converged successfully.
+1                | Failed execution   | Generic error during Chef execution.  
+-1               | Failed execution   | Generic error during Chef execution.
```
stevendanna: I'm not familiar with Windows, but for linux we might want to parenthetically remind users that if they are looking for this in an shell script, it is gonna be 255.

carpnick: Good call.

Linux:
```
# /opt/chef/embedded/bin/ruby -e "exit -1"
# echo $?
255
```

Windows:
```
>ruby -e "exit -1"
>echo %errorlevel%
-1
```

smurawski:
:+1:

exit-status.md
```diff
@@ -0,0 +1,67 @@
...
+
+Signal outside tools of specific Chef-Client run status.  Ability to determine results of a Chef-Client run.
+
+## Motivation
+    As a Chef user,
+    I want to be able to determine when a chef-client run is rebooting the node,
+    so that Test-Kitchen/Vagrant/any outside tool can wait for node to reboot, and continue converging.
+    
+## Specification
+* Chef-apply, Chef-client, Chef-Solo should honor the below exit chef run exit codes
```
btm: Mostly I want to mention utilities here and how they act, but maybe we can also avoid making an inclusive list here? There's an extra use of "exit" in here too.

* Chef applications (e.g. chef-client) that interpret recipes should use the specified exit codes
* Chef tools (e.g. knife) should behave appropriately for the exit code, or pass it to the user

exit-status.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Exit codes reserved across platforms
```
btm: Reserved by who? Chef? The links make it obvious if you read them, but it'd be nice if the header indicated to most people not to bother reading those links.

Exit Codes Reserved by the System?

exit-status.md
```diff
@@ -0,0 +1,67 @@
...
+* Windows- [Link](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681381(v=vs.85).aspx)
+* Linux - [Sysexits](http://www.freebsd.org/cgi/man.cgi?query=sysexits&apropos=0&sektion=0&manpath=FreeBSD+4.3-RELEASE&format=html), [Bash Scripting](http://tldp.org/LDP/abs/html/exitcodes.html)
+ 
+
+### Exit Codes usable across platforms
```
btm: Available Exit Codes, Remaining Available Exit Codes?

exit-status.md
```diff
@@ -0,0 +1,67 @@
...
+All exit codes defined should be usable on all supported Chef Platforms.  Also the exit codes used should be idential across platforms.  That limits the total range from 1-255.  Exit codes not explicitly used by Linux/Windows are listed below.  There are a total of 59 exit codes that overlap between the two platforms.
+ 
+ * Exit Codes Available for Chef use:
+     * ~~35,37,40,41,42~~,43,44,45,46,47,48,49,79,81,90,91,92,93,94,95,96,97
```
btm: What are the `~~`s for?

exit-status.md
```diff
@@ -0,0 +1,67 @@
...
+## Specification
+* Chef-apply, Chef-client, Chef-Solo should honor the below exit chef run exit codes
+
+### Exit codes reserved across platforms
+* Windows- [Link](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681381(v=vs.85).aspx)
+* Linux - [Sysexits](http://www.freebsd.org/cgi/man.cgi?query=sysexits&apropos=0&sektion=0&manpath=FreeBSD+4.3-RELEASE&format=html), [Bash Scripting](http://tldp.org/LDP/abs/html/exitcodes.html)
+ 
+
+### Exit Codes usable across platforms
+All exit codes defined should be usable on all supported Chef Platforms.  Also the exit codes used should be idential across platforms.  That limits the total range from 1-255.  Exit codes not explicitly used by Linux/Windows are listed below.  There are a total of 59 exit codes that overlap between the two platforms.
```
btm: "There are 59 exit codes that are available on both platforms."

exit-status.md
```diff
@@ -0,0 +1,67 @@
...
+* Precedence within a table should be evaluated from the top down.
+    *  Example - Audit Mode Failure would only apply on a successful execution.  But if the chef-run failed for any other reason, no reason to exit with audit mode.
+
+## Exit Codes in Use
+
+#### Reboot Requirement
+Exit Code        | Reason            | Details
+-------------    | -------------     |-----
+35               | Reboot Scheduled  | Reboot has been scheduled in the run state
+37               | Reboot Pending    | Reboot needs to be completed
```
btm: Can we call this "Reboot Needed" so we understand the system isn't going to do it for us?

exit-status.md
```diff
@@ -0,0 +1,67 @@
...
+41               | Reboot Failed     | Initiated Reboot failed - due to permissions or any other reason
+
+
+#### Chef Run State
+Exit Code        | Reason             | Details
+-------------    | -------------      |-----
+0                | Successful run     | Any successful execution of a Chef utility should return this exit code
+42               | Audit Mode Failure |  Audit mode failed, but chef converged successfully.
+1                | Failed execution   | Generic error during Chef execution.  
+-1               | Failed execution   | Generic error during Chef execution.  On Linux this will show up as 255, on Windows as -1
```
btm: Can we officially deprecate `-1`? I don't see the value in having two error codes for "generic error."

exit-status.md
```diff
@@ -0,0 +1,67 @@
...
+-------------    | -------------      |-----
+0                | Successful run     | Any successful execution of a Chef utility should return this exit code
+42               | Audit Mode Failure |  Audit mode failed, but chef converged successfully.
+1                | Failed execution   | Generic error during Chef execution.  
+-1               | Failed execution   | Generic error during Chef execution.  On Linux this will show up as 255, on Windows as -1  
+
+
+
+## Extend
+This RFC should be able to be ammended to include additional exit code functionality at a later date
```
btm: *amended

"Additional exit codes are assigned by pull request against this RFC as detailed in [RFC000](https://github.com/chef/chef-rfc/blob/master/rfc000-rfc-process.md#changing-an-accepted-rfc)"

exit-status.md
carpnick: Makes sense

exit-status.md
carpnick: Latter

exit-status.md
carpnick: This is to mention that they are strike-through.  In other words these codes are used in the spec.

exit-status.md
```diff
@@ -0,0 +1,69 @@
...
+### Exit codes Reserved by Operating System
+* Windows- [Link](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681381(v=vs.85).aspx)
+* Linux - [Sysexits](http://www.freebsd.org/cgi/man.cgi?query=sysexits&apropos=0&sektion=0&manpath=FreeBSD+4.3-RELEASE&format=html), [Bash Scripting](http://tldp.org/LDP/abs/html/exitcodes.html)
+ 
+
+### Remaining Available Exit Codes
+All exit codes defined should be usable on all supported Chef Platforms.  Also the exit codes used should be idential across platforms.  That limits the total range from 1-255.  Exit codes not explicitly used by Linux/Windows are listed below.  There are 59 exit codes that are available on both platforms.
+ 
+ * Exit Codes Available for Chef use:
+     * ~~35,37,40,41,42~~,43,44,45,46,47,48,49,79,81,90,91,92,93,94,95,96,97
```
btm: add a note as to what ~~ means? I presume these are codes we're using?

exit-status.md
```diff
@@ -0,0 +1,69 @@
...
+    *  Example - Audit Mode Failure would only apply on a successful execution.  But if the chef-run failed for any other reason, no reason to exit with audit mode.
+
+## Exit Codes in Use
+
+#### Reboot Requirement
+Exit Code        | Reason            | Details
+-------------    | -------------     |-----
+35               | Reboot Scheduled  | Reboot has been scheduled in the run state
+37               | Reboot Needed     | Reboot needs to be completed 
+40               | Reboot Now        | Reboot being scheduled means it might run eventually.  Forced means its rebooting now
```
btm: I think the use of 'forced' here is cruft from a prior "reason"

Change the details here to just "The system is rebooting" ?

btm:
:+1: A couple little nits, but I think this has come together. Thanks @carpnick

thommay:
Ok, by the power vested in me I approve this RFC. Let's get @btm's comments addressed and then we're off to the races. Thanks @carpnick

carpnick:
@btm  - Hopefully one last time.  Latest changes made.

andrewjamesbrown:
It would be interesting to have a special exit code that indicates that authentication failed with the client.pem, to e.g. trigger a process to delete the local client.pem and use the validator to create a new client object.  This would be useful for folks that have ephemeral chef servers that are rebuilt periodically.

coderanger:
@andrewjamesbrown If you aren't deleting the validation key ASAP (as soon as it is used once) that is a non-trivial security issue that you should address. There is no safe way to implement what you describe.

PR #159, opened 2015-10-22
Ohai Segments and Eager Loading

lamont-granquist:
Add an 'ohai' segment to cookbooks and eager load them after cookbook synchronization and library parsing, but before attributes/compile/coverge.

### DIFF
--- a/new/ohai-cookbook-segment.md
+++ b/new/ohai-cookbook-segment.md
@@ -34,8 +34,8 @@ The "segments" of a cookbook will be extended to include an "ohai" segment.  In 
 
 In the `Chef::RunContext::CookbookCompiler#compile` method a phase will be added after `compile_libraries` and before `compile_attributes` which will copy the Ohai plugins from the cookbook segment and will load all of the discovered plugins.
 
-The plugins will be copied from `<cookbookname>/ohai` into `/etc/chef/ohai/<cookbookname>` as their top level directory (recursively).  The state of the entire
-subdirectory tree under `/etc/chef/ohai` will be managed fully by chef-client so that any files which are not synchronized by chef-client will be removed, so
+The plugins will be copied from `<cookbookname>/ohai` into `/etc/chef/ohai/cookbook-plugins/<cookbookname>` as their top level directory (recursively).  The state of the entire
+subdirectory tree under `/etc/chef/ohai/cookbook-plugins` will be managed fully by chef-client so that any files which are not synchronized by chef-client will be removed, so
 that removal of a plugin from a cookbook or removal of the cookbook from `run_list` will result in the plugin being removed on the target host.
 
 The plugins directory will work similarly to libraries and other directions in that there will be no control over the inclusion of plugins below the level of the inclusion of the cookbook itself in the `run_list`.
--- a/new/ohai-cookbook-segment.md
+++ b/new/ohai-cookbook-segment.md
@@ -1,0 +1,45 @@
+---
+RFC: unassigned
+Author: Lamont Granquist <lamont@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Title
+
+Support ohai plugins under an `ohai` top level directory in cookbooks.  Load all
+ohai plugins in all synchronized cookbooks after cookbook synchronization.
+
+## Motivation
+
+    As a Chef User,
+    I want to have my custom Ohai plugins loaded on first bootstrap,
+    So I don't have to run chef twice.
+
+    As a Chef User,
+    I want my Ohai plugins loaded before attributes and compile/converge mode,
+    So I can use them without worrying about cookbook execution ordering.
+
+    As a Chef User,
+    I want my Ohai plugins synchronized with my cookbooks,
+    So that I don't incure more unavoidable round-trips to the Chef Server.
+
+    As a Chef Developer,
+    I want Ohai plugins as a first-class object,
+    So that I don't have to compile recipes to discover templates that drop plugins.
+
+## Specification
+
+The "segments" of a cookbook will be extended to include an "ohai" segment.  In this
+segment there will be plugins which are intended to be copied to the Ohai `plugin_path`.  All files in this segment will be copied, recursively, maintaining directory
+structure.
+
+In the Chef::RunContext::CookbookCompiler#compile method a phase will be added after
+`compile_libraries` and before `compile_attributes` which will copy the ohai plugins from the cookbook segment and will load all of the discovered plugins.
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.
--- a/new/ohai-cookbook-segment.md
+++ b/new/ohai-cookbook-segment.md
@@ -7,8 +7,8 @@ Type: Standards Track
 
 # Title
 
-Support ohai plugins under an `ohai` top level directory in cookbooks.  Load all
-ohai plugins in all synchronized cookbooks after cookbook synchronization.
+Support Ohai plugins under an `ohai` top level directory in cookbooks.  Load all
+Ohai plugins in all synchronized cookbooks after cookbook synchronization.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


lamont-granquist:
Throwing this up for discussion.  Came out of the discussions around Ohai at the summit.  In particular the run-chef-twice to get plugins issue.  When you look at what people do ceremonially in order to install plugins it also usually involves a template resource to install the plugin along with an Ohai resource to reload Ohai -- both of which run at converge time and unless the code has run before then all compile time references to those Ohai attributes will be broken the first time (and references in attributes files, etc).  A proposed solution to this is shipping the Ohai plugins with bootstrap code, but this means the user has to manually copy the distributed Ohai plugins in their cookbooks into their bootstrap code--bootstrap could try to scrounge around in the cookbook_path looking for Ohai plugins, but right now they're just random template resources dropped in the Ohai plugin path, so they're difficult to identify without converging a node.

There was much discussion of making Ohai plugins their own 'thing' completely separate from cookbooks.  When thinking about this option the path it leads down is that there's additional complexity and another 'thing' being added, and you'll need another whole round-trip to the chef server to expand Ohai plugins as well as cookbooks.  And when you think about where to put this phase it'll logically go right about where you synchronize cookbooks.

At the same time since Ohai plugins right now are just random templates in recipe code we can't easily identify them and do useful things with them until we actually do recipe parsing, so its difficult to make things better.

A solution that I think meets all the needs is to carve out Ohai plugins into their own 'segment' in cookbooks, and then eager load all the Ohai plugins early in cookbook compilation.  Then the distribution mechanism is still cookbooks (both on supermarket, and synchronization to the client from the server).  There's no need for an additional round trip.  By changing the client and adding a phase to cookbook compilation we can get the new attributes loaded very early which should satisfy the needs of people who presently find themselves unhappily running-chef-twice.  Also any attempt to solve the problem via modifying bootstrapping code to ship Ohai plugins similar to 'hints' will run into the problem of identifying which templates in community cookbooks are plugins.

Some issues I can think of here:

- Someone might actually want plugins to run before library parsing.  I considered adding an option to metadata force loading a plugin before library parsing, but I think those people can just lazy their access of the data.
- As written right now the plugins are files and not templates, so you lose templating (copying files in the bootstrapping process also loses templating though).  We could send `*.erb` files through erb though.

Those feel a little like overdesign to me, so I didn't include them to start with.

coderanger:
Adding new segments seems counter productive compared to nuking that system from existence. :+1: on the feature though.

ranjib:
this is one of those meta dependency requirements. chef_gem usage address exactly similar scenario. I dont think adding new segment solves the problem decently,  it will only work if the ohai plugin has no external dependency (gems etc). It certainly does not address the same bootstrapping problem across custom lwrps, where even to load them we need certain gems. We had several discussions on these in the mailing list, and following are two options on what i think will really help us  (install meta dependency):
- Fix bootstrap. Allow staged bootstrapping, where users can do staged chef runs. Custom ohai plugins, gems required by lwrps, custom handlers can be dropped during the first stage.. followed by the actual chef run. There are several system settings that also require similar workflows (like changing dhcp settings on aws). Have an orchestration engine powering the bootstrap (we use blender, Chef community can choose/write the one that fits us as a whole), and formalize a DSL (or yaml type) for bootstrapping, with hooks, error handling etc.
- Provide a first class workflow for repackaging omnibus. Users can add their own libraries, ohai plugins, lwrps , gems, and then use the resultant omnibus installer for bootstrapping
- Introduce initializers. i.e. similar to staged bootstrapping but for generic chef client run. Where whole cookbooks  marked as initializer run first in a separate chef run , followed by the rest of the cookbooks.

lamont-granquist:
@coderanger yeah should also RFC nuking segments from orbit, but wanted to keep this idea decoupled for now.

lamont-granquist:
So @ranjib I think external gem dependencies is an orthogonal issue.

We also have a problem where cookbooks can depend on external gems and while chef_gem can install the gem and make it accessible to recipe code there's no way to make it accessible to libraries, attributes or resources on first-run.  By extension I think we could solve a lot of that by allowing gem deps to go directly into metadata and adding another phase to cookbook compilation where all the gem deps are chef_gem-installed before library loading or something similar to that.

I've thought about omnibus-builds-as-a-service before, mostly just to solve the make-a-package-that-installs-somewhere-other-than-/opt/chef issue.  That is well outside of the scope of fixing the problems that people have right now with simple ohai plugins, and I just don't see that happening anytime soon (certainly nothing blocks anyone in the community from wiring that up though -- omnibus-chef + friends are all open-source).

I also don't see the need for added complexity in how chef runs occur with staged runs, orchestration, special initializer cookbooks, etc, when we haven't even considered how to modify and tweak the structure of the chef run that we currently have.  The benefit of this approach (and the approach of putting gem deps into the metadata) is that it leverages nearly all of the existing framework that we have now, with only minor tweaks and will definitely satisfy the 80/20 rule and cover 80% of use cases (if it doesn't actually wind up covering closer to 98%).

coderanger:
"By extension I think we could solve a lot of that by allowing gem deps to go directly into metadata" 😍😍😍😍😍

lamont-granquist:
^ could easily RFC that one, too, i might just spend all day firing off RFCs...  =)

nathwill:
> "By extension I think we could solve a lot of that by allowing gem deps to go directly into metadata" 

please oh please

jonlives:
Oh god so many +1s for cookbooks gem deps

martinisoft:
Huge :+1: on gem deps in metadata. It sounds like this could create some weird issues with dep solving, but as I dive deeper into shared libraries in cookbooks, the more I would want something like this because dependencies via libraries right now are sometimes tricky with order dependency which makes me :cry: sometimes.

new/ohai-cookbook-segment.md
```diff
@@ -0,0 +1,45 @@
...
+
+    As a Chef Developer,
+    I want Ohai plugins as a first-class object,
+    So that I don't have to compile recipes to discover templates that drop plugins.
+
+## Specification
+
+The "segments" of a cookbook will be extended to include an "ohai" segment.  In this
+segment there will be plugins which are intended to be copied to the Ohai `plugin_path`.  All files in this segment will be copied, recursively, maintaining directory
+structure.
```
mcquin: What's the protocol for resolving duplicate plugin file names? E.g., I have my own `hostname.rb` plugin file (hehe) under `cookbook/ohai/hostname.rb`. Would this conflict with or just override `lib/ohai/plugins/hostname.rb`? In the case of conflict, should we take the approach of the `ohai` cookbook and put custom plugins into their own directory?

btm: If we overwrite, we can workaround issues with existing core plugins temporarily. For instance, in https://github.com/chef/ohai/issues/420 a user is using an old version of hostname.rb to get back to older behavior that worked for them. 

Of course, I'd want to warn to keep someone from doing that forever or it happening unexpectedly, but then I'd want to be able to turn off the warning if my plugin replacement was never going upstream. Which is all a tiring pattern.

btm:
I'm wondering what level of control we want to give. As presented, every plugin in your cookbook is always loaded on all nodes. If you had a database cookbook with recipes for both clients and servers, with an ohai plugin to collect server data, it's going to get installed on the clients too. That's probably usually fine, as it should be up to the plugin to be well behaved. If you super-duper care about where the plugin is used, you give it it's own cookbook and control it's inclusion via the run_list. If you need to make that decision dynamically, e.g. with ruby in a cookbook, maybe that's your problem.

If we do load libraries first, perhaps someone could put ruby in there that adds the plugin to the list of plugins that are going to be loaded.

coderanger:
I think making it controllable which plugins go where (beyond the overall cookbook level) is probably over complicating things. Probably better to have the plugin written safely so that it no-ops if some required program or whatnot isn't installed, or make it require a hint file or config directive to enable.

btm:
@coderanger yeah, I agree the edges are probably rare and not worth the complication.

We should be clear if we're going to allow overwriting core plugins. Other than that, :+1:.

lamont-granquist:
Yeah I think conditional compilation should be like attribute files--if you need it to not appear on a node then you need to break it out into its own cookbook and make sure it doesn't appear in the run_list.

coderanger:
Note from IRC, we should consider how to clean up ohai plugins that came from a cookbook if that cookbook is no longer applied to the machine. Maybe write them to `/etc/chef/ohai/<cookbookname>/<pluginname>.rb` so that they can be efficiently removed.

lamont-granquist:
So will amend PR:

- note that it works like attributes files and the level of inclusion control is at the cookbook level
- add cleanup code and filesystem namespace
- warn on overriding plugins, but include a way to suppress the warn

thommay:
This looks like goodness :+1:

adamhjk:
@lamont-granquist - make a call on final directory location, update hte RFC, and then it's approved.

lamont-granquist:
After reflection I don't find the /var arguments compelling, so think the RFC can just stand.

I don't see any substantive difference between this config synchronizer dropping config files into /etc vs chef-client template resources later in the run dropping config files into /etc with "#updated by chef, do not touch" banners on them.  Feels like its getting into really nitpicking territory over the FHS and I don't see any value to /var over /etc.

However there is value in /etc over /var which is just the simplicity of having one place for config files, and that lowers the barrier to new users, which I think is the clear win.  One less thing.

This is all subject to the law that code beats bikesheds any day in my book and since there's debate over the question the actual implementer should feel free to override my opinion, and we can revisit in the PR against the code.

thommay:
Merged, given @adamhjk's pre-approval.

lamont-granquist:
Adding a couple notes that came up during ohai review:

someone wants to override the fqdn value with a plugin of their own, we should support that use case properly:

https://github.com/chef/ohai/issues/607

someone wants to use custom libraries from within a plugin of their own, we should support that use case properly:

https://github.com/chef/ohai/issues/520

PR #160, opened 2015-10-22
add 'gem' to cookbook metadata

lamont-granquist:
Allow cookbooks to specify gems in metadata.rb/metadata.json:

gem 'poise'
gem 'chef-sugar'

And those will be installed right after synchronization and usable in libraries, etc.

### DIFF
--- a/new/metadata-gem-installation.md
+++ b/new/metadata-gem-installation.md
@@ -31,7 +31,7 @@ The implementation will use an in-memory bundler Gemfile which is constructed ag
 at the same time.  The syntax of the 'gem' statement will support the bundler gem syntax, with the qualification that since it is compiled into metadata.json
 that arbitrary ruby code will be expanded at cookbook upload time.
 
-The resulting gemset bundle will be installed into the ruby that chef-client is running out of (typically omnibus ruby).
+The resulting gemset bundle will be installed into the LIBPATH of the running chef-client.  This may either be directly into the base ruby libraries (per current `chef_gem` behavior) or into a custom location with the LIBPATH of the chef-client extended to use that location--as an open implementation question.
 
 The normal Gemfile `requires` tag may be used by users to autoload files out of gems.
 
--- a/new/metadata-gem-installation.md
+++ b/new/metadata-gem-installation.md
@@ -23,9 +23,17 @@ Allow users to specify additional gem dependencies like:
 
 gem "poise"
 gem "chef-sugar"
+gem "chef-provisioning"
 
-In the Chef::RunContext::CookbookCompiler#compile method a phase will be added before `compile_libraires` which will install all of the gem declarations from all of the synchronized cookbooks before any other
-cookbook code is compiled.
+In the `Chef::RunContext::CookbookCompiler#compile` method a phase will be added before `compile_libraires` which will install all of the gem declarations from all of the synchronized cookbooks before any other cookbook code is compiled.
+
+The implementation will use an in-memory bundler Gemfile which is constructed against all gem statements in all cookbooks which are in the `run_list`, solved
+at the same time.  The syntax of the 'gem' statement will support the bundler gem syntax, with the qualification that since it is compiled into metadata.json
+that arbitrary ruby code will be expanded at cookbook upload time.
+
+The resulting gemset bundle will be installed into the ruby that chef-client is running out of (typically omnibus ruby).
+
+The normal Gemfile `requires` tag may be used by users to autoload files out of gems.
 
 ## Copyright
 
--- a/new/metadata-gem-installation.md
+++ b/new/metadata-gem-installation.md
@@ -15,7 +15,7 @@ other cookbook loading is done.
 
     As a Chef User,
     I want to be able to use additional gems in libraries, attributes and resources,
-    So I don't pull my hair/beard out in frustration.
+    to avoid complex workarounds and double-run converges.
 
 ## Specification
 
--- a/new/metadata-gem-installation.md
+++ b/new/metadata-gem-installation.md
@@ -25,7 +25,7 @@ gem "poise"
 gem "chef-sugar"
 gem "chef-provisioning"
 
-In the `Chef::RunContext::CookbookCompiler#compile` method a phase will be added before `compile_libraires` which will install all of the gem declarations from all of the synchronized cookbooks before any other cookbook code is compiled.
+In the `Chef::RunContext::CookbookCompiler#compile` method a phase will be added before `compile_libraries` which will install all of the gem declarations from all of the synchronized cookbooks before any other cookbook code is compiled.
 
 The implementation will use an in-memory bundler Gemfile which is constructed against all gem statements in all cookbooks which are in the `run_list`, solved
 at the same time.  The syntax of the 'gem' statement will support the bundler gem syntax, with the qualification that since it is compiled into metadata.json
--- a/new/metadata-gem-installation.md
+++ b/new/metadata-gem-installation.md
@@ -1,0 +1,35 @@
+---
+RFC: unassigned
+Author: Lamont Granquist <lamont@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Title
+
+Support a 'gem' DSL method for cookbook metadata to create a dependency on a rubygem.  The
+gem will be installed via `chef_gem` after all the cookbooks are synchronized but before any
+other cookbook loading is done.
+
+## Motivation
+
+    As a Chef User,
+    I want to be able to use additional gems in libraries, attributes and resources,
+    So I don't pull my hair/beard out in frustration.
+
+## Specification
+
+Allow users to specify additional gem dependencies like:
+
+gem "poise"
+gem "chef-sugar"
+
+In the Chef::RunContext::CookbookCompiler#compile method a phase will be added before `compile_libraires` which will install all of the gem declarations from all of the synchronized cookbooks before any other
+cookbook code is compiled.
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.


coderanger:
So a mechanism we would probably want is something like Halite's `cheftie.rb` loading system. It scans the gem code for a file named `cheftie.rb` and requires it for you. This allows stuff like registering resources/providers or DSL methods more directly. We could also say that Chef will require the gem name like Rails does, but less a fan of that approach.

lamont-granquist:
So this is another Minimum Viable RFC.

Obviously if you're trying to install a native gem and your image doesn't have compilers baked in then you're gonna have a bad day.  But it does solve some problems for common chef-enhancement gems that are common and which have no native dependencies, and which result in a lot of pain.

This does, oddly enough, go against the idea of moving gem installation to converge time and avoiding the use of compile-time installation of chef_gems.  However, we can't solve isues like being able to use chef-sugar DSL methods in attributes files with that approach.  It feels to me like this RFC better addresses the remaining use case left over *after* we try to push everyone towards not using compile_time for their chef_gem resources and making their resources lazy-require dependent gems.

I'm also not quite clear where we should 'put' the chef_gem resource that we use to install the gem.  Does that go into the parent resource collection and then get converged?  Do we open up a new special resource collection for all of these?  Do we not worry about resource collections at all and just build_resource these resources and then we still get reporting results via the event handlers firing when the install action runs?

lamont-granquist:
@coderanger yeah I was just thinking about how to get the gem installed, i hadn't considered hooks for auto-requiring resources.

Could we make that more generically configurable?

gem 'poise-ruby', require: [ '/lib/poise_ruby/cheftie.rb' ]

We could sugar that a bit with:

gem 'poise-ruby', cheftie: true

Being magic for searching through and finding all the cheftie.rb's in the gem.

coderanger:
@lamont-granquist Yeah, that sounds good on the config but we should make the default something that allows registering resources in one way or another.

ranjib:
what about versions? what if the gem requires a native package to be installed, is that out of scope?

coderanger:
I would say it should probably match Gemfile syntax. So `gem 'name', 'version', require: 'other'`. Native packages I would say is out of scope to worry about it.

lamont-granquist:
Yeah, version pins definitely.  I wonder if we could even support pulling from git/github if you have that installed on your system?  That would probably require bundler?  But we ship with bundler.  We could ship git in the omnibus packages.  I think that's is definitely getting out of scope though.  I'll definitely amend it to support version requirements/pins.

Native gems I think are inherently out of scope.  I think I'm going to cut another RFC to just ship nokogiri in omnibus-chef which I suspect is the 80% of that particular problem.

coderanger:
@lamont-granquist A solid solution to both the git install stuff and avoiding `chef_gem` would be to use the `gem` statements to build an in-memory Gemfile (via `Bundler::Gemfile` directly) and then install/activate it.

thommay:
:+1: so much yes! And I think Noah's suggestion of building an in memory Gemfile and just using bundler for the whole thing is solid.

lamont-granquist:
Okay, so we'll implement using in-memory Gemfile.  We'll leverage all the Gemfile syntax that we can and which makes sense.  Then we'll also include the 'requires:' specification to autoload files within the gem.  The 'cheftie' thing will get dropped, at least for now, per @coderanger

coderanger:
For future record: I asked to drop the cheftie thing because it could be implemented internally with just a Rails-style loading via `require 'poise/cheftie' if Chef.node` as the node is only set if we are currently in a converge.

ranjib:
currently everything that mutates a system goes in side recipes.. or gets triggered from there. if we implement this, how will that affect? if i declare a cookbook as dependency , but then not include any recipe from it, will the gems from dependency cookbook be installed? or it will be conditional ?

coderanger:
@ranjib It would get pulled in, everything in the dependency set would be loaded just like now for support files (libraries/ etc).

jkeiser:
:+1:

ranjib:
@coderanger it wont be good then right? like lots of cookbooks depends on windows related stuff for crossplatform thing, if they start declaring windows specific gems in metadata, it will bork the chef run

coderanger:
Also the statement that anything that mutates system state goes in recipes is aspirational at best ;-) Make a `libraries/default.rb` with ``` `rm -rf /` ``` and see what happens ;-)

coderanger:
@ranjib The goal is only to support pure-Ruby gems, which should install on any platform. They might not work, but they should at least install.

lamont-granquist:
Yeah that problem is identical with windows-specific resources required in core chef on Ubuntu boxes.

carpnick:
if this is pure ruby cant we do stuff like:

```
gem win32-api  = 1.1 if windows_os?

or

if windows_os? 
  gem win32-api  = 1.1 
else
  gem whatever = 1.1
end

```

or something like it?

ranjib:
@coderanger thats not the point, users can do something like this anywhere, thats intentional, not by design.  while the metadata specified automatic gem installation is by design. this breaks the whole recipe run list specified config management model. think about handlers, resource_updated style event hooks, where do they fit in for such gems. will we be able to do conditionals on the metadata? my understanding is in chef servers its a json file, not a ruby thing.

coderanger:
@carpnick Doing conditional stuff will have all the same problems as conditional `depends`, and if we ever come up with a satisfactory solution there we can easily use it for gems too. Specifically the metadata is baked to JSON on upload so code in the `metadata.rb` doesn't help.

@ranjib Still not sure what you mean, how would this be any more mutating the system than `libraries/` files getting loaded as they do today? This just allows having the equivalent of `libraries/` but the code happens to come from a gem. This can be done today with Halite, but it would be nice to make it simpler so people don't have to learn my weirdo tooling (and Halite comes with its own limitations like issues with lazy cookbook file loading).

ranjib:
@coderanger when libraries files are loaded, they dont install gems . it is not so much about mutating it, it more like i would not expect the metadata segment to trigger the mutation.
remember my earlier suggestion about stages and initializers? this is a miniature stage, except instead of the full resources you have only gem resource, also this is described in metadata, not in recipe or any other segment (i.e. content of a cookbook). To me if a cookbooks metadata specifies a gem, probably we can ensure that the gem is present before converging the cookbook and fail if not. i.e metadata mandates a requirement, but not does the actual installation. and it does not solve lot of other use cases with similar problem, for mysql gem based provider, i might need some native library as well,, and i wont be able to do that in this approach.
may be im not able to articulate, but even after reading all your and lamont's comment i feel this is half baked and very inconsistent with chef resource-convergence mental model. I definitely understand it solves 80% of that use case (which is pure ruby gems) and i would have ok with it if it was just half baked but not inconsistent. But its really inconsistent with how chef works, are how i understand chef works.. that all code leads to an expanded run list which is set of resources, then actual convergence begins. .. anyway i wont argue any more. lets see what others think :-)

lamont-granquist:
I can see your argument, but since this is being used to install gems into chef itself I think it falls under the model more of cookbooks and the file cache.  The whole cookbook synchronizer mutates state on the system under control and does so outside of any resources or the resource collection.  There is no reporting on it.  Files are not backed up.  The cookbook synchronizer is also 'half-baked' by all of your criteria.

I was initially considering using the chef_gem resource to do the work, but the in-memory Gemfile will work substantially better and be more flexible and automatically support the 'requires' syntax and many other extensions.  I'd prefer to have both the simplified implementation and to leverage the bundler library and syntax, rather than ship something that was compliant with the chef resource model but more complicated and without any of the additional flexibility.  I also don't see anything in practice that is harmed by not using the chef resource model.

So while you see it half-baked, I see the alternative as unnecessarily chaining us to a design philosophy that is poorly fit to solving this problem and which over complicates the implementation and potentially cripples some useful features.

lamont-granquist:
There also is the very obvious fall back that if users 'need' the chef resource model they can use chef_gem and get the chef resource model.  If you have to do a notify or subscribe off of the installation of them, then you simply don't want to use this--and i definitely don't want to solve the problem of having an action at metadata-parsing-time throw a notification forward into recipe code.  And that just isn't a use case that the users of this kind of pattern want (and you don't normally see notifications hanging off of chef_gem installs anyway, at least I can't recall the last time I saw one).

zuazo:
I have doubts, but I think I agree with @ranjib on this matter.

Summarizing:

### Use the Metadata

Example:
```ruby
# metadata.rb

depends 'nginx'
gem 'my-sweet-gem', '~> 1.0'

# and maybe in the near future:
package 'sugar' # my-sweet-gem requires sugar
```

Pros and cons:
```diff
+ Simple to understand and use.
+ Seems easy to implement.
+ Usable in attributes.
+ Support for version constraints.
- Inconsistent with the other settings: for instance, the `depends` does not install the cookbook.
- Rather limited environment. Does not support conditional statements.
- Support for required packages seems forced. It's almost turning it into a recipe.
```

### A New Resource Collection
OK, I put this as the radically opposite solution. There could be other viable intermediate solutions.

Example:
```ruby
# bootstrap/sugar_gem.rb, recipe/bootstrap.rb, cheftie.rb or whatever.rb

package 'gcc' if node['os'] == 'linux'
chef_gem 'my-sugar'

# and maybe in the near future:
depends 'apt' do # better to use cookbook 'apt' do ..??
  only_if { platform_family?('debian') }
end
```

Pros and cons:
```diff
+ Usable in attributes.
+ Support for version constraints can be implemented.
+ Consistent with how chef works.
+ Full support of statements conditionals.
- Sounds a bit over-engineered, it does not seem so simple to implement.
- Chicken or the egg dilemma may arise when requiring other cookbooks.
- Has some limitations like notifications or the support of some specific resources.
```

Personally I'd rather not use the metadata for this. Mainly because even though it sounds easier in the short term, it looks it will be be more difficult to expand it in the future considering it has to be a JSON.

lamont-granquist:
Breaking the chef-client run up into yet more stages is also more 'phases' to add on top of compile/converge mode which adds more cognitive complexity in addition to the code complexity.

This solution is substantially less flexible and doesn't allow arbitrary code which is a '+' in my mind because then its subject to less coding abuse, and the incremental amount that new users will need to learn will be lower.

Multiple phases is also just not likely to get implemented, I don't see anyone who is interested in writing that code.  If we go down this road and wind up discovering the gaps that we really need to fill, then we'll know why we need multiple phases at some later date and people will be more interested in writing that code.  For now, this solves a real problem that we have, in a fairly simple way, and eliminates all the extremely ugly code spread around the gem-cookbook world.

I think you're trying to engineer this RFC into a multiple phases RFC and I don't see that too likely to happen, you'll be better off accepting this one and then at a later date arguing that there's still gaps in what we can do.

zuazo:
Not at all. In fact, personally I hate the multiple phases of a Chef Run and I'm not in favor of adding another. I wrote this alternative as a radically different solution example and for contrast: to think of other ways to accomplish the same thing.

Perhaps using some kind of Gemfile instead of the metadata would be a more realistic alternative. But the metadata is clearly the winner there because using a Gemfile does not add any *chef-related* features.

Regarding *allowing arbitrary code* & *code abuse* I interpret it more as a *requested feature*. And the flexibility in Chef is a '+' in my mind.

Anyway, coming back to the reality, the truth is that using the metadata is not such a bad idea considering the possibilities we have.

Just for the record, I'm not hiding any shady purpose with this RFC. I promise :innocent:

jkeiser:
`gem` in metadata has two clear advantages over a separate phase:

1. It's simpler.
2. Dependencies from multiple cookbooks are solved *together* and the versions are applied to .

The second point is not mere cosmetics. A cookbook is solving a *different problem* than the `gem` metadata. Cookbooks let you install gems one by one, each satisfying its own requirements. `gem` takes gem requirements from all cookbooks together, and installs a single version of each gem that works for *all* of them.

Here is a scenario.
* Your run list is 'a,b'.
* Cookbook a is OK with any version of chef-provisioning from 1.0 on up.
* Cookbook b needs version 1.3.0 of chef-provisioning to function.

My question: how does each method handle this, and what version of the gem is loaded when `require 'chef-provisioning'` happens?

## Gem Metadata

The `gem` case is straightforward:

```ruby
# cookbooks/a/metadata.rb
name 'a'
gem 'chef-provisioning', '>= 1.0'
```

```ruby
# cookbooks/b/metadata.rb
name 'b'
gem 'chef-provisioning', '= 1.3.0'
```

In this case:
1. The `Gemfile` we create will have `gem 'chef-provisioning', '>= 1.0', '= 1.3.0'`, so chef-provisioning 1.3.0 gets installed.
2. `require 'chef-provisioning'` will load version 1.3.0 (even if there happens to be a later version on the system for some reason) because it's integrated with gem activation..

## "Preload" Cookbook Phase

Now let's look at what happens with a cookbook phase. For concreteness I imagine this as a `preload` recipe which is loaded in the order of the run list in the same manner as `libraries` or `resources`:

```ruby
# cookbooks/a/recipes/preload.rb
chef_gem 'chef-provisioning' do
  version '>= 1.0'
end
```

```ruby
# cookbooks/a/recipes/preload.rb
chef_gem 'chef-provisioning' do
  version '= 1.3.0'
end
```

What will happen:
1. chef-provisioning 1.4.1 will be installed by the first `chef_gem`.
2. chef-provisioning 1.3.0 will be installed by the second `chef_gem`.
3. `require 'chef-provisioning'` will load version 1.4.1

And thus, cookbook B breaks because it needed exactly version 1.3.0.

I can imagine ways to fix this, sure, but not without but not without inventing even *more* things than just a new cookbook phase, and bringing in the complexity that comes with it :)

I can imagine things I might like to do in a preload phase (build-essentials in particular comes to mind). But even if we determine we need a preload phase for something like that, I would prefer to have *both*.

zuazo:
@jkeiser That's a really good point. You have my :+1:

new/metadata-gem-installation.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Specification
+
+Allow users to specify additional gem dependencies like:
+
+gem "poise"
+gem "chef-sugar"
+gem "chef-provisioning"
+
+In the `Chef::RunContext::CookbookCompiler#compile` method a phase will be added before `compile_libraires` which will install all of the gem declarations from all of the synchronized cookbooks before any other cookbook code is compiled.
```
btm: *compile_libraries

stevendanna:
I'm :+1: on this idea.  

I'm not sure if it needs to be in the RFC, but I think that it will be very important for usability that we expose gem installation failures well.  Namely, however we store the combined Gemfile, we likely want to store some sort of annotation so that we can clearly show user what cookbook is introducing the problematic gems. i.e something like:

> Gem installation phase failed:
>    - Could not satisfy dependency on foo-2.3 because BLAH
>    - foo-2.3 is required by the foobar cookbook
> See /var/cache/chef/gem-install-failure.log for the full log data

lamont-granquist:
This got blocked by the discussion around if multiple convergence failures was a better and more comprehensive solution to this, waiting on a proposal for what that would look like from @ranjib

ranjib:
@lamont-granquist i doubt i'll have to time to work on the initializer, a proposal assumes author is willing to implement it. i want to finish the telemetry one first. 
in memory gemfile and bundler based implementation sounds reasonable to me,
:+1:

btm:
:+1:

new/metadata-gem-installation.md
```diff
@@ -0,0 +1,43 @@
...
+
+Support a 'gem' DSL method for cookbook metadata to create a dependency on a rubygem.  The
+gem will be installed via `chef_gem` after all the cookbooks are synchronized but before any
+other cookbook loading is done.
+
+## Motivation
+
+    As a Chef User,
+    I want to be able to use additional gems in libraries, attributes and resources,
+    So I don't pull my hair/beard out in frustration.
```
btm: As a Chef User,
I want to be able to use additional gems in libraries, attributes and resources,
to avoid complex workarounds and double-run converges.

thommay:
I approve this RFC. Take it away, @chef/rfc-editors

PR #161, opened 2015-10-23
Cookbook Segment Deprecation

lamont-granquist:
The elimination of cookbook segments from the server protocol that has been discussed for awhile now.

### DIFF
--- a/new/cookbook-segment-deprecation.md
+++ b/new/cookbook-segment-deprecation.md
@@ -11,10 +11,9 @@ Cookbook Segment Deprecation
 
 ## Motivation
 
-    As a Chef User/Developer,
-    I want to be able to extend the structure of cookbooks without fighting with a wire protocol
-    written 6 years ago,
-    Because nobody wants that.
+    As a Chef User/Developer
+    I want to be able to extend the structure of cookbooks easily,
+    in order to quickly adapt to new automation needs.
 
 ## Specification
 
--- a/new/cookbook-segment-deprecation.md
+++ b/new/cookbook-segment-deprecation.md
@@ -17,35 +17,29 @@ Cookbook Segment Deprecation
 
 ## Specification
 
-This RFC will add a special `all_files` segment to cookbooks which will duplicate (initially)
-all of the existing cookbook manifest along with arbitrary files outside of the current scope of
-any of the cookbook segments.  The Chef Server will need to be patched in order to support this
-new 'segment' and accept it since the Chef Server currently does strict enforcement of file
-segments server-side.  This will also generate a version bump of the Chef Server protocol which
-will be used by servers to determine if the Chef Client supports this field or not.  Clients MAY
-implement code which determines the acceptable API version of the Server and then only send
-the new API requests to the server.  Clients MAY instead send requests which include both the
-old segment and new segment data.  The Server MUST detect the Client's desired API version and
-only send responses formatted for that version.  The Server MUST be able to receive both old
-API requests and 'mixed' API requests with both the old segments and the new segment.
+This RFC specifies a new format for the GET and PUT requests to the cookbooks/NAME/VERSION
+endpoint.  This is a breaking change which will mandate a bumping of the API version to
+the protocol.  The existing 'segments' will be removed out of the cookbook version and a
+new 'all_files' segment will be introduced which will simply be a list of all the files in
+the cookbook.
 
-New knife clients that support this segment MAY use the Chef Server protocol version to determine
-if they MUST upload cookbooks using the old segment format, or if they CAN upload using only
-the `all_files` format.
+The Chef Server MUST respond to all GET requests that do not contain an appropriate API version
+with the old protocol with segments.  Cookbooks that have been uploaded in the new format
+MUST have their manifest information filtered so that an old style response can be constructed.
+When the Chef Server sees an API version in the GET request that accepts the new style it
+MUST respond only with the 'all_files' segment in the body of the response.
 
-For backwards compatibility, the Chef Server MUST serve cookbooks with both standard segments and the
-new `all_files` segment unless it can determine that the client prefers one or other.  If it can
-be determined then the server MUST serve cookbooks to old clients with the old format, and SHOULD
-serve cookbooks to new clients with only the `all_files` segment.
+All new Clients MUST set their API version correctly in order to get the new behavior on
+PUT or GET.  Since old Servers will not accept the new 'all_files' segment Clients MUST determine
+the server version they are talking to and send their PUT requests correctly.  Clients MAY
+use prior communication with the Chef Server (i.e. during the uploading of sandbox files they
+MAY determine the API version of the Chef Server off of the replies and use that information) to
+determine the correct API version to use and format their PUT request accordingly.  Clients
+MAY also PUT with the new format and after receiving a 4xx code from the Server retry the
+request in the old format and downgrade.
 
-New Chef Clients MUST accept either old style segments, cookbooks with the new `all_files` segment
-or cookbook metadata with both styles (and SHOULD favor the `all_files` segment).
-
-This will achieve backwards compatibility and the ability for knife, chef-client and the chef-server
-to be upgraded independently of each other and run in a "dual-stack" mode.  At some point in the
-future, backwards compatibility will be dropped in a major version bump.
-
-The implementation of this RFC must still fully support both settings of the `no_lazy_load` config parameter.
+The implementation of this RFC must still fully support both settings of the `no_lazy_load`
+config parameter.
 
 ## Copyright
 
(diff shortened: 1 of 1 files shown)


lamont-granquist:
Some noodling on the problem was here:

https://github.com/chef/chef_objects/pull/79

Somewhere there's a discussion between me + @danielsdeleo where he suggested how the backcompat should work.  Now that we have API versioning this should work even better (and might work even better than I've outlined here, I haven't looked at the details of that).

I don't like `all_files` very much but `files` would be my first choice and that top-level key is taken by the `files` segment already.  Maybe it'd be safe to leverage API versioning harder in order to 'break' this key (?) but I'd want someone who knew more about the implications of that to lead me down that road...

jaymzh:
How does this affect lazy loading?

coderanger:
Not at all, the actual loading code doesn't really use segments, it uses direct path globs like you would expect. Other than the file sync code and the templates/cookbook_files system, I don't think much really cares about segments as is.

jaymzh:
Well that's what I"m talking about the - the templates / cookbook_files system part of Chef. If I have a template, chef-client (if I've told it to) will lazy-load that source of the template only if it ends up needing it. You (for reasons I can't fathom) turned this option off in Chef 12, but we turned it back on (cause performance)...

danielsdeleo:
Nothing about this would make lazy loading impossible, there would just be a slightly different implementation internally (i.e., you would know that files are templates because the first element of the path is "templates").

As for the server's versioning capability, we're in a crappy spot where we can't use it to its full potential yet. This is because 11.x servers are still supported by 12.x clients, but 11.x servers are totally unaware of versioning. Therefore, they are incapable of rejecting requests for API versions that they don't support.

I don't remember exactly the details of what we decided for the compatibility behavior. Probably we need to think through the compat matrix of:

* server supports non-segments: (t/f)
* uploading client supports non-segments: (t/f)
* downloading client supports non-segments: (t/f)

jaymzh:
OK so it sounds like it's possible but will require some work to make work in the new model. I just want to make sure that's taken into account. Our performance *tanked* when lazy loading became optional, (and we put a bunch of stuff on hosts that didn't need it... ), so that has to stick around.

lamont-granquist:
Yeah, I wasn't planning on taking that away at all.  The way it works now is that you take all the files in the manifest and nuke the `files` and `templates` segments and use the resultant list to synchronize, its pretty straightforwards to just remove paths starting with `files/` and `templates/` instead.  We can call that out, but it just isn't going to be a change, so I didn't include it in the RFC.

jaymzh:
Cool. That makes sense. We should call it out just so it doesn't get dropped (since the default doesn't do this anymore, I can see someone not thinking about it when they go to implement it).

danielsdeleo:
The code would be exactly like the code that knife uses to categorize files into segments now, I would expect.

jaymzh:
@danielsdeleo  I'm not worried about knife... :)

lamont-granquist:
Okay so @danielsdeleo it sounds like I probably wrote this the correct way then.  Since we use a new 'all_files' segment we don't change the semantics of `files` and don't break the back-compat world there, and then the RFC is written dependent upon what can be determined on either side of the connection, and I called out the behavior that should occur if you can't determine the protocol to use as well.  Hopefully its also clear that back-compat takes priority over the letter of what is written in this RFC if it turns out we got any of it wrong.

I didn't want to go into details of the implementation that we'd talked about but one of the ideas was to use the initial sending of the cookbook files up to the server to determine the server version so that the PUT of the metadata at the end of the upload could pick the correct format to use.  We could use that, or we could use API versioning, etc.  I think that can be handled when writing the code though and in PR review...

thommay:
:+1:

btm:
:+1:

new/cookbook-segment-deprecation.md
```diff
@@ -0,0 +1,51 @@
...
+new 'segment' and accept it since the Chef Server currently does strict enforcement of file
+segments server-side.  This will also generate a version bump of the Chef Server protocol which
+will be used by clients to determine if the Chef Server supports this field or not.
+
+New knife clients that support this segment will use the Chef Server protocol version to determine
+if they MUST upload cookbooks using the old segment format, or if they SHOULD upload using only
+the `all_files` format.
+
+For backwards compatibility, the Chef Server MUST serve cookbooks with both standard segments and the
+new `all_files` segment unless it can determine that the client prefers one or other.  If it can
```
stevendanna: I believe we should be able to use the API version here as well.  For versions before this was introduced or no-version we return the current response.  For versions after it was introduce, serve the new response.  That is, I'm unclear on the cases where we would be unable to determine client preference.

What should the server do with all_files members that don't map to the old segments?

stevendanna:
Tentatively :+1: but I think for new API responses, the RFC should include example payloads and example returns.  Also, we need to specify what the version 0 or version 1 API response will look like when a cookbook was uploaded using all_files that includes segments that don't map to the existing segments.

new/cookbook-segment-deprecation.md
```diff
@@ -0,0 +1,51 @@
...
+be determined then the server MUST serve cookbooks to old clients with the old format, and SHOULD
+serve cookbooks to new clients with only the `all_files` segment.
+
+New Chef Clients MUST accept either old style segments, cookbooks with the new `all_files` segment
+or cookbook metadata with both styles (and SHOULD favor the `all_files` segment).
+
+This will achieve backwards compatibility and the ability for knife, chef-client and the chef-server
```
stevendanna: Doesn't need to be in the RFC, but we need to make sure that cheffs style commands are also compatible with this (we've broken those on past API changes) as well as widely used tools such as knife-backup and knife-ec-backup.

new/cookbook-segment-deprecation.md
```diff
@@ -0,0 +1,51 @@
...
+
+# Title
+
+Cookbook Segment Deprecation
+
+## Motivation
+
+    As a Chef User/Developer,
+    I want to be able to extend the structure of cookbooks without fighting with a wire protocol
+    written 6 years ago,
```
stevendanna: I lol'd but I'm not sure this is the best motivation description :).  How about,

> As a Chef User/Developer
> I want to be able to extend the structure of cookbooks easily,
> in order to quickly adapt to new automation needs.

smurawski:
:+1:

new/cookbook-segment-deprecation.md
danielsdeleo: > I believe we should be able to use the API version here as well.
I could go either way. What we have to do is support a matrix of old client/new server, and new client/old server, and we won't be able to 100% rely on versioning in the server until at least Chef 13.

> What should the server do with all_files members that don't map to the old segments?
You're gonna need a client and server that support "segmentless" cookbooks to take advantage.

new/cookbook-segment-deprecation.md
lamont-granquist: I'm hoping that the implementation is at a deep enough layer in the objects that client code won't have to be updated everywhere.

new/cookbook-segment-deprecation.md
lamont-granquist: Yeah, I find writing motivation descriptions completely demotivating and useless...

new/cookbook-segment-deprecation.md
```diff
@@ -0,0 +1,55 @@
...
+
+This RFC will add a special `all_files` segment to cookbooks which will duplicate (initially)
+all of the existing cookbook manifest along with arbitrary files outside of the current scope of
+any of the cookbook segments.  The Chef Server will need to be patched in order to support this
+new 'segment' and accept it since the Chef Server currently does strict enforcement of file
+segments server-side.  This will also generate a version bump of the Chef Server protocol which
+will be used by servers to determine if the Chef Client supports this field or not.  Clients MAY
+implement code which determines the acceptable API version of the Server and then only send
+the new API requests to the server.  Clients MAY instead send requests which include both the
+old segment and new segment data.  The Server MUST detect the Client's desired API version and
```
lamont-granquist: So I'm fairly certain that what I just wrote is broken.  If clients send up a cookbook version with 'all_files' set in addition to the other segments then on an old server it'll just barf.  That suggests to me that the RFC should really state that there is no 'mixed mode' and that servers will always know what the client supports so should send that format, and that clients MUST figure out if they're talking to an old server or not (we have no generic mechanism for this, so I can't suggest an implementation here, but for cookbook uploads it is simple and the client uploads all the sandbox files first and can determine the API version from those responses).  Then once the client determines the server API version it should send the appropriate requests.

lamont-granquist: ^ that comment is out of date, but i'm still seeing it on my GH page even though I've pushed a complete rewrite.

btm:
:+1:

thommay:
Approved, @chef/rfc-editors do your thing

PR #162, opened 2015-10-23
Add additional native gems to omnibus-chef

lamont-granquist:
Adds the nokogiri gem, pg gem, and mysql gem (linked against mariadb, not Oracle mysql libs) to omnibus-chef.

FIXME: should read "to the chef project in omnibus-chef" everywhere.

### DIFF
--- a/new/omnibus-chef-native-gems.md
+++ b/new/omnibus-chef-native-gems.md
@@ -1,0 +1,49 @@
+---
+RFC: unassigned
+Author: Lamont Granquist <lamont@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Title
+
+Omnibus Chef Native Gem Additions
+
+## Motivation
+
+    As a Chef User,
+    I want to have omnibus-chef ship with nokogiri,
+    Because that will speed up my initial converges, and ease my pain at installing it.
+
+    As a Chef User,
+    I want to have omnibus-chef ship with popular Open Source databases,
+    Because that will speed up my initial converaes, and again ease my pain.
+
+    As a Postgres Cookbook Author,
+    I want to obliterate what we had to do to install Postgres client into omnibus from a cookbook,
+    Because, seriously, have you look at that code?
+
+
+## Specification
+
+The supporting native libraries:
+
+* libxml2
+* libxslt
+* libpq
+* MariaDB Connector/C
+
+And the following gems:
+
+* nokogiri
+* pg
+* mysql2
+
+Will be added to the omnibus-chef distribution on all Tier #1 systems.
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.
--- a/new/omnibus-chef-native-gems.md
+++ b/new/omnibus-chef-native-gems.md
@@ -17,11 +17,10 @@ Omnibus Chef Native Gem Additions
 
     As a Chef User,
     I want to have omnibus-chef ship with popular Open Source databases,
-    Because that will speed up my initial converaes, and again ease my pain.
+    Because that will speed up my initial converges, and again ease my pain.
 
     As a Postgres Cookbook Author,
-    I want to obliterate what we had to do to install Postgres client into omnibus from a cookbook,
-    Because, seriously, have you look at that code?
+    I need a better way to install native gems, as the required workarounds use unstable Ruby APIs and links to the omnibus packaged OpenSSL library, which makes the cookbook very complex and brittle.
 
 
 ## Specification


lamont-granquist:
We already ship nokogiri in ChefDK.  Back before ChefDK was released I was running the builds across the then-existing omnibus-chef build matrix and got it compiling on CentOS5/FreeBSD/Solaris and everything other than AIX.  Pretty confident that we can add and ship that one in chef-client.

Oh I just realized I'm saying omnibus-chef and I mean the chef project in omnibus-chef, sorry if that confuses anyone, will fix tomorrow...

I'd also like to ship the pg gem in omnibus-chef because this bit of code is horrible:

https://github.com/hw-cookbooks/postgresql/blob/develop/recipes/ruby.rb#L56-L116

That entire blob of code is because if you attempt to load the system libpg into omnibus-ruby you'll wind up loading the system openssl into a process that has already loaded the omnibus openssl which will cross the streams and destroy the world.  So all that code is beating on libpg to get it to link aginst the omnibus-chef openssl and use that instead.  

We also broke this in Chef 12.0 because of ruby 2.0 changes to the build process that required very deep ruby hackery to fix:

https://github.com/hw-cookbooks/postgresql/issues/212

I had unsubscribed to that issue because it was generating noise from people who just didn't seem to realize they needed compilers installed before installing a native gem.  I hadn't checked back into the thread until now, and its pretty depressing how many people stumble trying to use that cookbook.  I think that is highly compelling evidence that we should fix that problem for our users (I thought we should fix that problem before reading all the trailing posts in that thread, now I really think we should make that problem go away).

We ship postgres already for Ubuntu and CentOS.  If we can extend that to the rest of the chef project matrix that would seem to be very useful for our customers so that we can compile it for them instead of them having to deal with that themselves.

I included mysql2 mostly out of 'fairness' and large user base and large demand, I don't know that one has been as problematic, but I suspect there's pain out there from that one as well.

lamont-granquist:
It was suggested to me at the Summit that we should ship an 'addons' package to omnibus-chef instead of adding these directly to omnibus-chef.  I don't like that for two reasons:

1.  I think it would create an unfunded mandate that will not get done, while this proposal as written now has a much higher chance of seeing it through (consider that I'm suggesting that we compile postgres and mariadb client libraries for AIX, Solaris and Windows which is already new territory).

2.  I think it would create more confusion over what people should install and that more choice and options is not necessarily better.

If this road leads us down the path of having a 5GB omnibus-chef package at some point in the future then I think we should fix the problem of a bloated chef client package -- but fix it **after** it has become a problem, and not fix it before it has become a problem.

someara:
We should stick docker in there too

On Thu, Oct 22, 2015 at 10:47 PM, Lamont Granquist <notifications@github.com

ranjib:
i think this is an arbitrary list :-( . we can just pick a handful of things that we think is common and shove them in

lamont-granquist:
Does that get us into a situation where people will install omnibus-chef on CentOS6 and then run their apps using our docker instance, and then cut tickets to us when we upgrade and break their production, or don't upgrade fast enough and their production is stuck on buggy code?

someara:
I meant the docker-api rubygem, so I don't have to bundle it in the
cookbook.
The resources in there will be moved into core eventually (docker_container
and docker_image at least)

It has a dependency on excon, which causes weird conflicts with
chef-provisioning because it somehow loads the gem twice.

https://github.com/someara/chef-docker/issues/516

-s

On Thu, Oct 22, 2015 at 11:33 PM, Lamont Granquist <notifications@github.com

someara:
re: arbitrary list.. I don't think it's arbitrary... there is definitely a
short-list of gems cause a lot of pain.. nokogiri being one of them.

The number of commonly used community cookbooks that end up needing the
"compile time arms race for compilers and junk" is annoyingly high.

-s

ranjib:
docker-api gem would only make sense on linux right? that too above certain kernel versions.
it is arbitrary cause the definition of common, short list etc are derived from your perception, not from any data. 
i can add several more gems in that list (statsd e.g. ), but its hard to know if its common enough..without some survey. 
i would say cookbooks like systemd, sudo etc are common enough that their lwrps should go inside core. and they are not arbitrary as almost all linux platforms now use them, but i dont know how nokogiri or docker fits this ..and i dont see any data to back that :0(

lamont-granquist:
It is entirely non-scientific, but its not entirely arbitrary.  The measurement is the pain threshold for those cookbooks.  The evidence is things like that awful issue against the postgres cookbook and the awful code that we have to do to try to get that thing to install into omnibus.  And the nokogiri gem you can simply time how long it takes to compile which is annoying to start with, and then there's lots of common problems and bitching about compiling nokogiri, and the install on MacOS for one thing is still pretty terrible.

Does the statsd cookbook have a lot of code in it which works around compling native gems into linking against omnibus libraries like postgres does?  Does it have a lot of common support burden where people who are just trying to get their work done constantly stumble across it and fundamentally struggle to get their work done?  Does it have a compile time as long as nokogiri or libgecode to the point where users constantly complain about it?

And the common cookbooks problem is yet another orthogonal issue.  I'd agree a lot of common cookbooks totally should be brought into core, but that is entirely out of scope, and a bikeshed for another day (and could be solved largely by people simply doing the work and cutting PRs and getting the code into core--more doing and less talking here would be good).

Docker certainly wins on popularity.  It would create an interesting exception in that we can't ship it on Solaris/AIX/FreeBSD/etc, but I think that implication is obvious and nobody would expect us to port docker to AIX, so there would be limited confusion.

lamont-granquist:
It would also be useful to have someone who enjoys License issues more than I do to vet these ideas and make sure we can do them.  I omitted MySQL c-client libs because that issue is obvious.

someara:
Re: docker-api - actually no, that gem just speaks the Docker remote API
over https... it would allow chef-client on Windows, AIX, or any other
platform to remote control a Docker daemon over the network.

That being said, I think we should forget about it (for now) and have this
RFC focus on gems that require native extensions.

-s

lamont-granquist:
Ah okay so it would theoretically install everywhere then, even AIX...  Yeah if its not a native gem though and it installs fairly easy with pure ruby code, I think its best to focus on popular gems that have massively worse problems...

jonlives:
I'm generally speaking in favour of adding these gems, having personally experienced pain with both of them :D I like @lamont-granquist's metric of solving for problems causing users a) known and b) significant pain rather than adding every gem we might want too.

coderanger:
:+1: docker-api is pure Ruby AFAIK so probably not a big deal. Do we still ship the `json` gem? Do we want to care about it?

lamont-granquist:
json is built-in these days, but chef itself is json-clean and only uses ffi-yajl.

thommay:
https://tickets.opscode.com/browse/CHEF-4675, heh. Needless to say I think shipping nokogiri is hugely valuable. 
I think we need to make clear in this RFC that we're not going to be able to include some projects due to licenses - for example, we're not able to ship GPLv3 licensed projects in omnibus - and that's ok.

new/omnibus-chef-native-gems.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Motivation
+
+    As a Chef User,
+    I want to have omnibus-chef ship with nokogiri,
+    Because that will speed up my initial converges, and ease my pain at installing it.
+
+    As a Chef User,
+    I want to have omnibus-chef ship with popular Open Source databases,
+    Because that will speed up my initial converaes, and again ease my pain.
```
btm: *converges

btm:
It might be nice to frame this RFC as all/most the stuff we put in the omnibus builds that aren't dependencies. If we want to add something in the future, we propose an update to this RFC.

It'd be nice if that list became a table that listed rough size (cost) of the additional bits.

lamont-granquist:
@thommay re: CHEF-4675, lol...

I guess we could genericize this RFC a bit and turn it into more of a living document that we'd do PRs against in order to include more bits.  In that case including a clause about GPLv3 licenced code is a show stopper would be a good idea.

I also briefly considered including Berkshelf in this the other day but that means taking on the yak shave of libgecode everywhere, and starts to really blur the lines between the omnibus chef and chef-dk packages.  Throwing that out there, but even if we did that, I think that's something that should be included by a subsequent PR (agile, iterative development and all that jazz -- as it stands this is more than big enough to start with).

stevendanna:
> It might be nice to frame this RFC as all/most the stuff we put in the omnibus builds that aren't dependencies. If we want to add something in the future, we propose an update to this RFC.

I like this idea.  Our version manifest says:

```

File: new/omnibus-chef-native-gems.md [hunk 2]
```diff
@@ -0,0 +1,49 @@
...
+
+    As a Postgres Cookbook Author,
+    I want to obliterate what we had to do to install Postgres client into omnibus from a cookbook,
+    Because, seriously, have you look at that code?
```
btm: *looked, but snark. Unless you wrote all that code and you're being self-deprecating, I don't think we should be negative about it (and thus its authors), even if it was a product of the available Chef features at the time.

lamont-granquist: Yes, I helped write a lot of it, it is a confusing mess.  It is inherently awful to read and maintain and the hooks that it uses internally inside of ruby are not stable (broke in ruby-2.x that I had to fix) and the hooks that we used in the postgres build to force it to link against the omnibus openssl code is about as complex and brittle.  It isn't "bad code" in the sense that its not written by a "bad programmer", but it is not possible to clean up any better than it is, and requires specialized knowledge of the ruby library build system to understand and debug.  Making that code not need to exist WOULD make the world a better place, and we need to have a way to communicate that without people getting offended.

btm:
```
As a Postgres Cookbook Author,
I need a better way to install native gems, as the required workarounds use unstable Ruby APIs and links to the omnibus packaged OpenSSL library, which makes the cookbook very complex and brittle.
```

lamont-granquist:
So, I don't think we should have an RFC that includes the fact that we ship liblzma in chef-dk.  That seems overly fussy and better handled via some kind of actual chef-dk documentation, which this is not.

I'm also not certain that berkshelf and other necessary components of chef-dk should be listed.

thommay:
Approved, make it so!

PR #163, opened 2015-10-28
Update versioning for a CI/CD future (and present).

btm:
The internal CI system has become nearly a CD system since we last wrote this. Modifying the version scheme for build numbers will allow us to remove a lot of massaging the system to produce builds with specific version numbers at specific times.

The "Chef Client Release Process" section needs updates too, but imma scope this.

### DIFF
--- a/rfc047-release-process.md
+++ b/rfc047-release-process.md
@@ -28,7 +28,7 @@ so I can react accordingly.
 
 ### Versioning
 
-We aim to adhere to the Semantic Version scheme published [here](http://semver.org/). However, there are existing exceptions. Most notably, we do not consistently and clearly delineate which public methods are also part of the public API and guaranteed to only change with major version releases.
+We generally adhere to the Semantic Version scheme published [here](http://semver.org/), with noted exceptions.
 
 From Semantic Versioning:
 
@@ -44,40 +44,51 @@ Additional labels for pre-release and build metadata are available as extensions
 Chef Client specific clarifications:
 
 ```
-Given a version number MAJOR.MINOR.PATCH, increment the:
+Given a version number MAJOR.MINOR.BUILD:
 
-MAJOR version when the release will break existing functionality
-MINOR version when you add any new features
-PATCH version when you are only fixing bugs, typically regressions
+Increment the MAJOR version when the release will break existing functionality
+Increment the MINOR version when you add any new features
+The BUILD version is increased automatically by the build system
 ```
 
-##### Prerelease versions
+#### Continuous Integration / Continuous Delivery (build system)
 
-Any prerelease must be versioned in the format MAJOR.MINOR.PATCH.PRETYPE.PREVER, where:
+The BUILD version is numeric and must be increased for each build.
 
-PRETYPE is the description of the prerelease that the Releaser chooses based on the state of the prerelease, e.g. alpha, beta, rc, dev.
-PREVER is an incremented number starting at zero.
+The build system may consume unreleased BUILD versions. For example, a build which is successful but does not pass integration testing may not have its version reused.
 
-For example: 12.0.0.dev.0, 12.2.0.alpha.2, 12.3.0.rc.3
+It is left to the build system as to when to reset the BUILD version to zero, provided that newer versions will always compare greater than older versions.
+
+##### Channels
+
+The build system provides at least these repositories of builds, designated as channels:
+
+stable: Official releases
+current: Builds that has passed all automated testing
+
+#### Caveats
+
+We do not consistently and clearly delineate which public methods are also part of the public API and guaranteed to only change with major version releases. Public methods may change in minor releases based on the likelihood of that method being used by other projects or cookbooks.
 
 #### Examples
 
-Major: When changing the load order of any cookbook segments (e.g. attributes, templates), the major version number shall be incremented.
+MAJOR: When changing the load order of any cookbook segments (e.g. attributes, templates), the major version number shall be incremented.
 
-Minor: When adding support to the mount provider for special filesystem types that were previously unsupported, the minor version number shall be incremented.
+MINOR: When adding support to the mount provider for special filesystem types that were previously unsupported, the minor version number shall be incremented.
 
-Patch: When a minor release is made that contains a new feature to a service provider that causes the service provider to no longer work for users on an older but still supported version of the platform, the patch version number shall be implemented.
+BUILD: Automatically when a build is started in the build system. 
 
-Changes that require the minor or patch version number to be implemented may be included in a release that increments a higher version number without incrementing a lower version number. That is, new features that don not exist in version 1.1.0 may be released in 2.0.0 without any intermediary releases.
+Changes that require the MINOR version number to be increased may be included in a release that increases the MAJOR and allow the MINOR to be reset to zero. That is, new features that do not exist in version 1.1.0 may be released in 2.0.0 without any intermediary releases.
 
 ### Releasing
 
 #### Release Candidates
 
-Each major and minor release should have at least one week where a public RC is available
-for Chef users to test and report feedback.
+Official releases are made by promoting builds from the current channel to the stable channel. We no longer use the addition of a two-part alphanumeric suffix to describe prereleases. The stability of a build is now indicated by the release channel, e.g. current or stable, that the build is available from.
 
-Patch releases are expected to have only small changes which are extensively tested, and thus may be safe to release without a release candidate.
+An announcement should be made to the Chef mailing list at least three business days prior to the release of a build with an increase of either the MAJOR or MINOR versions over the last release. 
+
+Releases which do not increase MAJOR or MINOR versions are expected to have only bug/regression fixes which have been extensively tested, and thus may be released without prior notification.
 
 #### Chef Client Release Process
 
--- a/rfc047-release-process.md
+++ b/rfc047-release-process.md
@@ -51,20 +51,24 @@ Increment the MINOR version when you add any new features
 The BUILD version is increased automatically by the build system
 ```
 
+* When MINOR increases, BUILD must be reset to zero.
+* When MAJOR increases, MINOR and BUILD must be reset to zero.
+* Changes that require the MINOR version number to be increased may be included in a release that increases the MAJOR and allow the MINOR to be reset to zero. That is, new features that do not exist in version 1.1.0 may be released in 2.0.0 without any intermediary releases.
+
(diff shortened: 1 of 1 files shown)


rfc047-release-process.md
```diff
@@ -44,40 +44,51 @@ Additional labels for pre-release and build metadata are available as extensions
...
-Any prerelease must be versioned in the format MAJOR.MINOR.PATCH.PRETYPE.PREVER, where:
+##### Channels
 
-PRETYPE is the description of the prerelease that the Releaser chooses based on the state of the prerelease, e.g. alpha, beta, rc, dev.
-PREVER is an incremented number starting at zero.
+The build system provides at least these repositories of builds, designated as channels:
 
-For example: 12.0.0.dev.0, 12.2.0.alpha.2, 12.3.0.rc.3
+stable: Official releases
+current: Builds that has passed all automated testing
```
schisamo: I would make this something like:

```
* **current** - Builds that have based minimal automated testing. In most cases this means the appropriate package has been installed on each supported tier 1 platform and unit/functional tests have been executed.
* **stable** - Builds that have passed full acceptance testing.
```

rfc047-release-process.md
```diff
@@ -44,40 +44,51 @@ Additional labels for pre-release and build metadata are available as extensions
...
 ```
-Given a version number MAJOR.MINOR.PATCH, increment the:
+Given a version number MAJOR.MINOR.BUILD:
 
-MAJOR version when the release will break existing functionality
-MINOR version when you add any new features
-PATCH version when you are only fixing bugs, typically regressions
+Increment the MAJOR version when the release will break existing functionality
+Increment the MINOR version when you add any new features
+The BUILD version is increased automatically by the build system
```
schisamo: I know you touch on it briefly below but do we want to mention that:

* `MINOR` and `BUILD` are reset to zero when `MAJOR` is incremented.
* `BUILD` is reset to zero when `MINOR` is incremented.

btm: @jkeiser and I talked about that and figured we'd leave it up to the implementation. That's the obvious choice, but maybe it's easier to keep incrementing like Microsoft does for a while like:

```
12.0.123
12.0.124
12.1.125
12.1.126
12.2.127
13.0.128
```

I would guess it'd be not much difference in effort to implement it either way. I didn't have a compelling reason to specify that behavior besides the convenience of knowing.

rfc047-release-process.md
btm: I do want to be accurate, but I don't want to be overly specific. I could take out "all", i.e. "builds that have passed automated testing" if we believe there's more automated testing to be done. Are you implying that the promotion process from current to stable would trigger more automated testing? I thought it was just a copy/move operation.

Also, I do want to use the word "release" somewhere with stable. We're not ready to approach regular stable builds with little to no ceremony that show up automatically. For now, builds that make it to stable are our releases, and will continue to have ceremony like blog posts.

rfc047-release-process.md
```diff
@@ -44,40 +44,51 @@ Additional labels for pre-release and build metadata are available as extensions
...
 ```
 
-##### Prerelease versions
+#### Continuous Integration / Continuous Delivery (build system)
+
+The BUILD version is numeric and must be increased for each build.
+
+The build system may consume unreleased BUILD versions. For example, a build which is successful but does not pass integration testing may not have its version reused.
```
thommay: MUST not have its version reused, i'd say.

thommay:
I wonder if, in practice, we'll just wind up incrementing MINOR for virtually every build.

rfc047-release-process.md
btm: Yeah, there's maybe a line in there where you could reuse a build number for failed builds. But I guess we should just say they must all only be used one by the system that generates them, since it's going to generate them pretty close to when it's going to use them for something.

btm:
It would probably be the responsibility of the maintainer who merges to bump minor if necessary. Bug fixes wouldn't need it, which could be a lot of the merges. But yeah, we will end up consuming a lot of minor version unless we declare here that minor is meaningless outside of "stable." That is, when we're getting close to a release, we could bump minor, but the next build wouldn't have to be the one promoted to stable.

rfc047-release-process.md
ranjib: i like the 0 reset on minor version bump more.

ranjib:
looks awesome to me. this is good enough for starting, we can keep improving this, i bet as chef releases happen we are likely to need some changes in the release process.

:+1:

rfc047-release-process.md
lamont-granquist: It should probably be clarified that users should not expect to see patch dot-zero builds in the stable channel.  Even if we reset patch to zero when minor is bumped, stable users may not see a build until patch dot 33 or something.

btm:
Updated to clarify that we reset MINOR/BUILD to zero when we all already think we do.

martinb3:
It might be worth clarifying what constitutes `break existing functionality` -- I've seen a few chef-client releases where a bug or regression caused a feature to be re-written in an entirely novel implementation.

When there's a brand new implementation, it often comes with additional bugs or breaks behavior that worked in the old implementation, even if it's unrelated to the regressions being fixed (one example just from memory - https://github.com/chef/chef/issues/3604).

To be clear, I am not trying to dredge up old wounds for anyone --  I'd like to pre-empt the conversations that go something like: "well it wasn't ever supposed to work that way" or "that wasn't a public API to begin with."

I'd like to make it clear that a new release `MAY` or `MUST NOT` break behavior in existing cookbooks. For me, there is a distinction without a difference between a major release that breaks compatibility and bug fix / regression that breaks compatibility. Operationally, I'm using versions to judge the risk in upgrading.

stevendanna:
@btm 
>The "Chef Client Release Process" section needs updates too, but imma scope this.

It seems like these changes have pretty good support (I'm :+1: also).  I'm hoping we follow up on these changes soon though as it feels a bit odd to commit an RFC that is in a known-incorrect state.

@martinb3 
>  For me, there is a distinction without a difference between a major release that breaks compatibility and bug fix / regression that breaks compatibility.

I think that the difference is in how we will respond to such breakages.  A quick read of the bug you posted leads me to believe that we followed up by fixing that bug.  Because that was released as a minor version bump, I see us doing that bug fix as nearly mandatory [1].  If this work had been done as a major version bump, whether it was fixed would been more up for debate and not mandatory.  Also, it feels like there is a bit of chicken and egg problem there.  We don't ship with know cookbook-breaking-bug-fixes, so a developer is always likely to bump the "doesn't break compat" level since they wouldn't be shipping it if it did.

>  I'd like to pre-empt the conversations that go something like: "well it wasn't ever supposed to work that way" or "that wasn't a public API to begin with."

I understand this desire, but I am not sure we should do it as part of this PR.  We have a lot of work to do in terms of better defining the public API if that is something we want to do.  

From what I have seen in recent feature/bug discussions so far is that the core maintainers have had a pretty big focus on backward compatibility.  A number of the recent changes were actually much smaller than originally intended specifically to ensure that we remain backwards compatible.

We have had more cookbook-breaking bugs than we would like recently. Part of this is because @lamont-granquist, @jkeiser, and others are tackling some long-standing problems in Core chef where the current behavior was basically implementation-defined. Those breakages have led to ongoing efforts to improve cookbook & library compatibility testing. 

[1] I want to exclude cases like "I was monkeypatching this class method and now it is broken."

martinb3:
> I think that the difference is in how we will respond to such breakages.

While I agree that it was handled well, I don't think the version numbering is intended to communicate how we will respond. I am looking for version numbers to be a better indicator as to what was truly a low-risk bug fix of existing code vs. what was a more compatibility-risky re-write of existing code.

> We don't ship with know cookbook-breaking-bug-fixes, so a developer is always likely to bump the "doesn't break compat" level since they wouldn't be shipping it if it did.

I think there's probably some more nuance there -- I feel that we'll generally know if we completely re-wrote a feature (e.g. custom resources) vs. just corrected a conditional expression (e.g. fixing a single conditional expression in the provider map logic). To me, those map very nicely to major and minor version bumps and make it very clear to me what the risk in upgrading would be. But under the current & proposed schemes, the two cases above are both minor version bumps.

> We have had more cookbook-breaking bugs than we would like recently. Part of this is because @lamont-granquist, @jkeiser, and others are tackling some long-standing problems in Core chef where the current behavior was basically implementation-defined. Those breakages have led to ongoing efforts to improve cookbook & library compatibility testing. 

And to re-state this again, I'm very much in favor of this, and they've done a ton of valuable work that I don't want to disparage in any way. They really have done some amazing work that I use every day. I'd actually fully support bumping major _more_ often, in more of rolling release style, to free up those folks who are solving the hard problems. Perhaps I'm advocating for more of a rolling release, and just not calling it that?

> [1] I want to exclude cases like "I was monkeypatching this class method and now it is broken."

:+1: I agree completely, and I'm happy to put aside the public/private API discussions.

stevendanna:
@martinb3 Thanks for the thoughtful response.  Not gonna respond completely here but look forward to chatting in the IRC meeting.

> compatibility-risky re-write of existing code

I think the frequency of such large changes might be a bit over-represented in both of our memories because recent ones were very frequent.  If such rewrites became less frequent in the future, I think I would have little problem with bumping the major version.  My biases against bumping the major version frequently are mostly around user-education (if we think it is safe for someone to upgrade, I want them to upgrade as soon as possible and not be deterred by a major version bump) and logistics around how we will handle building packages, urgent security fixes, etc, with frequent major version bumps.  I'd think we can get there eventually, I just worry about the near term impact of such a policy.

btm:
If a rewrite of a feature is supposed to be backwards compatible, it should be a minor bump. We have to trust the intentions of the developer are good and they don't mean to screw anyone over. Any breakage that doesn't get fixed due to unintended uses is almost always the fault of:

https://github.com/btm/chef-rfc/blob/release/versioning/rfc047-release-process.md#caveats

> We do not consistently and clearly delineate which public methods are also part of the public API and guaranteed to only change with major version releases. Public methods may change in minor releases based on the likelihood of that method being used by other projects or cookbooks.

Which is a big bad elephant in the room, but also a big effort to correct. Someone needs to write the RFC regarding promises about public APIs in the code, then we start marking where we're going to enforce it.

btm:
This was approved by @adamedx as the decider in the [2015-11-12](https://github.com/chef/chef-community-irc-meetings/blob/master/2015/2015-11-12/2015-11-12-minutes.md) meeting.

PR #164, opened 2015-11-04
rfc-for-hugs

jjasghar:
There was discussion at the London summit 2015 on how to hug. Joking aside, this should be added as an RFC so our community acknowledges that this is a accepted policy.

/cc @martinisoft and @jjasghar

### DIFF
--- a/new/rfc-for-hugs.md
+++ b/new/rfc-for-hugs.md
@@ -17,12 +17,13 @@ In order to bridge the cultural gap between the US and Worldwide community the c
 
 ## Specification
 
-When approaching another Chef Community Member that you wish to engage in a hug. In order to comply with the already
-existing Chef Community Guidelines and Code of Conduct at various community events. You must seek permission from the
+When approaching another Chef Community Member that you wish to engage in a hug. To comply with the already
+existing Chef Community Guidelines and Code of Conduct at various community events, you must seek permission from the
 other community member by either verbally or physically indicating that you wish to hug them.
 
 An example, but not limited to, of a verbal indication for permission to hug is to ask "Can I give you a hug?" or "Can we hug?"
-The other party will then respond with verbal consent; for example "Yes, we can" or "Yes, you may."
+The other party should then respond with verbal statement; for example "Yes, we can" or "Yes, you may." is a positive reaction, whereas
+"No, Thank you." or "No, thanks" are considered negative reactions.
 
 An example, but not limited to, of a physical indication for permission to hug is spread your arms apart from your body. The other
 party can indicate permission by repeating the same physical expression to provide consent.
--- a/new/rfc-for-hugs.md
+++ b/new/rfc-for-hugs.md
@@ -22,21 +22,23 @@ existing Chef Community Guidelines and Code of Conduct at various community even
 other community member by either verbally or physically indicating that you wish to hug them.
 
 An example, but not limited to, of a verbal indication for permission to hug is to ask "Can I give you a hug?" or "Can we hug?"
-The other party will then respond with verbal consent; for example "Yes, we can" or "Yes, you may"
+The other party will then respond with verbal consent; for example "Yes, we can" or "Yes, you may."
 
 An example, but not limited to, of a physical indication for permission to hug is spread your arms apart from your body. The other
 party can indicate permission by repeating the same physical expression to provide consent.
 
 If Community member would like to express a hug, but has a limitation via a sickness or unable to use the typical hug process
-they may also express via other means, examples but not limited to:
+they may also express via other means, examples are but not limited to:
 - fist bumps
+- high fives
 - backward head nods by moving the chin upwards
 - elbows to elbows
 
-If a Community member would like to signal a rejection of a hug one of the possible physical signs one can do, but not limited to:
+If a Community member would like to signal a rejection of a hug one of the possible physical signs one can do, some examples are, but not limited to:
 - walking away
 - shaking ones head in a sign of "No"
-- crossing your arms over your chest placing your right hand on your left shoulder and left hand on your right shoulder could be a suggested way of indicating that you're not open to being hugged.
+- verbally expressing "No"
+- crossing your arms over your chest placing your right hand on your left shoulder and left hand on your right shoulder
 
 ## Copyright
 
--- a/new/rfc-for-hugs.md
+++ b/new/rfc-for-hugs.md
@@ -31,8 +31,10 @@ If Community member would like to express a hug, but has a limitation via a sick
 they may also express via other means, examples are but not limited to:
 - fist bumps
 - high fives
-- backward head nods by moving the chin upwards
-- elbow to elbow
+- in the case of an illness
+  - [elbow to elbow](https://en.wikipedia.org/wiki/Elbow_bump)
+  - backward head nods by moving the chin upwards
+
 
 A Community member may signal their desire not to participate in the hug. Such signals may include, but are not limited to:
 - walking away
--- a/new/rfc-for-hugs.md
+++ b/new/rfc-for-hugs.md
@@ -27,6 +27,17 @@ The other party will then respond with verbal consent; for example "Yes, we can"
 An example, but not limited to, of a physical indication for permission to hug is spread your arms apart from your body. The other
 party can indicate permission by repeating the same physical expression to provide consent.
 
+If Community member would like to express a hug, but has a limitation via a sickness or unable to use the typical hug process
+they may also express via other means, examples but not limited to:
+- fist bumps
+- backward head nods by moving the chin upwards
+- elbows to elbows
+
+If a Community member would like to signal a rejection of a hug one of the possible physical signs one can do, but not limited to:
+- walking away
+- shaking ones head in a sign of "No"
+- crossing your arms over your chest placing your right hand on your left shoulder and left hand on your right shoulder could be a suggested way of indicating that you're not open to being hugged.
+
 ## Copyright
 
 This work is in the public domain. In jurisdictions that do not allow for this,
--- a/new/rfc-for-hugs.md
+++ b/new/rfc-for-hugs.md
@@ -32,12 +32,12 @@ they may also express via other means, examples are but not limited to:
 - fist bumps
 - high fives
 - backward head nods by moving the chin upwards
-- elbows to elbows
+- elbow to elbow
 
-If a Community member would like to signal a rejection of a hug one of the possible physical signs one can do, some examples are, but not limited to:
+A Community member may signal their desire not to participate in the hug. Such signals may include, but are not limited to:
 - walking away
 - shaking ones head in a sign of "No"
-- verbally expressing "No"
+- verbally expressing "No, Thank you."
 - crossing your arms over your chest placing your right hand on your left shoulder and left hand on your right shoulder
 
 ## Copyright
--- a/new/rfc-for-hugs.md
+++ b/new/rfc-for-hugs.md
@@ -1,0 +1,35 @@
+---
+RFC: unassigned
+Author: Aaron Kalin and JJ Asghar
+Status: Draft
+Type: Informational, Process
+---
+
+# Hugging in the Chef Community
+
+In order to bridge the cultural gap between the US and Worldwide community the concept of hugging should be explained.
+
+## Motivation
+
+    As a Chef Community Participant,
+    I want to express positive emotion to or for another Chef Community Participant,
+    so that we can express our friendship physically.
+
+## Specification
+
+When approaching another Chef Community Member that you wish to engage in a hug. In order to comply with the already
+existing Chef Community Guidelines and Code of Conduct at various community events. You must seek permission from the
+other community member by either verbally or physically indicating that you wish to hug them.
+
+An example, but not limited to, of a verbal indication for permission to hug is to ask "Can I give you a hug?" or "Can we hug?"
+The other party will then respond with verbal consent; for example "Yes, we can" or "Yes, you may"
+
+An example, but not limited to, of a physical indication for permission to hug is spread your arms apart from your body. The other
+party can indicate permission by repeating the same physical expression to provide consent.
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.
--- a/new/rfc-for-hugs.md
+++ b/new/rfc-for-hugs.md
@@ -7,7 +7,8 @@ Type: Informational, Process
 
 # Hugging in the Chef Community
 
-In order to bridge the cultural gap between the US and Worldwide community the concept of hugging should be explained.
+In order to bridge the cultural gap between the US and Worldwide community the
+concept of hugging should be explained.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


nathenharvey:
should include some alternatives to hugging in case of possible communicable disease/illness ("conference cold" is a thing):

* fist bumps
* backward head nods
* elbows

nathenharvey:
crossing your arms over your chest placing your right hand on your left shoulder and left hand on your right shoulder could be a suggested way of indicating that you're not open to being hugged.

someara:
:+1: for fist bumps

martinisoft:
For context. At the Chef Community Summit in London today, @jonlives explained the process of a hug to the European audience who were not used to a hug. Shortly after, this was demonstrated to the audience by @jjasghar and @jonlives

scarolan:
How about little colored stickers for badges?
Red: Not a hugger, please respect my space
Yellow:  Ask permission and get a 'yes' first
Green:  hug me!

jjasghar:
@scarolan I like the idea of a "green" sticker, maybe a little green bear (bear hugs right :wink: ) or something that says you're open to hugs.

But that's probably as far as i think we should go. Putting the negative stickers seems exclusive, or requires our users to broadcast their desire not to participate, which goes against our community as a whole.

damacus:
:+1:  for outrageous fist bumps

scarolan:
@jjasghar I like it.  Yes, you are right, no red or yellow sticker is better

coderanger:
The WHO recommended protocol for physical greetings in an infection control situation (i.e. con flu) is the [elbow bump](https://en.wikipedia.org/wiki/Elbow_bump).

robbkidd:
:+1: On green bear hug stickers. Like the ambassador program Chef has at conference. Also like good security practice: default deny, then apply whitelist. Or in this case a greenlist.

Suggest a re-order of alternatives:

```
- fist bumps
- high fives
- in the case of illness
    - [elbow to elbow](https://en.wikipedia.org/wiki/Elbow_bump)
    - the "sup nod": a backward head nod by moving the chin upwards
```

tbunnyman:
:+1:  on the default deny and using green stickers for whitelisting.

munjeli:
While I like green bear, and none, there's maybe? one more level, which is about familiarity: Hug me if you know me, or Ask First. And then No Hugs.

someara:
http://truestorieswithgill.com/wp-content/uploads/2013/09/20130915-190532.jpg

tbunnyman:
I think that no matter what *nothing* you stick on a badge should be full a replacement for obtaining verbal consent for physical contact. I think having whatever simple marker should indicate the person is more willing than most to hug or whatever physical affection is chosen, just like the orange sticker. I'm not obligated to talk to everyone but people know I'm open to talk.

munjeli:
For sure. But categorizing the yellow and red as negative is perhaps not fair; people have different tolerances for contact especially with strangers and I think it's open to acknowledge that. It's also in line with other conferences I've been to. Unfortunately, safety is a concern for many people and giving more choices about how to participate is more inclusive. Last year I went to a conference with over 10,000 people where less than 2% were women; put yourself in those shoes and whether you want to open the door to random hugs, especially at drunken afterparties. Anyhow, I applaud the consideration.

tbunnyman:
I can understand red as a warning sign of "please stay away" but I worry about using marks on people (especially with a color like red) that make someone be intentionally or unintentionally avoided. I'm not trying to call anyone's preferences good or bad but wearing a sign that says "don't touch me" in a public place is going to invoke internal avoidance responses on various levels for different people. I thought relying on the Chef Code of Conduct as a baseline that everyone applies to and then giving out "totally yes I want to interact with you no matter who/what you are on X level" markers would have a more welcoming reply across the board.

I think this can be really loaded if we start turning it into a hanky code level of "yes || no || maybe || but only if" and especially if it starts to feel like a social pressure to participate or some way to create a baseline above or below the Code of Conduct. In that case I'd want to just stick to the Ambassador program and let people verbally negotiate affection at all times.

jaymzh:
Getting into how and when and why to get permission is... at best, hairy. I don't think it belongs here. I think what belongs here is the original intent: to define a hug, and that like all physical contact requires permission. The more you try to define it the more likely you are to end up either opening someone up to "having an excuse" or stifling everyone's behavior. I would say we should leave it at something simple like: "as with all physical contact, permission is required and unwanted physical contact is not OK. Please see the code of conduct for details on our community's standards."

munjeli:
Agreed that the baseline is define a hug, and if nothing more, than that's a win. But talking about code of conduct, the chef code of conduct is forked from the ada initiative and many conferences are doing a three tier yes || no || maybe for everything from 'talk to me' to 'take my photo' so it's pretty typical these days. Color red as scary, yes! Make it blue or whatever if that's a path. I don't think having this conversation is stifling. We should be having it. I thank you for having it. We should trust people to have it. Still, I favor simplicity, and the hug definition is cool.

jaymzh:
@munjeli - I have no objection to having the conversation. Sorry. If I gave that impression, it was a mistake. On the contrary, I think we should have more of these conversations in more venues with more people. I'm glad we're discussing it.

Perhaps I can clarify.

My first (and biggest) concern is re-codifying things that are already (hopefully?) clear in the COC/Community Guidelines (which @nathenharvey has put a ton of awesome effort into it along with, I'm sure, many others). I'd rather point to that as the guide. Otherwise there's conflicts, duplicate effort, more things to read, more things to miss, which leads to other problems. Unix philosophy: do one thing and do it well. A hug is a XYZ and requires permission. What is permission? COC/CG can tell you! What if we make another RFC that relates to a thing that requires permission - are we going to re-re-re define what permission is in that? If we decide we defined it wrong and we update some but not others, what then, does the model for permissions change for different things because not every place the defined how to ask for permission was updated? (we all know what we'd expect in that case, I'm not trying to pretend we don't, but that situation is much messier to deal with than the alternative)

My second concern is with going too far into codifying exact procedures. They run into problems. They don't always work for ... say different culture, location, or group of people. Or they allow people to follow the letter of the law without following the spirit of it. Specifics are great as examples, but not, IMO, as the actual code. I've been the enforcer/safe-person behind quite a few similar guidelines at many events and the ones that were by far the easiest to enforce were the ones that said "don't make our attendees uncomfortable" the ones that were overly specific ends up in semantic arguments... or someone coming along and going "this thing you say to do is awkward for everyone on both sides, wth" or "this thing we can't do is super common among the people here who are already friends!" Or whatever else. Because at the end of the day whether someone said "yes" or gave a headnod or sarcastically said "no" as long as they're comfortable, I think we're good... our goal (I think) is to prevent unwelcome contact/attention, not to enforce a specific set of procedures.

All that said - I'm never in the oppressed class on these sorts of things, so I'm sure there are pieces of this I don't see and don't understand. But I wanted to raise the things I _have_ seen being involved with other similar situations.

jaymzh:
(it occurs to me, @munjeli  you may not have be replying to me... if not, well.. er, at least I clarified what i was saying, heh)

jonlives:
While I agree that the CoC covers a lot of situations, I think that this specific case is worth dealing with separately...hugging is often socially acceptable, but also a relatively intimate form of physical contact, and I think we need to reinforce that hugging without specific consent from the recipient is a bad thing.

Regardless of the actual mechanism employed, I feel like we should perhaps err on the side of consent being explicitly necessary...essentially, let's not put the onus on people who *don't* want to be hugged to tell us that, let's put the onus on those who *do* want to be hugged to indicate that, verbally, via sticker, however.

jaymzh:
OK with some sleep and another read through comments on various other bits I missed:
* While I like "default deny" - I worry about the whole "everyone who doesn't have a sticker probably doesn't want a hug" will make a *drastic* change in the warm friendly huggy culture we have. That would be a shame. As @munjeli mentioned, there's probably 3 levels, not 2.
* @jonlives Yup, a definition of a hug should include that consent is required. Absolutely, no arguments. I just don't want it to be an RFC on the details of what constitutes asking or not asking for consent (which is where I'd defer to the CoC).
* @jjasghar Thanks for writing this. Talk about opening a prickly can o worms :)

jkeiser:
We need a section on failure handling--timeouts and exponential backoffs are always a good policy.

jonlives:
@jaymzh agreed that we don't want to nitpick on what exactly consent is as long as it's made explicit that it's necessary.

With that said, I feel that 'default deny' is absolutely necessary - you have to remember, to you and I a hug may be welcoming and friendly by default so to us, random hugs doesn't feel like such a big issue. To others, it may be unwelcome physical contact from (most likely) a male stranger.

I guess I'm saying I'm not advocating 'default deny' for people like me, I'm advocating it for people who come to tech events and might not feel safe / comfortable with random physical contact, and *especially* for people who might not feel comfortable speaking up about that.

new/rfc-for-hugs.md
```diff
@@ -0,0 +1,48 @@
...
+
+# Hugging in the Chef Community
+
+In order to bridge the cultural gap between the US and Worldwide community the concept of hugging should be explained.
+
+## Motivation
+
+    As a Chef Community Participant,
+    I want to express positive emotion to or for another Chef Community Participant,
+    so that we can express our friendship physically.
```
stevendanna: The "I want to" and "so that" lines seem to be a bit redundant here.  Also, I think it could do more to point out what is core to this RFC which is that we want to do these things *while respecting the boundaries of others*.

JillJubs:
1) Why haven't I been consulted about this? 
2) I would be reluctant to put too many constraints around hugging as it seems like counterintuitive to the 'hugging movement' itself. Obviously, the goal is never to make anyone uncomfortable, but to show a genuine emotion towards another human being. I worry that making too many rules around showing love in general will deter people from wanting to do it at all as they will always be worrying about doing it 'wrong' or getting in trouble for it!

munjeli:
@jaymzh I wasn't but thanks! 
Again I really appreciate this document which articulates the custom very well, including explicit permissions in several forms, and options. It's very brave to support the culture with a doc like this - I wholly support the idea of hugs in this setting and despise the modern tendency to make rules of defense  rather than rules of affection in the name of 'inclusion'. My interest in the conversation is about the wondrous model of friendship it represents and I feel that has a place in other projects and orgs. So I'm sold any which way it's practiced, ultimately. 

But to address the argument that multi-tier is too complex, I haven't really witnessed problems at other conferences with it. People either participate in color coded schemes or not, and those who don't make their own way. It hasn't been the topic of a lot of discussion in place or debugging the rules or anything. It's just another part of the party. Also people doing automated configuration are rather clever and often enjoy complexity.

kdaniels:
One thing that I think about in terms of whether there is a default-allow or default-deny is whether we are making people explicitly opt in or opt out of something. An issue that I've had in opt-out/default-allow is that it puts the onus on people to explain not only that they want to opt out, but also to explain _why_ they don't want to participate. Like @jonlives said, there can be a huge range of what feels "friendly" and "safe" to different people, and having a default-allow has, in my experiences, led to questions of "why not?" and "but you hugged so-and-so, why not me?" and tends to lead to discussions I don't want to have about why I don't want to hug a particular person.

I'd strongly lean towards a default-deny - I think having people feel more physically safe trumps the concern of people worrying that they might be hugging "wrong".

jkeiser:
Instead of having a default, why don't we just make super sure that at checkin for all conferences we present the color coding to people and tell them "if you're not sure yet, take the Yes or Maybe thingy and pin it up on yourself until you *are*. We don't want anyone to feel uncomfortable in this space!"

Default if you don't have a color on can safely be "hug," but everyone can still feel safe because we are super up front about the choice.

philoserf:
:+1:

jonlives:
@jkeiser I disagree with making the default if you're not wearing a colour as "hug" because even if you *do* make the choice super clear, it's still putting the onus on folks to indicate when they *don't* want something to happen to them. The priority here has to be the safety and comfort of community members.

jjasghar:
I'd like to thank everyone that has commented on this thread. Yep, you're right @jaymzh, this was a massive can o' worms, but oddly enough i think it has helped open some dialogs that our community has needed to have.

I think the scope of this RFC should be a basic explanation of a core action of our community. It's an agreed upon simple explanation of what a hug is, a way to acknowledge a hug, a way to deny a hug, and a way to still show the expression if someone is/was sick.  

Anything else probably should either be another RFC or have another forum for that discussion.

tbunnyman:
Good point. I'm :+1: on keeping the scope of this RFC as written so we can codify this and opening another RFC open for our further discussions the default/marker

robbkidd:
:+1: hug specification and basic `syn/syn+ack/ack` hug negotiation protocol.

miketheman:
(Joke comment) The RFC needs to address security and prevention of man-in-the-middle attacks on the hug protocol. This is commonly known as "Hug Hijacking".

jaymzh:
@miketheman Damnit, now I want to try that!!! (amongst people I know, of course)

jjasghar:
I think I'll let this sit for any more comments and prep it for the 2015-11-19 meeting. Please raise any objections sooner then later.

new/rfc-for-hugs.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Specification
+
+When approaching another Chef Community Member that you wish to engage in a hug. In order to comply with the already
+existing Chef Community Guidelines and Code of Conduct at various community events. You must seek permission from the
```
miketheman: The "In order" sentence does not make sense, it ends prematurely, or possibly should end in a comma before "You must".

new/rfc-for-hugs.md
```diff
@@ -0,0 +1,50 @@
...
+other community member by either verbally or physically indicating that you wish to hug them.
+
+An example, but not limited to, of a verbal indication for permission to hug is to ask "Can I give you a hug?" or "Can we hug?"
+The other party will then respond with verbal consent; for example "Yes, we can" or "Yes, you may."
```
miketheman: Needs a mention of negative reaction.

miketheman: Or at least "will" should be changed to "may" - "will" infers that the other party is always going to accept. Something I wish to be the case, but it not universally true.

jjasghar:
@miketheman updated per your suggestions.

ctdk:
Extremely minor point, but on the off chance the hug stickers end up being
red/yellow/green objects, could the red and green ones be differentiated a
little more than merely being red bears and green bears (or worse, small
red and green dots). There are those of us who don't do so well with that.

-j

gwaldo:
I consider the sticker / badge / whatever as a sign of your defaults a fine idea.  But consent should always be sought.

I would rather see the badge indication in the affirmative to mean "I'm open to be solicited for hugs", and the negative to mean "I would prefer to initiate", rather than "hug away" vs "back off".  In my mind (a silly place, to be sure), it seems that this might save the subtle rejection felt when arms are spread and met with a shaken head.

While not color-blind myself, I've received feedback that to those with red/green color blindness, a simple colored dot will not be useful at all.  :white_check_mark: vs :x: might be more apparent/useful.

jjasghar:
@ctdk, @gwaldo, completely agreed. The idea of a sticker is a great one. But the scope of this RFC is just to describe the act of hugging an nothing more. It seems like i should spin up another RFC, sooner then later, on the stickers/notifications etc. I have a gut feeling that there will be a large discussion around it.

new/rfc-for-hugs.md
```diff
@@ -0,0 +1,51 @@
...
+
+    As a Chef Community Participant,
+    I want to express positive emotion to or for another Chef Community Participant,
+    so that we can express our friendship physically.
+
+## Specification
+
+When approaching another Chef Community Member that you wish to engage in a hug. To comply with the already
+existing Chef Community Guidelines and Code of Conduct at various community events, you must seek permission from the
+other community member by either verbally or physically indicating that you wish to hug them.
```
btm: 1) The first sentence ends abruptly. Editing error?

2) We should cite [RFC020](https://github.com/chef/chef-rfc/blob/master/rfc020-community-guidelines.md) and the chefconf code of conduct. maybe the latter as an example, because code of conducts vary between events.

3) The tone of "you must seek permission" combined with "to comply [with the rules]" doesn't feel right to me. 

For 1+2+3, I'll suggest:

Foremost, we want all community members to feel safe and welcome. This is expressed in the Chef Community Guidelines ([RFC020](https://github.com/chef/chef-rfc/blob/master/rfc020-community-guidelines.md)) and the Code of Conduct for our events (e.g. [ChefConf](https://www.chef.io/blog/chefconfcodeofconduct/)). When approaching another Chef Community Member that you wish to engage in a hug, you must first seek permission from the other community member by either verbally or physically indicating that you wish to hug them.

adamleff:
This was approved by @thommay in the 11/19/2015 meeting pending tidying-up by @jjasghar, which was completed in 60c388c.  I will accept and merge this. :)

PR #165, opened 2015-11-11
Add another 'open to hugs' suggestion

martinb3:
Suggest we could do badge stickers for being 'open to hugs' like we had at the last ChefConf for people who are open to discussing specific topics.

### DIFF
--- a/new/rfc-for-hugs.md
+++ b/new/rfc-for-hugs.md
@@ -32,6 +32,7 @@ If Community member would like to express a hug, but has a limitation via a sick
 they may also express via other means, examples are but not limited to:
 - fist bumps
 - high fives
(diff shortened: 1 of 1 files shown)


jjasghar:
Until the sicker RFC or that decision is made, I don't think I want to put it in hugging RFC. It'll be a great addendum, but at the time being we should take note of this, as something we can work toward.

martinb3:
@jjasghar What's the sticker RFC? :)

jjasghar:
Per the https://github.com/chef/chef-rfc/pull/164 we decided to only scope the RFC as just the *act* of hugging. The sticker RFC which still needs to be written will cover the badge stickers and the like.

martinb3:
Ah! :+1: then. I'm behind the times :eyes:

PR #166, opened 2015-11-12
The mailing list migration is complete

nathenharvey:


coderanger:
In theory final is only used for standards track but seems fine either way, what say the other @chef/rfc-editors?

nathenharvey:
Is there something other than "Final" such as "Completed" or "Implemented"?  

"Accepted" feels to me like the idea has been accepted but I don't know how to indicate that the work has been completed.

thommay:
Final feels a bit wrong for an informational RFC but I'm not sure it's worth the bikeshed of coming up with something else :)

adamleff:
I think "Final" doesn't appropriate represent what we're trying to convey, but I'm with @thommay that a bikeshed on what it SHOULD be might not be worthwhile :)

I'd like to propose we simply add a "Note: The mailing list migration was completed on Month Day, Year." right after the metadata table and leave it as Accepted.

nathenharvey:
great idea, @adamleff.  I've reset the status and added a note about the cutover date.

adamleff:
:+1:

btm:
:+1:

btm:
This was approved by @adamedx as the decider in the [2015-11-12](https://github.com/chef/chef-community-irc-meetings/blob/master/2015/2015-11-12/2015-11-12-minutes.md) meeting.

PR #168, opened 2015-11-30
Chef code should have consistent style

thommay:
Please let's not discuss the actual checks we want to disable here,
but rather let's just discuss having such a thing *at all* and then use the chef/chefstyle repo to define a set of checks that we want.

cc @chef/engineering-services

### DIFF
--- a/new/chefstyle.md
+++ b/new/chefstyle.md
@@ -24,7 +24,7 @@ for its open source ruby code and apply it.
 
 https://github.com/chef/chefstyle provides a rubocop configuration that
 is intended to be applied to chef/chef and related projects, such as
-chef/ohai, and chef/mixlib-*. The community should arrive, via
+chef/ohai, and chef/mixlib-\*. The community should arrive, via
 pull requests to `chefstyle`, at a set of style checks that can be
 enforced (via travis, etc) on chef community maintained ruby projects.
 
--- a/new/chefstyle.md
+++ b/new/chefstyle.md
@@ -1,0 +1,40 @@
+---
+RFC: unassigned
+Author: Thom May <thom@chef.io>
+Status: Draft
+Type: Standards Track
+Tracking:
+ - https://github.com/chef/chefstyle
+---
+
+# Chef code should have a consistent style
+
+There is a high cognitive load involved in switching between projects
+that have different or inconsistent code styles that impacts developer
+productivity and happiness. Chef should produce a consistent code style
+for its open source ruby code and apply it.
+
+## Motivation
+
+    As a developer,
+    I want to be productive quickly in all Chef's projects,
+    so that I can concentrate on writing code.
+
+## Specification
+
+https://github.com/chef/chefstyle provides a rubocop configuration that
+can be applied to any ruby projects. The community should arrive, via
+pull requests to `chefstyle`, at a set of style checks that can be
+enforced (via travis, etc) on any community maintained ruby projects.
+Chefstyle, by default, disables all rubocop cops, providing a clean
+sheet for us to build upon.
+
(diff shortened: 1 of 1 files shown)


jaymzh:
OMG yes.

:+1: 
:+1: 
:100:

patrick-wright:
Great idea.

coderanger:
Not sure how implicit we want to leave this, but some Chef-owned repos should probably be excepted. Like Supermarket, Rails code generally has very very different style than anything else.

lamont-granquist:
Not a repo.   Like chef-config have another gem we ship out of chef/chef and is tied to the chef version.

thommay:
@lamont-granquist I'm not sure I like that. We should release and version chefstyle appropriately but I think tying it to the chef repo just adds noise to chef PRs and is slightly more awkward to work with in other repos.

lamont-granquist:
And we really need to be clear here that this is going to be a gem which will target chef/chef and pretty much only chef/chef (and maybe the related mixlib's) and will not be architected to the needs of cookbook authors at all, and is not intended as a general purpose ruby style guide for the chef community at large.

That is because some issues only come up in chef/chef (like rescue Exception being perfectly acceptable if you're decorating the exception and re-raising it), and because some issues may be large yak shaves to fix correctly in the chef/chef codebase (thinking about the `File.exists?` to `File.exist?` PR that got lost somewhere in the bowels of FileAccessControl and died)

lamont-granquist:
Think I just answered your ninja post there.  We need to target chef/chef's needs specifically, and that will be the really hard part, so I think it lives in the chef/chef repo.

thommay:
@coderanger i kinda felt that rails projects could be additive over the top, but I'm easy.

thommay:
well, but it should be targetting chef, chef-zero, ohai, cheffish, mixlib-* at least.

coderanger:
It sounds like we need to first decide if this is for chef/chef only, most Chef Software projects, or "the community". If we end up with 2 we can talk about which projects to use it on.

lamont-granquist:
I am 1,000,000% against having a single ruleset for everyone plus chef/chef, and I'm positive that way will end up in endless bikesheds, yak shaves and tears.

damacus:
👍 for consistency across repos.

Makes it much easier to teach new comers if styling is consistent with
"current"  practise (whatever we chose at the time).

thommay:
I made some changes to try and clarify @lamont-granquist's concerns and ensure that rules are added in  a pretty conservative manner.

lamont-granquist:
I think we need to tackle chef/chef first and that will be the hard problem.  I think ohai and the mixlib's come along for the ride.  I doubt we'll hit anything in those repos that we don't hit in core chef.

coderanger:
Using chef/chef to figure out which cops to enable seems like a reasonable plan, but then expect to enable it for other repos after that "training wheels" stage?

lamont-granquist:
Yeah once we've applied stuff to chef/chef then it should be straightforward to apply it to all the other repos.

schisamo:
Do we envision chefstyle supporting multiple rulesets in the future? This PR is obviously discussing the Ruby project-specific ruleset. We could also ship a ruleset aimed at cookbooks and cookbook authors. Or is the intention to have a completely different gem?

lamont-granquist:
I'd really like to have a completely separate gem to keep the bikesheds segregated.

thommay:
@schisamo completely separate gem

btm:
:+1: congrats on pre-bikeshed foundation-shedding.

adamhjk:
I am the decider, and I approve this rfc. Go for it, @chef/rfc-editors

PR #171, opened 2015-12-07
Chef, upgrade thyself

thommay:
Specify a system, inbuilt to chef, to cause the chef client to upgrade itself at appropriate times.

### DIFF
--- a/new/chef-upgrade.md
+++ b/new/chef-upgrade.md
@@ -1,0 +1,70 @@
+---
+RFC: unassigned
+Author: Thom May <thom@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Chef should be able to upgrade itself
+
+Users should be able to automatically upgrade to the desired version of chef without having to use third party cookbooks. The upgrade should occur without affecting the normal chef client run, ensuring that a consistent state is preserved.
+
+## Motivation
+
+```
+  As an operator, I want chef versions to be consistent across my infrastructure,
+  so that I can ensure my code works correctly
+```
+```
+  As an operator, I want to easily and safely upgrade chef to my desired version,
+  so that I can easily roll out new versions.
+```
+```
+  As a developer, I want our users to get new features and bug fixes in a timely and safe manner,
+  so that everyone's happy.
+```
+
+## Specification
+
+At the end of each chef run, just before the reboot handler, the chef client should check with an update service to discern whether there is a new version of chef. If there is, the client should decide whether or not to upgrade, based on configuration supplied by the administrator.
+Should the decision to upgrade be made, the client will download the new version, and install it appropriately. If a reboot is not triggered, the client run should exit using an appropriate exit code to signal an upgrade.
+
+### Determining the desired chef version
+
+The desired chef version should be provided to the node through a new first class environment or node attribute, named `chef_version`. Once RFC 45 is completed, that attribute will be marked as desired state, and not mutable by the node. The `chef_version` attribute must contain a SemVer parseable version string. A version always takes the form x.y.z, where x, y, and z are decimal numbers that are used to represent major (x), minor (y), and patch (z) versions. One-part (x) and two-part (x.y) versions are allowed.
+
+Chef will also add a `chef_upgrade` resource to help administrators control the upgrade process, for example by selectively enabling upgrades based on search results or custom attributes. The `chef_upgrade` resource will expose a `version` attribute, allowing one to override the `chef_version` attribute. It will expose one action, `upgrade`.
+
+
+### Checking for an upgrade
+
+Currently, Chef provides an omnitruck API service that allows one to query for versions of packages. https://docs.chef.io/api_omnitruck.html describes the API, which provides the ability to request various groups of versions. In future it may be desirable to provide a similar API in the Chef Server, to provide upgrades to nodes that have no external connectivity.
+
+### Installing the upgrade
+
+The chef client will use the most appropriate mechanism to install the latest version of chef. The chef client will then exit with an appropriate error code, allowing the parent process to restart gracefully.
+
+### Example scenario
+
+Given the environment below:
+```json
+{
+  "name": "production",
+  "chef_version": "12.5"
+}
+```
+A chef client on an ubuntu node in the `production` environment, will make a request to omnitruck:
+```
+http://www.chef.io/chef/metadata?p=ubuntu&pv=14.04&m=x86_64&v=12.5
+```
+and receive metadata containing the latest package in the `12.5` series. 
+Should that package be newer than the current version, chef will download
+the deb package, install it via `dpkg`, and end the chef run with a suitable
+exit code.
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.
--- a/new/chef-upgrade.md
+++ b/new/chef-upgrade.md
@@ -32,12 +32,12 @@ If the administrator has specified a version of chef different to the one curren
 
 ### Determining the desired chef version
 
-The desired chef version should be provided to the node through a new first class environment or node attribute tree, named `__chef_version__`. Once RFC 45 is completed, that attribute tree will be marked as desired state, and not mutable by the node.
+The desired chef version should be provided to the node through a new first class environment or node attribute tree, named `__chef_upgrade__`. Once RFC 45 is completed, that attribute tree will be marked as desired state, and not mutable by the node.
 
 The attribute tree will contain one mandatory attribute, `version`, and
 some additional attributes. Currently, we intend to support `channel`
 to allow the administrator to specify that they would like to consume
-unreleased builds. Channels are documented in [RFC 47](https://github.com/chef/chef-rfc/blob/master/rfc047-release-process.md#channels)
+unreleased builds. Channels are documented in [RFC 47](https://github.com/chef/chef-rfc/blob/master/rfc047-release-process.md#channels).
 
 The `version` attribute shall contain either MAJOR, MAJOR.MINOR, or
 MAJOR.MINOR.BUILD as documented in [RFC 47](https://github.com/chef/chef-rfc/blob/master/rfc047-release-process.md#versioning).
@@ -64,7 +64,7 @@ Given the environment below:
 ```json
 {
   "name": "production",
-  "__chef_version__": {
+  "__chef_upgrade__": {
     "version": 12.5"
   }
 }
--- a/new/chef-upgrade.md
+++ b/new/chef-upgrade.md
@@ -26,23 +26,37 @@ Users should be able to automatically upgrade to the desired version of chef wit
 
 ## Specification
 
-At the end of each chef run, just before the reboot handler, the chef client should check with an update service to discern whether there is a new version of chef. If there is, the client should decide whether or not to upgrade, based on configuration supplied by the administrator.
-Should the decision to upgrade be made, the client will download the new version, and install it appropriately. If a reboot is not triggered, the client run should exit using an appropriate exit code to signal an upgrade.
+At the beginning of each chef run, the client should decide whether or not to upgrade, based on configuration supplied by the administrator.
+
+If the administrator has specified a version of chef different to the one currently running, the chef client should check with an update service to ensure the specified version is available. If so, the client will download the specified version, and install it appropriately. The client run should then exit, using an appropriate exit code to signal an upgrade.
 
 ### Determining the desired chef version
 
-The desired chef version should be provided to the node through a new first class environment or node attribute, named `chef_version`. Once RFC 45 is completed, that attribute will be marked as desired state, and not mutable by the node. The `chef_version` attribute must contain a SemVer parseable version string. A version always takes the form x.y.z, where x, y, and z are decimal numbers that are used to represent major (x), minor (y), and patch (z) versions. One-part (x) and two-part (x.y) versions are allowed.
+The desired chef version should be provided to the node through a new first class environment or node attribute tree, named `__chef_version__`. Once RFC 45 is completed, that attribute tree will be marked as desired state, and not mutable by the node.
 
-Chef will also add a `chef_upgrade` resource to help administrators control the upgrade process, for example by selectively enabling upgrades based on search results or custom attributes. The `chef_upgrade` resource will expose a `version` attribute, allowing one to override the `chef_version` attribute. It will expose one action, `upgrade`.
+The attribute tree will contain one mandatory attribute, `version`, and
+some additional attributes. Currently, we intend to support `channel`
+to allow the administrator to specify that they would like to consume
+unreleased builds. Channels are documented in [RFC 47](https://github.com/chef/chef-rfc/blob/master/rfc047-release-process.md#channels)
 
+The `version` attribute shall contain either MAJOR, MAJOR.MINOR, or
+MAJOR.MINOR.BUILD as documented in [RFC 47](https://github.com/chef/chef-rfc/blob/master/rfc047-release-process.md#versioning).
+It can also be the special string "latest", signifying that the node
+should always upgrade to the latest available version.
 
 ### Checking for an upgrade
 
-Currently, Chef provides an omnitruck API service that allows one to query for versions of packages. https://docs.chef.io/api_omnitruck.html describes the API, which provides the ability to request various groups of versions. In future it may be desirable to provide a similar API in the Chef Server, to provide upgrades to nodes that have no external connectivity.
+Currently, Chef provides an omnitruck API service that allows one to query
+for versions of packages. The API is [documented](https://docs.chef.io/api_omnitruck.html),
+and provides the ability to request various groups of versions. In
+future it may be desirable to provide a similar API in the Chef
+Server, to provide upgrades to nodes that have no external connectivity.
 
 ### Installing the upgrade
 
-The chef client will use the most appropriate mechanism to install the latest version of chef. The chef client will then exit with an appropriate error code, allowing the parent process to restart gracefully.
+The chef client will use the most appropriate mechanism to install the
+latest version of chef. The chef client will then exit with an appropriate
+error code, allowing the parent process to restart gracefully.
 
 ### Example scenario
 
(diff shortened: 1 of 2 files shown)


new/chef-upgrade.md
```diff
@@ -0,0 +1,70 @@
...
+Should the decision to upgrade be made, the client will download the new version, and install it appropriately. If a reboot is not triggered, the client run should exit using an appropriate exit code to signal an upgrade.
+
+### Determining the desired chef version
+
+The desired chef version should be provided to the node through a new first class environment or node attribute, named `chef_version`. Once RFC 45 is completed, that attribute will be marked as desired state, and not mutable by the node. The `chef_version` attribute must contain a SemVer parseable version string. A version always takes the form x.y.z, where x, y, and z are decimal numbers that are used to represent major (x), minor (y), and patch (z) versions. One-part (x) and two-part (x.y) versions are allowed.
+
+Chef will also add a `chef_upgrade` resource to help administrators control the upgrade process, for example by selectively enabling upgrades based on search results or custom attributes. The `chef_upgrade` resource will expose a `version` attribute, allowing one to override the `chef_version` attribute. It will expose one action, `upgrade`.
```
sersut: In `mixlib-install`, we have been using `:channel` in addition to `:version` to determine what version of Chef to install. It can take any value from `:stable`, `:current` or `:unstable`.

new/chef-upgrade.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+## Specification
+
+At the end of each chef run, just before the reboot handler, the chef client should check with an update service to discern whether there is a new version of chef. If there is, the client should decide whether or not to upgrade, based on configuration supplied by the administrator.
+Should the decision to upgrade be made, the client will download the new version, and install it appropriately. If a reboot is not triggered, the client run should exit using an appropriate exit code to signal an upgrade.
+
+### Determining the desired chef version
+
+The desired chef version should be provided to the node through a new first class environment or node attribute, named `chef_version`. Once RFC 45 is completed, that attribute will be marked as desired state, and not mutable by the node. The `chef_version` attribute must contain a SemVer parseable version string. A version always takes the form x.y.z, where x, y, and z are decimal numbers that are used to represent major (x), minor (y), and patch (z) versions. One-part (x) and two-part (x.y) versions are allowed.
```
lamont-granquist: I still like having the `chef_upgrade` resource better and letting users drive it with whatever resource they like.

lamont-granquist:
We should bake into this an option that wipes /opt/chef on upgrade and defaults to true.  The sooner we start doing this the sooner we can rely on this as an answer to issues with chef_gems left around from the prior install.

new/chef-upgrade.md
```diff
@@ -0,0 +1,70 @@
...
+  so that I can easily roll out new versions.
+```
+```
+  As a developer, I want our users to get new features and bug fixes in a timely and safe manner,
+  so that everyone's happy.
+```
+
+## Specification
+
+At the end of each chef run, just before the reboot handler, the chef client should check with an update service to discern whether there is a new version of chef. If there is, the client should decide whether or not to upgrade, based on configuration supplied by the administrator.
```
coderanger: Having it at the end of the run seems odd. If I was setting is to X I would expect it to upgrade to that before running whatever other changes were part of that. This means you have to ensure everything runs at least once before using new syntax etc rather than being able to match it directly.

new/chef-upgrade.md
coderanger: :+1: as a low-level impl.

new/chef-upgrade.md
```diff
@@ -0,0 +1,70 @@
...
+
+Chef will also add a `chef_upgrade` resource to help administrators control the upgrade process, for example by selectively enabling upgrades based on search results or custom attributes. The `chef_upgrade` resource will expose a `version` attribute, allowing one to override the `chef_version` attribute. It will expose one action, `upgrade`.
+
+
+### Checking for an upgrade
+
+Currently, Chef provides an omnitruck API service that allows one to query for versions of packages. https://docs.chef.io/api_omnitruck.html describes the API, which provides the ability to request various groups of versions. In future it may be desirable to provide a similar API in the Chef Server, to provide upgrades to nodes that have no external connectivity.
```
jtimberman: If we're going to integrate this tightly with omnitruck, I think for the benefit of our enterprise-type customers, we should ensure there's a way for this to run on premises. I'm fairly certain people have discussed having some kind of artifact service endpoint in the Chef Server before, but I can't find it right now.

new/chef-upgrade.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Installing the upgrade
+
+The chef client will use the most appropriate mechanism to install the latest version of chef. The chef client will then exit with an appropriate error code, allowing the parent process to restart gracefully.
```
jtimberman: Should this be explicit that it will properly tie in with reporting/error handlers, Chef Reporting, so as to not indicate a failed run, but instead an upgrade occurred?

new/chef-upgrade.md
```diff
@@ -0,0 +1,70 @@
...
+}
+```
+A chef client on an ubuntu node in the `production` environment, will make a request to omnitruck:
+```
+http://www.chef.io/chef/metadata?p=ubuntu&pv=14.04&m=x86_64&v=12.5
+```
+and receive metadata containing the latest package in the `12.5` series. 
+Should that package be newer than the current version, chef will download
+the deb package, install it via `dpkg`, and end the chef run with a suitable
+exit code.
```
jtimberman: How do we account for users that have configured their systems to use the apt/yum repositories instead of using omnitruck?

new/chef-upgrade.md
jtimberman: I've gone back and forth on this over the years, and I think having it at the end is the right thing because we don't have to do weird things with regard to the run state or trying to kill the chef run and catch the next process.

new/chef-upgrade.md
thommay: yup, that's a good idea.

new/chef-upgrade.md
thommay: @coderanger I originally wrote the RFC so that the upgrade came first - before any part of the chef run. But I've come round to the idea that it's worth doing after, so that the user can provide configuration to the upgrade as part of the chef run. And now that we can run stuff as a handler, we can do it after a complete chef run.
I'm not convinced the timing will be a huge problem; it's likely that the chef upgrade will happen some time before any new feature is used.

new/chef-upgrade.md
thommay: It'll run after the report handlers, and it won't fire if there's any failures. So I don't think we need to be more explicit than this, and lean on the work in [RFC 62](https://github.com/chef/chef-rfc/blob/master/rfc062-exit-status.md).

new/chef-upgrade.md
coderanger: @thommay I'm thinking of the case of an old VM coming back online after being down for a while, which I guess would be no worse than it is today at least. Having it as a resource to allow for "no really do it nao and then restart" would be a decent middle ground I think though. For most people it can run at the end, but if you need to do complex nonsense then it is your own problem. Combined with the new chef_version restrictions on cookbooks and I think the UX would be solid.

tas50: Having the upgrade occur at the beginning of the chef run is key for many users.  Here's the two scenarios I've experienced:

1) System images / AMIs:  You roll out a AMI with Chef 12.1 and you bootstrap outside of knife, as many people do. You want to be able to spin up a new host, have chef upgrade to the version your code requires (12.5 maybe for custom resources), and then have the node converge.  Having the upgrade occur at the end would require you to handle the upgrade outside chef and defeats the purpose of this feature.

2) You want to roll out new chef functionality without first rolling out chef and your environment is not entirely convergent all the time. Again back to the 12.5 custom resource example. You want to start using custom_resources and you're not running 12.5 everywhere.  You shouldn't need to upgrade your fleet first then roll out the custom resources. This becomes really key when you consider that in hardware environments a large number of systems are down due to failures at any time.  If you roll out 12.5 and then bump your cookbooks any node that comes back up will fail.  At a previous job our failure rate hovered from between 2-5% due to node/rack/DC failures.  This was 500 - 1250 nodes at our scale so manually resolving chef run failures wouldn't be an option.

lamont-granquist: So that eliminates the ability to control this with any kind of resource.  Do we expect to be able to tweak this with attribute files in cookbooks?  Because that implies it runs after cookbook synchronization, library compilation and attribute compilation but before LWRPs/resources.

new/chef-upgrade.md
jtimberman: :+1: "Mark as resolved"

new/chef-upgrade.md
btm: I don't think we should specify `node[:chef_version]`. The `chef_upgrade` resource should take a number of properties including version and the user could set that to whatever they want. That could be a fixed string, `node[:chef_version]` or whatever.

You should call out the properties after the next paragraph, e.g. version, channel. When you specify version, you should refer to RFC047#versioning rather than duplicating the explanations here.

new/chef-upgrade.md
btm: :raccoon: but could we just ship omnitruck with chef server and have it use bookshelf (or future replacement)?

new/chef-upgrade.md
tas50: I guess I'm not thinking of the implementation, but how people would use this.  I spent an enormous amount of my time upgrading chef clients across the world.  A datacenter goes down for a week and you roll out new chef version + cookbooks.  The old DC comes back up and you now have to manually resolve failed runs.  It's a huge pain.  Any talk of upgrades has to take into account that no large environment is 100% functional at any time.  That's just the nature of large systems and it has to direct how we build auto-upgrade functionality.

ranjib: chef client should only dial home if the administrator has configured it to do so. otherwise chef-client should not even hit the omnitruck url.

coderanger: This entire system should be disabled by default until a user requests (via node attr or otherwise) a specific version be deployed.

new/chef-upgrade.md
thommay: Have a `strategy` or `provider` or something attribute on the `chef_upgrade` resource and use that to signal?

new/chef-upgrade.md
thommay: @coderanger / @ranjib yup, agreed.

new/chef-upgrade.md
adamedx: I think the update source should not be tied to the implementation. We should allow the source to configurable using client-side logic, possibly a chef config, so that in the short term users can configure the location and avoid Internet access (People already do things like this today). The rest of the update logic in this RFC would kick in, the config would only affect the location from which the client is downloaded.

Longer term as @btm and @jtimberman state we could have omnitruck and content hosting directly integrated with the chef server infrastructure.

lamont-granquist: yep, we've been talking about doing that for at least 3 years now...

but, whatever the implementation, it should support flags like bootstrap does to override the url, override the flags, or override the entire command.

new/chef-upgrade.md
lamont-granquist: prefer to have `omnitruck_chef_upgrade` and `packagecloud_chef_upgrade` resources and providers with provides lines hooking up `chef_upgrade` to `omnitruck_chef_upgrade` by default.  that is a way better pattern and gets the validation correct, rather than buidling a massive single fascade resource.

adamedx:
I think I'd like to see information about ongoing support of the upgrade protocol. If we can upgrade from arbitrary versions (starting from the first version of chef-client that supports this), then the upgrade protocol in this RFC needs to be supported on all future revisions of chef client and server. I think there should be something explicit in the RFC that implementers are expected to support this forever...

adamedx:
Another question @sersut  -- will the protocol work without authentication, or must the client auth first? If it requires auth, then we may hit issues if the server is updated to a new version that requires a new version that uses a new auth protocol.

Trying to think if there are other issues with requiring authentication...

thommay:
> Trying to think if there are other issues with requiring authentication...

Well, for the client to know what version it should be updating to, it has to communicate with the chef server to get configuration - so it'll need to authenticate.

adamedx:
> Well, for the client to know what version it should be updating to, it has to communicate with the chef server to get configuration - so it'll need to authenticate.

Unless the endpoint is unauthenticated. :smile: I guess that's my question -- is the protocol for update authenticated? For comparable services in the past, the answer has been "no," but that may not be the right answer for us.

lamont-granquist:
So every implementation that we've been discussing about driving IF the upgrade should happen and WHICH version the client should update to has been pretty much inherently authenticated.

I'm going to strongly suggest that discussions about replacing omnitruck with an authenticated protocol or shipping an omnitruck endpoint with the chef-server need to be out of scope and raccoon'd.  This is not that RFC.

thommay:
> Unless the endpoint is unauthenticated

No, I mean: we're not proposing to blindly upgrade the client to the latest version, we're planning to let the admin specify which version to upgrade to - and to get the specification, you need to authenticate.

juliandunn:
Is the scope of the RFC just about upgrading the bits on disk? What's the desired behavior for restarting daemonized Chef Clients?

coderanger:
@juliandunn At least on non-windows it doesn't need to restart anything, it can just re-exec itself. Windows will require something more complex I think :-/

adamedx:
On Windows I would advise something that operates async of the chef-client run. An example would be to schedule a task that runs the upgrade. I think the requirement for all platforms is that we don't require reboots or any other special handling. The next time chef-client runs after a successful upgrade, it should be running the new version. 

Windows Update client does some pretty crazy stuff to update itself -- it's installation is more complex than chef-client and involves lots of reg keys, com registrations, outstanding API calls, local file system and registry state, etc.. We could certainly use techniques it uses including renaming in-use files so the new ones can be installed and leveraged on restart, but I think that's going overboard for us. As an implementation, I'd like to explore the async upgrade via task scheduler because it's pretty simple and has good retry behavior.

That said, for this RFC I think the details of how this is done are less important other than perhaps saying we should not require reboots or anything else just to upgrade chef-client.

lamont-granquist:
```
Is the scope of the RFC just about upgrading the bits on disk?
```

IMO yes, or this will turn into the chef-client_NG cookbook rewrite bikeshed.

Handling service restarts is clearly a necessary bit of later functionality, but I don't think it should be in the first deliverable.

nkadel-skyhook:
Enabling this by default is a horrible, horrible idea. There have been repeated cookbook/chef client incompatibilities that would break chef with an ubscheduled, untested update, even in environments that have tested the chef update on staging servers.  And the breakage might not occur with this update, but with the second, third, or tenth. I also see no current provision for release locking or rollback. Production systems should *never* be updated by default, especially with the frequent schedule used by production chef clients. This is exactly where an environment or role could be used to activate or include such an operation on a one-time basis.

Frankly, it would be safer as a cookbook to be included, not a chef built-in . a yum-chef cookbook might work well if it can be forced to be a "last" iperation. And I'd love to see a similar auto-update feature for chefdk.

Also note this could cause a serious bandwidth and resource hit on the upstream chef-client repositories when many systems in a cluster check for updates, that's another  reason not to enable this by default unless the local chef server can itself host the update.

coderanger:
@nkadel-skyhook enabling it by default has never been on the table and never will be.

lamont-granquist:
and this has to go in core, getting the mechanisms we're talking about can't be done right from a cookbook.

thommay:
> . I also see no current provision for release locking or rollback

The "Determining the desired chef version" section is all about release locking, @nkadel-skyhook .

nkadel-skyhook:
coderanger, "disabled by default" is not clear from the proposal. I admittedly assumed that it would be enabled by default from my reading: It would be very helpful to put that in the proposal and any relevant README.md

nkadel-skyhook:
@lamont-granquist Why should this need go to core instead of a cookbook? "chef" is a package: The package installers can and do correctly handle RPM, apt, and other deployment mechanisms. The completion of the task while chef is running could present issues.

lamont-granquist:
The mechanisms being discussed involve upgrading the client early, before any recipe code has been executed; upgrading the client extremely early, before any cookbooks have been synchronized; and upgrading the client extremely late in a way that will require a custom handler like the reboot handler in core chef.

lamont-granquist:
At some point we also do need to address the earlier problem with restarting the service that I suggested was out of scope, and cleanly causing the service to re-exec itself will likely require some cooperation from core chef on both ends.

thommay:
@lamont-granquist yeah. I'm hesitant to talk too much about implementation here because the scope is huge, but we'll need to solve it.

thommay:
Consensus from the mailing list is that early or extremely early is preferable.

nkadel-skyhook:
Starting with a cookbook would be valuable, especially for "knife ssh" based deployment. It would quickly reveal the necessary options in a testable configuration, and would work for ststens that have not yet updated their clients to the new, auto-update structure.

Structures that might host designated updates on a chef server  for local client download might also be good, but that's a very different use of a chef server itself. It's also useless for chef-solo based sestens with no chef server configured.

lamont-granquist:
Again, the 'consensus' on when to upgrade is 'early or extremely early'.  That requirement means doing this in a cookbook with a resource will not be technically possible as an implementation (the requirement is basically that no cookbook code runs before chef-client updates itself, so that if a crashed client is revived it will first always update itself before it pulls down new cookbook code that has been written which will fail on the version it is running).  The cookbook approach also already exists (omnibus_updater), so that implementation is already out there and reasonably tested and its warts are fairly well known.

thommay:
this is approved. Take it away, @chef/rfc-editors

new/chef-upgrade.md
```diff
@@ -0,0 +1,86 @@
...
+    "version": 12.5"
+  }
+}
+```
+A chef client on an ubuntu node in the `production` environment, will make a request to omnitruck:
+```
+http://www.chef.io/chef/metadata?p=ubuntu&pv=14.04&m=x86_64&v=12.5
+```
+and receive metadata containing the latest package in the `12.5` series.
+Should that package be newer than the current version, chef will download
```
jeremyolliver: What happens if it's older? It'd be nice to be able to downgrade, if it turns out the new chef-client version didn't work as expected - but there's always the risk that the client is downgraded to a version earlier than support for this protocol and you can no-longer use it after that.

lamont-granquist: Downgrades should be supported as well.  If you downgrade to a version that doesn't support this protocol then we'll just give you what you ask for.  Over time that becomes much less of a problem.

PR #174, opened 2016-01-07
Add author responsibility to template

thommay:
It is a problem that we have a lot of unfunded RFCs. We need to remind authors that they're explicitly committing to do the work to implement the RFC. @chef/rfc-editors @adamhjk

### DIFF
--- a/new/rfc-template.md
+++ b/new/rfc-template.md
@@ -8,8 +8,6 @@ Type: <Standards Track, Informational, Process>
 <  - https://github.com/chef/chef/issues/X>
 ---
 
-# Unless otherwise stated, the RFC author is also volunteering to implement the feature or process if accepted.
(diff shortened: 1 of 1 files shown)


coderanger:
Agree with the intent, but not in love with having that there mostly because we're just going to have remind everyone to remove it before merge (or do it ourselves).

coderanger:
Maybe in the Specification section add something like `A detailed description of the planned implementation, which the RFC author agrees to execute.`?

thommay:
Yeah, that might be better.

coderanger:
:+1:

lamont-granquist:
:+1:

btm:
:+1:

ranjib:
:+1:

PR #177, opened 2016-01-13
Subject Tagging for discourse or emailing.

jjasghar:
I added `tagging messages` and some `general topics` to start the discussion.

### DIFF
--- a/new/rfc-general-topics.md
+++ b/new/rfc-general-topics.md
@@ -1,0 +1,36 @@
+---
+RFC: unassigned
+Author: JJ Asghar <jj@chef.io>
+Status: Draft
+Type: Informational
+---
+
+# Topics for discourse or email messaging
+
+This is a list of possible general topics for discourse to help categorize messages.
+
+## Motivation
+
+    As someone who wants to help categorize their messages in discourse or email,
+    I want to add a [general topic] to the subject line of the message,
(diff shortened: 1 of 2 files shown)


adamleff:
I'm :-1: on this. We have a lot of newcomers to "the mailing list" (aka Discourse) that come to ask "new user" questions, and I would hate for their first interaction to be an over-eager member replying with the equivalent of "you're using the mailing list wrong - go read this instruction sheet for next time." It doesn't feel welcoming.

If this is an issue on which many people in the community are feeling pain, I'd suggest we look at adding additional categories to Discourse and allowing users to opt-in/out of the categories they care about. I personally don't see the value currently based on current post volume, but I'll leave that up to other people to discuss. :smile:

thommay:
I'm with Adam on this :-1:

jaymzh:
Agreed on the :-1: - though as with many mailing lists, if you make it a habit of doing nice subject tagging, others may follow suit. That said, I'm not opposed to forking of a separate mailing list for topics that turn out to be large-volume. One dumping ground can get very noisy.

thommay:
The general consensus is that we don't really need to be prescriptive about tags right now, given the relatively low traffic volume through discourse.

If our community want to volunteer their time to tagging posts after the fact that would be appreciated and definitely not discouraged, though.

PR #178, opened 2016-01-14
updated CBGB members

Karobeck:


coderanger:
RFCs are generally policy documents. This should probably not live here. Maybe the MAINTAINERS.toml in the chef repo?

nathenharvey:
We'll implement following the guidance in #179.  A `CBGB.md` file will be added to chef/chef.

PR #179, opened 2016-01-15
Add a CBGB markdown file to the repository

nathenharvey:
Rake task for generating CBGB markdown from a toml file

### DIFF
--- a/CBGB.md
+++ b/CBGB.md
@@ -1,0 +1,40 @@
+<!-- This is a generated file. Please do not edit directly -->
+<!-- Modify CBGB.toml file and run `rake cbgb:generate` to regenerate -->
+
+# Chef Board of Governance (CBGB)
+
+  Chef was designed from the outset to have a very open structure, including open design, open contribution, and consistent use of tools across the project. Given the large numbers of contributors, users, and companies with a stake in the future of the project, Chef leadership has established an advisory board, as part of its long term commitment to open governance.
+
+  The Chef Board of Governance (CBGB) shall advise the Leadership on matters related to supporting the long-term governance, structure, and roadmap of the Project.
+
+More information can be found in the [Chef Board of Governance RFC](Chef Board of Governance).
+
+# Board of Governers
+
+## Project Lead
+
+* [Adam Jacob](https://github.com/adamhjk)
+
+### Users/Contributors (4)
+
+* [Ranjib Dey](https://github.com/ranjib)
+* [Doug Ireton](https://github.com/dougireton)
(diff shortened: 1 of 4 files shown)


nathenharvey:
This is an alternate implementation of #178.  Instead of listing the CBGB members in the governance RFC, we would list them in a separate `CBGB.md` file in the RFC repository.

/cc @Karobeck @coderanger

coderanger:
I'm still not sure why we are putting this in the RFC repo. This isn't an RFC. Is there a reason for it?

nathenharvey:
What do you think? 

Should the list of CBGB members be stored:

:one: - Directly in the RFC that describes the CBGB
:two: - In a separate CBGB file in the RFC repository
:three: - As a section in the MAINTAINERS file of the main chef project
:four: - In a separate CBGB file in the chef repository

Please vote below.

jonlives:
I'd lean towards 4) just to keep it distinct from the Maintainers file, as not all CBGB members are necessarily maintainers.

thommay:
I'd be 2 - 4 - 1 - 3.

btm:
I guess 4, 2.

[RFC029](https://github.com/chef/chef-rfc/blob/master/rfc029-governance-policy.md) scopes "Project: Means the Chef open source software and all applicable policies and procedures and guidelines." That sounds like chef/chef, so it makes sense to keep it in chef/chef.

We're not clear on how [RFC030](https://github.com/chef/chef-rfc/blob/master/rfc030-maintenance-policy.md) applies to projects other than chef/chef. There are different interpretations, and so depending on who is doing the work different repositories are seeing difference processes.

ranjib:
3 or 4, if we do it at all.

I expect the CBGB folks will change with time, and the core chef repo is
authoritative source for most of the people stuff (maintainers, lts etc),
and has highest amount of folks as watching/working on it, with much better
responsibility delegation (who will update this file if it stays in
chef-rfc repo?). Also chef-rfc repo is not very intuitive to navigate
(unless you have some aprior knowledge about an rfc)

regards
ranjib

nathenharvey:
@ranjib the CBGB will change at least annually and the secretary of the CBGB or the Community Ombudsperson should be responsible for updating the file, wherever it exists.

martinb3:
I think 4 makes the most sense. I'd love to figure out how to link it widely and externally and on the Chef.io site as well.

thommay:
Decision was that the CBGB file would be in chef/chef and then widely disseminated from there.

nathenharvey:
https://github.com/chef/chef/pull/4445

PR #184, opened 2016-02-09
Update supported platforms

tas50:
Remove platforms that are EoL with their vendors and add new versions of
openSUSE and OpenBSD

### DIFF
--- a/rfc021-platform-support-policy.md
+++ b/rfc021-platform-support-policy.md
@@ -33,7 +33,7 @@ Platform | Versions | Architectures | Package Format | Built on
 --- | --- | --- | --- | ---
 AIX | 6.1, 7.1 | ppc64 | bff | AIX 6.1
 CentOS | 5, 6, 7 | i386, x86_64 | rpm | RHEL 5
-Debian | 6, 7, 8 | i386, x86_64 | deb | Debian 6
+Debian | 7, 8 | i386, x86_64 | deb | Debian 6
 FreeBSD | 9, 10 | i386, amd64 | pkg_add pkg | FreeBSD 9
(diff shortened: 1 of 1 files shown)


lamont-granquist:
:+1:

ranjib:
:+1:

rfc021-platform-support-policy.md
```diff
@@ -33,7 +33,7 @@ Platform | Versions | Architectures | Package Format | Built on
 --- | --- | --- | --- | ---
 AIX | 6.1, 7.1 | ppc64 | bff | AIX 6.1
 CentOS | 5, 6, 7 | i386, x86_64 | rpm | RHEL 5
-Debian | 6, 7, 8 | i386, x86_64 | deb | Debian 6
+Debian | 7, 8 | i386, x86_64 | deb | Debian 6
```
coderanger: Are we actually building on Debian 6?

coderanger: Apparently we are, plan is to fix that later.

thommay:
:+1:

adamhjk:
gif-keyboard-16951731606949895246

I approve this RFC. /cc @chef/rfc-editors

rfc021-platform-support-policy.md
lamont-granquist: Indeed we is.   It came up recently because we started down the road of shaving that yak in EC2 for the new acceptance tests (testing chef-client against a pile of cookbooks automatically before we ship it) and @jkeiser was getting sucked into the yak hole and I pulled him out before he fell all the way in:

https://github.com/chef/chef/pull/4461

rhass:
Based on this RFC, versions RHEL < 5.10 should be on the chopping block as well.

lamont-granquist:
because of ABI compatibility we consider all RHEL 5.x to be a single platform version.

PR #186, opened 2016-02-17
RFC-meetbot

jjasghar:
This RFC is for adding an IRC/Chatbot to track our meetings for the community.

### DIFF
--- a/new/rfc-meetbot.md
+++ b/new/rfc-meetbot.md
@@ -7,21 +7,21 @@ Type: Process
 
 # Enabling meetbot for Chef IRC Community meetings.
 
-I would like to enable [meetbot][meetbot] for the Chef IRC Community Meetings.
-The ability to automatically take notes and executive summaries can open up
-people to have conversations instead of the dogma of making sure the conversations
-stay on track.
+We will enable [meetbot][meetbot] for the Chef IRC Community Meetings.
+Shifting responsibility for taking meeting notes from the designated secretary
+to software allows the person acting as the secretary to instead participate
+fully in the discussion.
 
 ## Motivation
 
     As a community member of Chef,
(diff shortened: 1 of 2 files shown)


new/rfc-meetbot.md
```diff
@@ -0,0 +1,65 @@
...
+Status: Draft
+Type: Process
+---
+
+# Enabling meetbot for Chef IRC Community meetings.
+
+I would like to enable [meetbot][meetbot] for the Chef IRC Community Meetings.
+The ability to automatically take notes and executive summaries can open up
+people to have conversations instead of the dogma of making sure the conversations
+stay on track.
```
btm: From IRC conversation, I think you mean something like:

Shifting responsibility for taking meeting notes from the designated secretary to software allows the person acting as the secretary to instead participate fully in the discussion.

new/rfc-meetbot.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Motivation
+
+    As a community member of Chef,
+    I want to have a higher level of engagement from everyone,
+    so that we automate the human process away, and the bot does the note taking.
```
btm: nit, " and the bot does the note taking" doesn't feel like it belongs in the user story.

As a community member of Chef,
I want the meeting notes to be automated,
so that the secretary can participate fully in the discussion.

btm:
I'd like the writing itself to be more declarative, take out "I thinks" and "I believes." 

Also watch tense. I'm having a hard time figuring out a good way to explain this, but I think when accepted the language of the RFC should describe how something is, not how we would potentially implement it. That isn't to say that we should describe the current state instead of the future state, I'm just referring to the language.

thommay:
approved, although we probably won't implement this until #191 is shaken out. @chef/rfc-editors

PR #187, opened 2016-02-18
Multiple policy support

coderanger:
Pushing this up for early feedback. Would love a better list of all the reasons this is a bad idea :)

### DIFF
--- a/new/mulit-policy.md
+++ b/new/mulit-policy.md
@@ -39,6 +39,25 @@ are wonderful options.
     I want to release cookbook updates,
     so that I can deploy changes quickly.
 
+With that said, a more specific case that isn't currently handled well:
+
+The Database team owns MySQL and uses Chef to deploy it along with related configs
+and tools. The Monitoring team owns Collectd, which is deployed to every server
+in the organization, and they are responsible for configuring collectd and deploying
+new versions of it as needed. The core team owns ntpd and is responsible for its
+configuration.
+
+All teams want to follow the snapshot-based approach where they have a mono-repo
+of their own cookbooks with a few shared cookbooks pulled in via dependencies.
+When a team wants to do a release they run `chef update` to recompile their
+policy, and then use `chef push` to roll it out to each environment in sequence
+(with time in-between for testing et al). The teams also do not want to have to
+ask permission from any other team when deploying new versions of a service they
+are responsible for.
+
+This combination of a snapshot-based workflow without explicit team coordination
+is not currently an easy thing to do with the policy systems.
+
 ## Specification
 
 A solution for this is to use multiple, decoupled policies on the same node.
--- a/new/mulit-policy.md
+++ b/new/mulit-policy.md
@@ -62,6 +62,75 @@ attributes), but everything else (resource collection, cookbook collection)
 would be reset between policies. An error in one policy would abort the run,
 meaning later policies would never get run.
 
+## Implementation
+
+The major implementation change would be altering `Chef::Client#build_node` to
+yield a node object to a block rather than setting it as a singleton. This
+will allow running the block (containing all the same logic as we have now for
+running the converge) multiple times.
+
+Given a Chef run like `chef-client --policy-name one,two` we would do the
+following:
+
+1. Initialize the client up until `load_node` in `Chef::Client#run` as per normal.
+2. When the policy builder checks which implementation to use it would see that
+   multi-policy run has been requested.
+3. It would build the first node object as per normal except that the cookbook
+   synchronizer would sync to `Chef::Config[:file_cache_path]/cookbooks_one` or
+   similar. This may require some work on the `Chef::FileCache` code.
+4. Before yielding the node out to the run block to policy builder code calls
+   fork. The original process blocks, while the subprocess yields the node object
+   and runs the first converge.
+5. The first converge finishes, control returns from the run block. The policy
+   builder would save node attributes to a file and exit.
+6. Control resumes from where the original process blocked in step 4. It checks
+   that the child exited with 0, reads in the saved node attributes and updates
+   the node object, and then GOTO 3 to sync the next cookbooks and run the next
+   converge.
+7. When the last policy has finished, control flows back to the `#run` method.
+   It would pick back up at `run_status.stop_clock` to do the end-of-run stuff.
+
+If a multi-policy run has an exception we would serialize it to JSON, return
+an error exit code from the sub-process, and have the main process bubble the
+(fake) exception back up the chain.
+
+## Problems
+
+The biggest issue is that this makes no attempt to prevent teams from stepping
+on each others. If one policy has `package 'foo' { version '1.0' }` and another
+has `package 'foo' { version '2.0' }` it will oscillate between versions without
+warning. This is probably not a big issue as today Chef will do the same thing
+in similar situations (i.e. if the names don't match but `package_name` does).
+Weird stomping could also occur if two policies are trying to "own" a service
+like Nginx and using two different versions of the same cookbook. These are all
+deemed problems out of scope for a technical solution. When using mutli-policy
+workflows, external synchronization (i.e. talking to each other) will be needed
+to establish some idea of who is responsible for what.
+
+Because each converge will happen with its own (forked) copy of the run context,
+resources from one policy with not be able to see those from any other. This
+means cookbooks like Zap will likely be non-functional. It also doesn't help
+with situations where the interface from one team to another is a resource or
+helper library, those situations will still require a single policy with all
+the snags that implies today.
+
+Overall the added fork involved in multi-policy converges may produce additional
+unforeseen complications, such as cookbooks that register event handlers or other
+callbacks that need to "live" past the converge phase. We could ameliorate this
+by saying the first (or last?) policy will be run without that extra fork and
+so can interact with those pieces of the run. This still leaves that "slot" as
+special though, and may just kick the can down the road as far as contention.
+
+## Other Possibilities
+
+The main suggestion for an alternate way to handle this would be to combine
+multiple Policyfiles into a single compiled policy somehow. These solutions all
+suffer from a shared flaw though, that individual teams would be unable to use
+the "compile, push staging, push prod" workflow that Policyfiles support today.
+This "loose-coupling" approach may be workable in some places where there are
+strong release controls and tooling, but will not support the full Policyfile
+workflow without extensive team coordination.
+
 ## Downstream Impact
 
 Any tooling that consumes node data and expects `policy_name` to be a single
--- a/new/mulit-policy.md
+++ b/new/mulit-policy.md
@@ -1,0 +1,75 @@
+---
+RFC: unassigned
+Author: Noah Kantrowitz <noah@coderanger.net>
+Status: Draft
+Type: Standards Track
+---
+
+# Multiple Policyfiles and Teams
+
+Policyfiles allow powerful new workflows around cookbook management, but one
+area they are currently lacking is in dealing with mutli-team organizations.
+
+Currently each node can be attached to exactly one policy and one policy group.
+The organizational manifestation of this is generally that there needs to be a
+single "owner" for the policy of that node. This might be a CI tool like Jenkins
+or Chef Delivery, but in a lot of cases in the current workflow this will be the
+team within the company that uses the machine most.
+
+This works great for the "owner" team, they can use all the power and
+flexibility of the Policyfile workflow to its fullest. For other groups
+managing software on the node, the picture is less clear. As an example, take a
+database server owned by an application team. On this server we have a mix of
+software, the MySQL install and configuration is manage by the DBA team while
+ntpd and rsyslog are managed by a core infrastructure team. These live in
+different cookbooks so for the most part things are copacetic. When the DBA
+team wants to roll out a new MySQL configuration entry they update the cookbook,
+recompile the policyfile, `chef push` that compiled policy to the QA policy
+group, and then follow the usual Policyfile roll-out procedure. For the core
+team, things are more complex. They could follow the same process, but this
+would require careful coordination with the DBA team to ensure they don't have
+a roll-out in progress (see https://yolover.poise.io/#21). They could also
+release their cookbook changes either to a git branch or Supermarket server,
+and then notify the DBA team that they should run a roll-out. Neither of these
+are wonderful options.
+
+## Motivation
+
+    As a maintainer of core, company-level cookbooks,
+    I want to release cookbook updates,
+    so that I can deploy changes quickly.
+
+## Specification
+
+A solution for this is to use multiple, decoupled policies on the same node.
+This can be done today by using two different node names on the same machine,
+each with their own policy. While workable, this solution is inelegant and
+frustrating to manage.
+
+This could be done more directly by allowing the `policy_name` on a node to be
+set to an array, running each policy in isolation in order:
+
+```bash
+$ chef-client --policy-name one,two
+Converge policy one
+# ...
+Converge policy two
+# ...
+```
+
+Any data that is server-resident could be shared between runs (eg. node
+attributes), but everything else (resource collection, cookbook collection)
+would be reset between policies. An error in one policy would abort the run,
+meaning later policies would never get run.
+
+## Downstream Impact
+
+Any tooling that consumes node data and expects `policy_name` to be a single
+string would need to be updated.
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.
--- a/new/mulit-policy.md
+++ b/new/mulit-policy.md
@@ -8,7 +8,7 @@ Type: Standards Track
 # Multiple Policyfiles and Teams
 
 Policyfiles allow powerful new workflows around cookbook management, but one
-area they are currently lacking is in dealing with mutli-team organizations.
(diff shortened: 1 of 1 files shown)


jonlives:
The first thing that sprung to mind here would be that this might make it harder to detect duplicate resources, ie two cookbooks doing different things to the same file / package. Without a single resource collection, I'm not sure there'd be any way to diagnose this without a lot of digging.

lamont-granquist:
I think we need something like this.  At a very Enterprise-ey level different groups have largely orthogonal concerns and need compartmentalization.

To first order, duplicate resources shouldn't occur because of orthogonal concerns.  The security group writing code which strips the setuid bit off of binaries in order to lock down avenues of local access being exploitable to gain root has nothing to do with an application group deploying a rails or java application.

To second order, groups need to be writing defined APIs.  So the monitoring group needs to own the code which sets up the monitoring infrastructure and installs the necessary clients and things on the server.   They also need to vend resources to the application groups so that the application policies can use those resources to configure monitoring of the application without stomping on the resources that is in the separate policy that the monitoring group runs on the box.   If Enterprise-ey users don't approach the problem like this and the monitoring group just tosses some of their own code in a policy and then all the app groups poke random hand-rolled resources to talk to the monitoring service on the box, then it'll turn into a mess, but I don't think that should be viewed as a problem to solve in the underlying infrastructure, that's just shitty design.

Also i think this would inherently implement:

https://tickets.opscode.com/browse/CHEF-2070

Then the security group's policy could always run first and always run to its completion.  If the Software Devs pushed out bad code that crashed chef-client later, the security policy is still guaranteed to be applied.

So :+1:

coderanger:
@lamont-granquist Yeah, I the case of library-ish access is well handled (if rarely done in practice) by current patterns, the case I'm more concerned with is the monitoring team being able to roll a new version of collectd to the whole infra without disrupting app teams.

jonlives:
I'm not sure depending on people to only use this in the way we would like them to is the safest mechanism to count on here - that said, I do like the underlying idea, but I think we need to think a bit more about how multiple resource / cookbook collections per run will change a lot of assumptions people have historically been able to make about how Chef works during a run, the above being one example.

lamont-granquist:
We don't try to prevent people from writing chef recipes using nothing other than execute resources to create bash scripts in Chef.  Falls under the same umbrella to me.

I also think I like this much more than trying to sort out multiple completely orthogonal chef-client runs on the same box.  If the array of policies acts like a meta-run_list so that the compile/converges always happen and always happen in a given order, and the node data flows through all of it, then it solves a lot of potential cross-talk issues.

coderanger:
Agreed, most of the reason I want this to stew for a while is to try and find the structural reasons it won't work. I think at a minimum we'll need to fork before each multi-policy converge because we can't unload cookbooks to any useful degree. That would keep everything isolated by default in memory, but we would need to implement something for sync'ing node data between the sub-runs.

new/mulit-policy.md
```diff
@@ -0,0 +1,75 @@
...
+RFC: unassigned
+Author: Noah Kantrowitz <noah@coderanger.net>
+Status: Draft
+Type: Standards Track
+---
+
+# Multiple Policyfiles and Teams
+
+Policyfiles allow powerful new workflows around cookbook management, but one
+area they are currently lacking is in dealing with mutli-team organizations.
```
thommay: "multi"

lamont-granquist:
Do we need that complexity, though?  I would hope the different teams with orthogonal concerns manage completely different cookbooks, so we could figure out what cookbooks all the policies needed and synch them all...  I guess it gets a little whacky if different teams have libraries which stomp on each other which would monkeypatch the same process...  And what to do about loading resources from cookbook thats are in other runs, and attribute files, hrmmmm....

jonlives:
@lamont-granquist there's dealing with every possible case and dealing with reasonably foreseeable ones, of which - much tho we might wish otherwise - duplicate resources are one of the more common. 

I like your suggestion of meta run lists though, that sort of thing is much more how I'd like this to work in my head, so we still preserve a consistent overall compile / converge

jonlives:
It's also worth considering the monorepo case which often lends itself (and does in our case) to multiple cross-functional teams working out of the same cookbooks rather than the cookbook-per-repo model. Really what I'm getting at here is that I don't think we should depend on people  to use cookbooks in our ideal way to avoid running into issues here.

What I really want to avoid is having no way to indicate to people that "hey, these two policies you're using clash in some manner", because that has the potential to be unpleasantly surprising in all manner of ways :p

coderanger:
@lamont-granquist I don't see any generalized way around the talking stick problem. Even with partial policy updates (so the monitoring team could recompile all policies in the org for the new collected cookbook), if any team has a deploy running it would go poorly. And if the new collectd cookbook has some dependency upgrades, that could easily break things for some app team's cookbooks somehow.

lamont-granquist:
Hmmm, yeah the depsolver version of the diamond inheritance problem.  Also if the Monitoring team vendors a cookbook with resources and you have two different Application teams that both depend on it, but need to update separately.  That's annoying.

coderanger:
I think we should separate out two use cases,

One is a monitoring team running a collectd daemon on all hosts that wants to be able to upgrade collectd at any time with no disruption to other teams.

Two is a monitoring team maintaining a core collectd cookbook that other teams can use, via a `collectd_plugin` resource, to install plugins into a system level collectd.

Those are very overlapping, but with super different workflows for how updates are distributed.

lamont-granquist:
Can we compile the set of cookbooks like we compile a C application?  So a policyfile by the application team might have an unlocked-unsatisfied dependency on the collectd cookbook that the monitoring team pushes out.  We would then "link" all the policyfiles together and expect all the "symbols" to resolve.  The monitoring team could them bump the version of the collectd cookbook and both application teams would pick it up.  We could detect conflicts at "link" time if the App team had constraints on their dep that were not met by the collectd cookbook the monitoring team was pushing out (obviously this gets found somewhere back in acceptance testing rather than failing in prod or possibly sooner if the "link" happens on a workstation).

coderanger:
@lamont-granquist I feel like that would remove some of the promises of policyfiles, "this is exactly what will run no matter what". Maybe that would be okay though, if we could make the linkage points somehow explicit? The trouble points would probably be versions of dependent cookbooks that are common between both policies. Also we should make any solution here work for more than the two teams.

charlesjohnson:
So it sounds like there's two value propositions we're looking for here:

1. Multiple teams can each be responsible for some slice of a host configuration, without stepping on each other or invalidating configs laid down by another team.

2. Restore the broad composability that role inheritance allowed, that was lost in move to policies.

In addition, there are valuable things policies have brought to the table in their current form, that we don't want to lose while adding that value:

- Predictability. I can tell roughly, at a high level, approximately what configuration will be applied to a node based on inspecting a human-readable document (the policyfile / policyfile.lock).
- Repeatability. I can be much more confident that applying a policy to a node today, will give me the same result if I apply that policy in 6 months.
- (Rough) Immutability. I have a single, verifiable bundle that contains the policy and all of the code necessary to apply that policy, and it cannot be changed without recompiling and getting a new signature.
- Simplicity. Once I have compiled my configuration into a locked policy, I no longer have to manually manage or link object relationships or dependencies in order to apply or socialize that policy.

(I'm sure there's value goals that I'm missing, and that there's existing value to protect that I've overlooked, so please pipe up.) As we look to add these new pieces of value, we should be very careful in considering which pieces of existing value we trade in order to provide the best set of features.

OK, product guy hat off, OSS nerd guy hat on (implementation ideas follow).

This is rough, but:
Rather than allowing a node to have multiple policies, I propose we keep one-policy-per-node, but allow policies to chain, inheriting from one-and-only-one upstream.In this way we can build layers of policy: 

`Infra policy <-- Security policy <-- Runtime policy <-- App policy` etc. 

Changes to upstream policies could be grabbed at policy compile time. Enforcement of upstream policies could be done by checking the policygroup prior to push (server rejects push of app policy because its upstream security policy has the wrong hash for this policygroup).

Attribute precedence could flow in a predictable way: defaults flow bottom-up (app beats infra), overrides top-down (infra beats app). If this is compiled into the policyfile.lock at build time, then the policyfile.lock can tell us which policy the final attribute value came from. We can also more easily find conflicts at build time, and tell the user what conflicts where.

This might be much easier to reason about, and it would be possible for humans to understand what the final config solution would look like before actually applying the config to a node.

Thoughts?

coderanger:
I agree that would be much easier, but I don't think it helps with the required use cases. If there is one policy, then there is one policy compile event. We would need to be able to recompile the "infra policy" without changing the "app policy" or anything else in the chain. Additionally this is weird (but workable) for multiple teams that have no hierarchy between them. Should monitoring inherit from logging or vice versa? In the end it probably wouldn't matter, but I don't think it meshes well with actual org structures.

coderanger:
Also anything with a single final policy will have issues with the talking stick/single pipeline issue, though with a tree-based approach (a la Docker's `FROM`) we could try to track things and at least warn when a policy is being applied out of order.

charlesjohnson:
In this idea, if infra policy is recompiled and pushed to the policygroup, then app would show as outdated. New pushes to app would be disallowed unless they inherit from the new infra policy.

This doesn't allow for infra to push a change without waiting for app to intentionally pick it up, but there's a case to be made for that behavior to be more a feature than a bug, since app is typically much closer to business value than infra.

coderanger:
@charlesjohnson That doesn't cover the use case of "monitoring team pushing an upgrade of collectd" though, they would have to push a new shared policy and then bug every team to run a release, which they might not be in a position to do immediately.

charlesjohnson:
It does cover that case, just not in a way that the monitoring team might like very much. But App team is probably thrilled that monitoring can't change the underlying infrastructure without telling them.

coderanger:
I mean yes, but the whole point is to allow that use case smoothly. Otherwise we can do all this with existing tools for the most part :) From what I've seen of large-scale users, the teams don't care about each others' services, they just want it all to work. Upgrading a core service like collectd or ntpd seems like something we should support given how often that kind of stuff comes up.

coderanger:
If the decision of the org is that app teams should have total control over rollouts to their boxen, then that's totally cool and pretty much works already. That isn't the case I'm concerned with since it seems to only happen at relatively small scales.

lamont-granquist:
I quit at a previous job because the app teams wound up owning the boxes and it was becoming completely politically untenable and infeasable for me to be able to manage and rollout the 'base role' and i was pretty sure that i'd become unable to get my change reviews pushed out, and if i ever screwed up once and one of the (hundreds of) app teams threw me under the bus i'd have gotten fired.  When there's 30,000 server images and 7,000 "roles" and a couple hundred different little software teams being the one guy who is trying to upgrade the base platform becomes kind of a shitty job.  Technological hurdles that require you to beg and plead for every single app team to do a rollout in order for all your changes to hit prod would have gotten rid of me a lot quicker...

So, yeah, to make this feature worthwhile, the Monitoring team needs to be able to do a rollout to every single box under their control, without changing the pinned versions of cookbooks that the app teams are using.  As long as everyone is semver and API compatible that is expected to not break anything.  If it does break anything, acceptance testing should explode and prevent the rollout to production.  The rollout to production of the monitoring app should not cause changes to software.

And the use case is a bit more solid if you consider the needs of the security or system ops teams when a shellshock-style vulnerability gets deployed and they need to quickly react and upgrade bash everywhere and report on their compliance.  Begging 100 app teams to go push a button is a very bad way to do security compliance.

Lots of potential for chaos, but we have to assume automated acceptance testing here.  The centralized team needs to be able to push global changes.  The app teams need to have at least basic flame tests so that an accidental `chmod 0600 /etc/resolv.conf` or something causes the rollout to fail somewhere in acceptance/union/rehersal stages.  If the app teams don't have any kind of testing then the centralized team can completely break DNS on their servers and their app will crash in production.  The biggest problem to fix there is the lack of app testing.

coderanger:
I'm also happy with this being a hazmat-level solution. If we can get something logistically equivalent to the `chef-client -c /etc/chef/chef-core.rb && chef-client -c /etc/chef/client-app.rb` multi-client solution but less painful to maintain, it would be a start to at least build a workflow on.

lamont-granquist:
Yeah, that multi-client solution is terribad.  I think we need one node, one client, one ohai run, and the node needs to thread through the whole thing and get POST'ed at the end.  Then we have two resource collections that get constructed and converged in order.  That's the obvious stuff that I think we agree on?

coderanger:
Yeah, I think that's the right structure. The details of how we load two possibly-unrelated sets of cookbooks is the biggest falling point right now I think.

lamont-granquist:
I guess one implementation question is if there need to be multiple policyfile.locks or if there could be one "workstation" (CI job most likely) responsible for compiling multiple policyfiles into a single lock to apply to the node?  There's already feature requests to have policyfiles slurp up other policyfiles (somewhere?  i can't find it.  maybe that already got implemented?).  There's also all the various feature requests kicking around for having multiple compile/converges within the same chef client run.

coderanger:
I don't see how we would implement that without breaking the stability/repeatability promises of policyfiles though. If the two policies being combined have, say, two different git SHAs for a shared dependency (whatever was current when each team ran `chef update`) there is no programmatic way to unify them. Even with a CI server handling the actual policy compilation (a la Delivery) there is still the problem of any single compiled policy having a single pipeline so we would need some kind of locking mechanism or another way to detect when it is safe to `chef push`.

ranjib:
This RFC is very similar to the chef-stage stuff, except the chef-client runs are targeted with specific policy files. I concur with @jonlives  that in its current form this is confusing and risky. Two teams can just rewrite each others resources.
For me this might be ok to have in chef core if we have analogous abstractions like resource collection and why run mode that spans across the chef run  (i.e multiple policyfile stages). Also what details of node's state is stored, and remain searchable (currently each run will invoke node.save), or how this impacts desired state RFC .

danielsdeleo:
> I guess one implementation question is if there need to be multiple policyfile.locks or if there could be one "workstation" (CI job most likely) responsible for compiling multiple policyfiles into a single lock to apply to the node?

> I don't see how we would implement that without breaking the stability/repeatability promises of policyfiles though. If the two policies being combined have, say, two different git SHAs for a shared dependency (whatever was current when each team ran chef update) there is no programmatic way to unify them.

It wouldn't be that difficult to declare one policy the owner, which would then feed into the merge process whenever there's a conflict. Downside of that is that it opens up the possibility of using different code than you tested locally (and SemVer just means that no one intended to break anything, not that they actually didn't), so the merged thing would need to go through some kind of pipeline to test the combined thing. And if you had that pipeline, you could just as easily implement a fan-out for base things to trigger recompile and re-test of their downstreams and you wouldn't need this feature at all (side note, this is how I did it for a MVP integration with delivery).

On the other hand, if you're gonna allow two policies to have a different version of the same cookbook, that pretty much necessitates doing at least multiple forks within the chef run to ensure that you don't get weird effects from loading the code twice (e.g., getting a class variable set to a bad value if a library gets loaded multiple times). Which means to make the node object have the right state at the end, you'd need more advanced IPC than we've implemented thus far.

It would be nice to simplify things by disallowing conflicting dependencies, but that would probably be a disaster for Windows which has a lot of "should-be-core" functionality in the windows cookbook (also only one team could use a cookbook like chef-sugar).

lamont-granquist:
So I guess that's what I was trying to get at.  It seems like there's two solutions:

There's multiple PolicyFiles, but one thing controls producing the pinned lockfiles and that artifact is what gets shipped through CI.  That solves the problem at the higher level of putting together disjoint concerns into a single run_list and converging it.  But I'm not sure if just having the client do compile/converge-per-logical-policy gets to solving the root issue here.   That approach does completely avoid the multiple-cookbook-versions concern though because there's one actor responsible for putting all the logical Policyfiles together and depsolving them.

If you're really putting multiple Polifyfile.locks together with pinned cookbook versions then it gets a bit weirder.  Presumably though you could design a button where you depsolve where a set of existing Policyfile.locks becomes part of the constraint.  That way the ops team could take the lock file that the dev team was using in prod yesterday and depsolve their Policyfile against that to produce a lockfile which was consistent which they would then deploy?  The net result might be that something like chef-sugar would not necessarily float across the same version in prod for all products -- if that became a problem, though, it could be made the concern of only one group who would be responsible for pinning it and then everyone would depsolve against their artifact.

lamont-granquist:
But yeah :+1: on the details of the implementation need to be fleshed out a bit more about how the policyfiles combine.

coderanger:
I think any solution where the different policies get depsolved together at any point is just a total non-starter. The CI-driven merge/solve is cool if your org can do that, but that works okay with existing tools. The point of this is to avoid the need for the teams to worry about breaking each other indirectly.

lamont-granquist:
Yeah, i think you need to write up the technical details of the workflow, and where the workflow solves problems that the default workflow doesn't have, from a sort of keyboard-centric perspective, because it all feels a bit too handwave-y and abstract right right now...

coderanger:
Okay, I've updated this with a bit more detail.

thommay:
I am :+1:

lamont-granquist:
:+1:

danielsdeleo:
From IRC:

> I'm +1. I think the correct thing to do for teams is to implement Ci that does the fanout from cookbook updates to policies that include them, but I think Chef should let you be productive without that

:+1:

thommay:
accepted. Please let's get this merged, @chef/rfc-editors

charlesjohnson:
I am a hard :-1: on this as written. The explicit team coordination required by policyfile as it exists is a feature, not a bug.

adamleff:
Merging in now - accepting as RFC 075.

PR #189, opened 2016-02-23
Update RFC062's defined and deprecated exit codes

smurawski:
Chef Client currently defined "2" as the exit code for a SIGINT signal and "3" as the exit code from the SIGTERM signal.  This PR adds those to the currently defined list of exit codes from Chef Client and notes their deprecated state.

In working to implement this feature, I've found that the difference between "Reboot Now" and "Reboot Scheduled" is a very fine line.  If the reboot is scheduled to happen immediately, Chef Client doesn't actually get to terminating cleanly.  If the reboot has a sufficient delay that Chef Client can terminate, the reboot will be scheduled with a timer that is asynchronous to the chef-client's termination.  I think, given the Motivation defined in the RFC, "Reboot Scheduled" fills the need to notify the user or other applications that a reboot has been scheduled.

### DIFF
--- a/rfc062-exit-status.md
+++ b/rfc062-exit-status.md
@@ -14,7 +14,7 @@ Signal outside tools of specific Chef-Client run status.  Ability to determine r
     As a Chef user,
     I want to be able to determine when a chef-client run is rebooting the node,
     so that Test-Kitchen/Vagrant/any outside tool can wait for node to reboot, and continue converging.
-    
+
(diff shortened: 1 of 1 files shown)


thommay:
:+1:

thommay:
17:27 < ranjib> +1
17:28 <@coderanger> +1
17:28 < martinb3> +1

PR #190, opened 2016-02-25
Add rfc around artifact yanking

nellshamrell:
This comes up fairly frequently - whether or not we should allow someone to yank a whole cookbook or a specific version of a cookbook.  This RFC is intended to provoke discussion on this with the intention of coming to a decision.

### DIFF
--- a/new/artifact-yanking.md
+++ b/new/artifact-yanking.md
@@ -16,15 +16,15 @@ Cookbook and Cookbook Version Yanking Policy
 
 As a Supermarket User
 
-When I download an artifact (for now, a cookbook) from Supermarket
+When I download an artifact from Supermarket
 
 And I depend on that artifact
 
 I do NOT want that artifact to be removed from Supermarket
 
-As a Artifact author or collaborator
+As a artifact author or collaborator
 
-When I upload a bad version of a cookbook to Supermarket
+When I upload a bad version of an artifact to Supermarket
 
 I want a way to hide it from the Supermarket UI
 
@@ -32,7 +32,7 @@ And from the /universe endpoint
 
 ## Specification
 
-Currently a Supermarket artifact author cannot remove a cookbook version from Supermarket without contacting the Supermarket admins.
+Currently, a Supermarket artifact author cannot remove a cookbook version from Supermarket without contacting the Supermarket admins.
 
 A cookbook author CAN remove a cookbook completely from Supermarket.
 
--- a/new/artifact-yanking.md
+++ b/new/artifact-yanking.md
@@ -1,0 +1,51 @@
+---
+RFC: unassigned
+Author: Nell Shamrell-Harrington <nshamrell@chef.io>
+Status: Draft
+Type: Process
+<Replaces: RFCxxx>
+<Tracking:>
+<  - https://github.com/chef/chef/issues/X>
+---
+
+# Title
+
+Cookbook and Cookbook Version Yanking Policy
+
+## Motivation
+
+As a Supermarket User
+When I download an artifact (for now, a cookbook) from Supermarket
+And I depend on that artifact
+I do NOT want that artifact to be removed from Supermarket
+
+As a Artifact author or collaborator
+When I upload a bad version of a cookbook to Supermarket
+I want a way to hide it from the Supermarket UI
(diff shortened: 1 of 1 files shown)


nellshamrell:
Relevant discussion: https://github.com/chef/supermarket/issues/954

coderanger:
RubyGems does this right in my opinion. You can yank a version (but not a whole gem AFAIK) which will nuke the files and mark it in the UI as yanked. Once yanked, nothing can ever be re-uploaded to that version. This does mean you can break other people when yanking, but given that even Chef has had to yank releases due to major problems I think we've all benefited from it.

nellshamrell:
More relevant discussion https://github.com/chef/supermarket/issues/954

lamont-granquist:
actually, I'd argue that rubygems does it kinda wrong.  yanking gems is terrible, and i've created chaos doing it.  most of the time the person yanking the gem shouldn't be yanking the gem and should solve their problem some other way.

See:

https://github.com/rubygems/rubygems/issues/858
https://github.com/rubygems/rubygems/issues/1506

Particularly:

https://github.com/rubygems/rubygems/issues/1506#issuecomment-188448748

The 'hide' proposal here is way safer.  Since most people who want this feature are just trying to reduce cookbook versions and make depsolving faster, all we have to do is hide the cookbook versions from /universe and it accomplishes what they want.  For wiring it up to 'unshare' it makes more sense since that'll be the goto API for people who think they want to delete things and I'd prefer they don't wind up doing a yank.

We could include another API which is the permadelete-and-mark-version-as-forever-not-being-usable-again that is the rubygems 'yank' behavior but that should really be moved somewhere that it wasn't the first API that new users reached for when they wanted to 'unshare'.

lamont-granquist:
I do think the proposal could be a bit more explicit about the API.   Such as the 'unshare' is going to POST to an API to update the cookbook that it was hidden, and that url and format should be specified.  Then it should be specified that supermarket responds to the cookbook version download with 200s for hidden cookbook versions and give a couple of examples of APIs where it still appears, then I *think* the only endpoint it disappears from is /universe.  For the endpoint that returns the cookbook version metadata it should probably include a 'hidden: true' boolean flag for cookbooks that are hidden, so that would need to change.  The tarball download endpoint should obviously continue to function identically.

coderanger:
The times I've used gem yanks was because the release was a total non-starter. Like the last one with Chef was "if $HOME is unset it crashed on code load". I don't really see a great reason to support anything beyond a "break glass in case of emergency" solution. Anything less serious should be a new release. If it's bad enough people need to be blocked from upgrading, we probably don't want that version in the wild if possible.

coderanger:
As an example, the linked Supermarket issue about "old versions" is an example where we should fix the tools not allow people to yank, hide, or do anything else on the server side IMO.

lamont-granquist:
It still seems to cause quite a bit of chaos if not done correctly with rubygems.   The github_changelog_generator yanking this morning didn't fix our CI pipelines at all:

https://rubygems.org/gems/github_changelog_generator/versions

And we had to manually work around it with this silliness:

https://github.com/chef/chef-zero/pull/200/files

And 'hide' would also satisfy the use case of totally busted chef-client pushes that were unfit for purpose as well.  If its hidden from depsolving then that fixes all new deploys going forwards.  We probably would have used that instead of yank if it existed for rubygems.

someara:
Here is Rust / Cargo's handing of yanks for comparison

http://doc.crates.io/crates-io.html#cargo-yank

lamont-granquist:
Yeah so the problem there is that rubygems uses "yank" to refer to the thing that tries to delete secrets, while cargo uses "yank" to refer to the thing that just hides the version.  I'd prefer we build the cargo version of "yank" rather than the rubygems version of "yank".  Really, I think there should be both, but the cargo version should be exposed via a web API and command line tools, while the rubygems version takes a supermarket admin to poke it...

coderanger:
:+1: to what @lamont-granquist said.

thommay:
This was approved at the meeting, so take it away @chef/rfc-editors

adamleff:
Rebased and merged in fb58014 - RFC072 accepted!

PR #191, opened 2016-03-02
Community Slack migration proposal

coderanger:
As discussed at the board meeting, the CBGB suggests that the community look in to moving the main public chat resource for the Chef community from IRC to Slack. This document covers that transition, the how, why and why not.

Speaking personally (i.e. taking off my CBGB hat) I think this is probably the right move, but it is going to be a lot of work to do right. I'm especially concerned about the lack of moderation tools and the fact that email addresses are required to be public. The Wordpress community has what seems like a reasonable workaround for the latter with their forwarding mail server, and the former can be fixed with some custom tooling around user signup to enforce IP bans or require linking to an external authentication system (Hosted Chef OAuth or GitHub maybe?). Also for the record, I am in no way signing up to do all this work.

Have at it!

### DIFF
--- a/new/community-slack.md
+++ b/new/community-slack.md
@@ -1,0 +1,170 @@
+---
+RFC: unassigned
+Author: Noah Kantrowitz <noah@coderanger.net>
+Status: Draft
+Type: Informational
+---
+
+# Chef Community Slack
+
+This document describes a plan to migrate the "default" chat for the Chef
+community from IRC to Slack.
+
+## tl;dr
+
+We are creating a public Slack team, name tbd. IRC will continue to exist but
+new users will receive a message from an IRC bot suggesting they ask questions
+in Slack instead, with a link to the SlackIn instance. Existing "web chat" links
+from the Chef website and documentation will point at (or embed) SlackIn instead.
+One channel in the Slack team will be bridged to each of the IRC channels
+(`#chef`, `#chef-hacking`) however this may be removed pending further review of
+how useful it is.
+
+## Personas
+
+Three distinct user stories are being considered. First we have the way many
+people start with IRC, seeking help with a question. We will refer to the user
+in this story as "Alex".
+
+    As a Chef user,
+    I want to get near-real-time support from the Chef community,
+    so that I can learn Chef and debug problems.
+
+Then we have more the answer-er side of things, wanting to help others so that
+they can be successful with Chef. This persona is named "Bonnie".
+
+    As a Chef community member,
+    I want to help other community members,
+    so that I can help the community grow.
+
+And then finally  a more nebulous use case, but no less important. Since day
+one, Chef has thrived on having a strong, welcoming, friendly community. Any
+change to our community tools must maintain this high standard of social
+interaction for the future. This persona is dubbed "Colin".
+
+    As a Chef community member,
+    I want to talk with other community members,
+    so that I can socialize with other people.
+
+These personas are named for use in this document and the surrounding discussion
+to allow easier communication about plans and potential issues.
+
+## The Plan
+
+The first and most obvious step is to register a new Slack team. This will be
+distinct from Chef Software Inc's Slack team as Slack does not offer a
+permissions system which would allow these to co-exist in the same structure.
+
+[SlackIn](https://github.com/rauchg/slackin) will be deployed and mapped to
+`slack.chef.io`. This tool allows users to "sign up" for Slack by requesting an
+invitation email. [Slack-IRC](https://github.com/ekmartin/slack-irc) will be
+deployed to map the two Freenode IRC channels to Slack. An IRC bot will be
+written and deployed to notify new users of the Slack channel and provider them
+a link to SlackIn. This information will also be added to the `/topic` of both
+channels.
+
+Running the new services (SlackIn and the various bots) is left to the discretion
+of Chef Software in their role as stewards of the Chef community.
+
+The new Slack channel will be advertised on the [Chef Forums](https://discourse.chef.io),
+the pending `community.chef.io` landing page, and the Chef documentation.
+
+## Why Slack
+
+IRC has been the backbone of the open-source software community for decades,
+but it is more or less a UI/UX dead end. As the Chef community grows, we are
+having increasing problems keeping our IRC channels a healthy part of it. This
+has two sides. One is that Alex, in most cases, is unfamiliar with IRC and the
+world of clients (web or otherwise), social customs (pastebins etc), and overall
+usage that it entails. This has led to many Alexes unable to find the support
+they need to be successful with Chef. The flip is that the Bonnies of the world
+often don't want to deal with the usability nightmare that is IRC, and thus are
+not able to jump in and answer questions even when they would otherwise have the
+time and energy to do so. This has left the Bonnie role on the shoulders of
+relatively few individuals in an overall unsustainable way. Slack helps both
+Alex and Bonnie to reach their goals with a more modern experience and tools
+they are already familiar with.
+
+## Prior Art
+
+Some other communities have gone through similar transitions recently, with
+mixed results. The most high-profile negative experience was from [Free Code Camp](http://blog.freecodecamp.com/2015/06/so-yeah-we-tried-slack-and-we-deeply-regretted-it.html),
+who quickly hit problems around membership caps. [Reactiflux](https://facebook.github.io/react/blog/2015/10/19/reactiflux-is-moving-to-discord.html)
+had the same problem at around 7,500 users. The `#chef` IRC channel has had
+3,137 distinct usernames in the past year, so we aren't likely to immediately
+hit that kind of cap, but it persistent risk factor going forward.
+
+More positively, [Wordpress](https://make.wordpress.org/chat/) migrated from
+IRC to Slack and has been successful with over 8000 users. The HangOps community
+Slack has also been far more successful than it's prior IRC channel. The Elixir
+and Clojure communities have both started a similar migration, with around 5000
+users each in their Slack teams.
+
+Overall it seems like communities that have done similar IRC to Slack migrations
+have been happy with them, but those hitting hard limits on the Slack side have
+had zero recourse but to abort and move elsewhere.
+
+## Alternatives
+
+For the two problem cases mentioned above, they chose to migrate to [Gitter](https://gitter.im/)
+and [Discord](https://discordapp.com/) respectively. Gitter doesn't have the
+same level of mobile device support and might confuse some users as our
+support discussions generally happen in a single, shared channel rather than
+the smattering of repositories that make up Chef, Chef Server, etc. Discord
+seems reasonable as an alternative, but the market penetration of Slack is a
+strong nod in their favor.
+
+There are a few generally-similar FOSS options
+([Mattermost](http://www.mattermost.org/), [Zulip](https://www.zulip.org/), etc)
+but none seem to be well polished enough to be worth migrating to at this point.
+Additionally most require self-hosting, which would increase the operational
+load on Chef Software's staff.
+
+Another option is to simply stick with IRC, and try to set up a mix of hosted
+bouncers/proxies and an improved web client. While this is possible, I've not
+heard of any other major community providing this level of service for IRC.
+
+## Problems
+
+The biggest problem is mentioned above, unknown and nebulous user limits. We
+will not be butting up against those immediately, but based on IRC usage numbers
+I would expect us to be within striking distance within a year. This could
+probably be solved with some level of pruning of idle users, only 14% of our
+yearly IRC users have been active in the last month, but that brings really
+gross UX with it. Additionally our projections are based on IRC, which could be
+a drastic under-estimation as the improved UX of Slack encourages more users to
+participate.
+
+The second big problem is that Slack requires making your email address used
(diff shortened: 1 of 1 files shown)


lamont-granquist:
:+1: 

I'm one of the "Bonnies".

IRC is down at third on the list of chat clients that I use so it loses and just stays closed until I open it for the thursday meeting.  I'd be about a million times more likely to lurk around on `#chef` if it was on Slack.

new/community-slack.md
```diff
@@ -0,0 +1,170 @@
...
+to allow easier communication about plans and potential issues.
+
+## The Plan
+
+The first and most obvious step is to register a new Slack team. This will be
+distinct from Chef Software Inc's Slack team as Slack does not offer a
+permissions system which would allow these to co-exist in the same structure.
+
+[SlackIn](https://github.com/rauchg/slackin) will be deployed and mapped to
+`slack.chef.io`. This tool allows users to "sign up" for Slack by requesting an
```
jtimberman: Perhaps we may want `chat.chef.io` too?

fabiendelpierre:
I guess I don't really see the point. You did outline some of Slack's advantages compared to IRC, but it just seems like a lot of energy for little benefit. You said:

> Alex, in most cases, is unfamiliar with IRC and the world of clients (web or otherwise), social customs (pastebins etc), and overall usage that it entails. This has led to many Alexes unable to find the support they need to be successful with Chef.

I'm probably biased here (~20-year IRC veteran) but I feel that the learning curve of IRC is trivial. Connecting to a server, joining channels, setting/registering/changing your nickname, using something like Pastebin to show/share code or errors. That's about all you need. Those are not challenging concepts, and there are countless tutorials you can find online through your favorite search engine. I get that's not as user-friendly as Slack, but really... anybody learning or using a tool such as Chef can figure this out. Slack just adds eyecandy and potential headaches for you.

Moderation (or Slack's lack thereof) is sure to be or become a problem for you at some point, as you noted. IRC is enormously better at that. With Slack, a troublemaker might create infinite email addresses and is impossible to get rid of, whereas what you can do with IRC to ban someone is far more annoying for most people to circumvent.

Call me pessimistic but I'll bet you the use of a Slack-IRC bridge will become a problem at some point, leading to Slack becoming the "preferred" platform when the bridge fails, in turn leading to a community divide. Nothing permanent, but a potential annoyance nonetheless. I actually wonder how gracefully such a bridge will handle Slack's formatting markup. By the same token, the moderation tools of IRC are nullified by Slack's lack thereof if and when a troublemaker decides to be a nuisance, until such time that the troublemaker is removed from Slack, or you dynamite the bridge temporarily.

I like IRC because it's a tried-and-true platform. It's open, has dozens of clients, including terminal and/or FOSS clients for those who care, and you don't have to maintain any infrastructure for it since Freenode is there and probably isn't going anywhere. (Granted, Slack is also a hosted service, but if you intend on keeping IRC, you'll need to maintain that bridge.)

So I guess as long as that bridge exists, introducing Slack is fine, but I personally worry that you, as the community "leaders", will start to favor it, i.e. you present Slack as a more accessible and feature-rich method of joining the community. Certainly it'd be a shame if adding Slack led to the removal of IRC. 

In other words, sure, go ahead, add Slack, but be careful not to alienate parts of your community.

I hang out on a number of channels on Freenode and I use Slack at work, so assuming Slack supports being connected to multiple teams at the same time, it won't change much for me, but... yeah, I'm not sure there's tremendous value in going over to Slack.

new/community-slack.md
```diff
@@ -0,0 +1,170 @@
...
+a company structure, so it does not offer any form of public chat or
+user-initiated signup. SlackIn offers this, but the UX is a bit awkward compared
+to normal web application signup.
+
+A fourth general systemic issue is that Slack, as a company, seems to not
+consider this to be a use case worth pursuing. This means they have [stated
+directly](https://www.quora.com/Is-there-a-way-to-ignore-certain-users-in-Slack)
+that they do not plan to offer any community control or moderation features.
+This means no user ignores, no kick/ban abilities, no way to mark trusted users
+(equivalent to op/mod in IRC). Slack also doesn't offer public logs, though they
```
juliandunn: I guess we're implicitly assuming that a) nobody finds the current public logs useful (or consults them frequently enough) and b) that's a feature we're ok giving up?

grimm26:
Just my $0.02, but connecting to slack from work is trivial.  Connecting to IRC, I have to create an ssh tunnel to my server at home and then tunnel IRC through that.  I've never worked anywhere with 6667 outgoing open (just my experience).  Ports 80/443 are trivial to use from anywhere - 6667 not as much.  Yes, there are IRC web clients.  They suck.

new/community-slack.md
coderanger: Or that we would hook up the existing log bot via Slack's IRC gateway? I for one would be really sad if they went away which is why I listed it as a problem we need solve :)

lamont-granquist:
"I'm probably biased here (~20-year IRC veteran) but I feel that the learning curve of IRC is trivial."

I first used IRC in 1990, which I guess makes me a 26-year IRC veteran.  I have to use Slack (and I actually like it better) and I'm not switching back to IRC.  I'm switching more of my personal chats with friends over to slack to get them off of gtalk and facebook.  I shut down the IRC server I used to run probably a decade ago.

coderanger:
For some other food for thought, [Freenode itself is at best in a state of flux from years of "it's working so don't break it" management](https://blog.freenode.net/2016/02/recent-events-and-future-changes/). All their docs and websites are currently directing to that post.

Like  @lamont-granquist I'm probably the closest to a militant pro-IRC zealot on the maintainers and honestly at this point #chef is the last major community I'm part of that hasn't switched off IRC or dissolved into non-existence as people "age out" and stop caring. I sit in a few other social channels, but most are relative ghost towns. When looking for help with something I no longer think "I should ask on IRC". Poking the creator on Twitter is usually a faster route to backchannel support for most projects.

kilasuit:
Being less in touch with Chef community than I would like I would have to say moving across to Slack makes sense to me.

For example there is a vibrant PowerShell Community in the Slack team powershell.slack.com and there is a number of diffferent IRC to Slack Bots that will allow those hesitant to move from IRC to still actively participate.

I for one think it is more likely to improve the community standing and also @fabiendelpierre if you use the Slack App (Most platforms) then you can easily be in multiple teams at the same time (I'm in 8 already for example)

coderanger:
The IRC<->Slack bridges (as opposed to the Slack IRC gateway which is different) do exist and the main one is linked in the RFC but they have generally subpar UX in most IRC clients as all messages on the IRC side show as coming from a single nick (the bot). We should probably run it to catch stragglers asking questions on the IRC side, but the expectation would be that IRC would be a degraded experience because of it. I think that's a reasonable compromise given what we have to work with.

new/community-slack.md
lamont-granquist: I'd vote that its not useful.  The most utility I've gotten out of `#chef` logs is in catching up on the discussion from the past couple of minutes to answer a question, or in some cases to catch up on the past couple of hours when someone has been behaving like they need to get banned.   Whatever happened a week ago is generally not useful.  And any useful content on IRC typically winds up as github issues, stackoverflow questions, or wind up on the mailing list.  When I'm searching on non-chef stuff and IRC logs turn up they're invariably poor quality information compare to stackoverflow.

So yeah, I think IRC logs are mostly digital litter that should be recycled, rather than historical data that needs archiving...

stevendanna: I personally find them very useful for getting backscroll when someone pings me outside of IRC for an issue I might be able to help with.  I also find them useful to find older conversations that talked about a bug that I think I might have more info on.

stevendanna:
Some unordered thoughts:

- The IRC<->Slack bridges I've seen are sub-par enough that we should more or less assume that if our main IRC contributors move to Slack. IRC usage will eventually die out other than as a way for people to learn to move over to slack.

- I am a "Bonnie" in the personas listed.  My usage of #chef has fluctuated substantially.  For a time I was likely one of the most active members.  Now, I rarely log on.  When I think about the reasons, most of them have to do with the varying demands of my personal and professional pursuits and not about the usability of IRC. However, I am fairly confident that a move to Slack would see an uptick in participation from Chef Software, Inc members (at least at the beginning) if only because suddenly the chat will be accessible in the same desktop application that we need to use on a daily basis for work.

- I don't love the idea of moving another communication channel of our community into a closed-source system.  I'd prefer to see our community support open protocols and free software tools.  However, I agree with the post that other than IRC, there are few alternatives that offer the experience of Slack.

Many of my other concerns [0] are handled by the presence of an IRC<->Slack bridge, even if it is sub par, at least in the near term.  If the vote on this were today I would be +0.  Not thrilled by it but I think that our main IRC contributors over the last year or so should be the ones driving the discussion and they seem to be on the side of this RFC.


[0] Mostly the inability to use slack without "signing up" and the fact that for a certain class of user #whatever on Freenode is still their first guess on where to get help on an open source project.

new/community-slack.md
```diff
@@ -0,0 +1,170 @@
...
+There are a few generally-similar FOSS options
+([Mattermost](http://www.mattermost.org/), [Zulip](https://www.zulip.org/), etc)
+but none seem to be well polished enough to be worth migrating to at this point.
+Additionally most require self-hosting, which would increase the operational
+load on Chef Software's staff.
+
+Another option is to simply stick with IRC, and try to set up a mix of hosted
+bouncers/proxies and an improved web client. While this is possible, I've not
+heard of any other major community providing this level of service for IRC.
+
```
Tensibai: Could a StackOverflow chat room be another alternative ? 
The UX is not that bad, there's link preview and code formatting, no mail disclosure is needed and there's a kind of moderation available for room owners. (just my 2cts)

mburns:
As a 'Bonnie' that has greatly benefited from being a lurker that has occasionally provided support to Alex when the opportunity arose, I feel this change would not improve my situation and potentially worsen it. As the Chef community continues to grow, I run the risk of getting kicked out of the chat room due to max user capacity, missing scrollback due to Slack's limits, or becoming a second-class citizen through a IRC <-> Slack bridge if I don't migrate with the community to an application I don't otherwise run.

I'm also not sure how the onboarding UX is improved via the Slack+SlackIn workflow (as mentioned in the Problems section) due to registration, etc, versus a web IRC client (they don't all "suck", though some certainly do) in terms of discoverability or lowering the barrier to entry for new users like Alex. It seems like a wash, particularly for a first-time Alex who asks "drive-by" questions.

If we are willing to run services to improve a Slack migration, I would hope we consider what comparable services (irc bots, web clients that integrate with the community chef site, etc) could be used to improve the situation as-is, without changing protocols.

agh:
I'm in the "Bonnie" camp.  I find Slack to be more 'accessible'. I don't really mean that in the ADA sense, having no experience with screen readers or the like, but rather that from a corporate network using HTTP proxies getting to TCP/6667 (or one of the alternate ports) involves hijinks similar to @grimm26. Getting to TCP/80 or TCP/443 is much easier, and the protocol often matters, e.g. you can't get to an SSH daemon listening on HTTPS through our proxies.

Another point which I'd consider relates to 'accessible' is the much more compelling story with Slack around multi-client persistence and mobile, which promotes participation "anywhere, or anytime". Obviously that brings some complexity, participants will need to configure their notification preferences according to their tastes, but I think it's an overall positive.

I have a story related to the second point: I spent a good amount of time in hospital in 2014 quite suddenly and unexpectedly. Aside from that treatment, I found one of the more jarring aspects to my newfound life away from laptops was losing contact with some people I primarily talk to via IRC. Accessing IRC was *really* hard without a laptop, there aren't many good options on iOS that I found. Somewhat in contrast, Slack was around at that time, and was possible to use, which was huge for me. There's been a groundswell of communities moving to Slack since 2014, and I'd say I now spend more time interacting in various Slack teams than I do IRC, so a big :+1: for this RFC.

jtimberman:
I fit both the Bonnie and Colin personas. 

I've used IRC since 1993. I also am in the top 3 active users on #chef according to @coderanger's IRC logs. Maybe higher, as his logs only go back to 2011, and I've been in #chef since 2009 :grin:.

That said, the only time I open IRC anymore is to join the weekly dev meetings in #chef-hacking. I much prefer Slack for many of the reasons that people have mentioned above. My only concerns about that platform over IRC is the sign up process, as it's "Yet Another Sit Login" to manage. For people who already use slack, it's also "Yet Another Slack Team" (with customizations to be had, since those are team specific per account). This doesn't really bother me, but it may bother others. 

I have no objections, and I am :+1:.

smurawski:
I'm all for it.. consolidating down to the number of "chat" apps is a win for me. :+1:

ranjib:
:+1:

mattstratton:
:+1:

new/community-slack.md
mburns: @lamont-granquist I politely but strongly disagree. Framing them with a negative name isn't particularly helpful, either.

Chat transcripts are useful (to me) for various use cases. Doubly so if we move to Slack.

mattstratton: This might be a possible solution for the issue? 

http://slackarchive.io/

thommay:
This is approved, @chef/rfc-editors . We'll do a follow up around the signup flow.

adamleff:
Rebased/merged in e5f89b67 - all set!

PR #193, opened 2016-03-10
Change Ohai version numbers to match Chef

thommay:
The main intentions are to make it easier to do CI by ensuring that version numbers are simplified and more consistent, and to ensure that dependencies on Chef/Ohai/ChefConfig are clear and well understood.

### DIFF
--- a/new/reversion_ohai.md
+++ b/new/reversion_ohai.md
@@ -20,13 +20,14 @@ numbers were aligned.
 
 ## Specification
 
-Ohai be renumbered to 12.x.0, where x is the same as the current
-development minor version of Chef. Chef would then have a pessimistic version
-pin to Ohai of "~> 12.x.0", and Ohai would have a similar pessimistic
+Ohai will be renumbered to 12.x.y, where x and y are the current development
+minor and patch versions of Chef. Chef would then have a equality version
+pin to Ohai of "= 12.x.y", and Ohai would have a similar equality
 pin to ChefConfig.
 
-When Chef or Ohai update their minor version, the other must do so in
-lockstep.
+The version bot currently used to automatically change version numbers
+for Chef would also be used for Ohai, and any changes to Ohai would
+similarly cause version bumps for Chef and Ohai.
 
 The change from 8 to 12 would not be considered a
 major version bump for API impact considerations, so no deprecated
--- a/new/reversion_ohai.md
+++ b/new/reversion_ohai.md
@@ -1,0 +1,45 @@
+---
+RFC: unassigned
+Author: Thom May <thom@chef.io>
+Status: Draft
+Type: Process
+---
+
+# Change Ohai version numbers to match Chef
+
+In general, Ohai and Chef are released together, and their functionality
+is interdependent. It would be much simpler to reason about how
+functionality added to either would affect the other if their version
+numbers were aligned.
+
+## Motivation
+
+    As a developer,
+    I want to use the correct version of Ohai,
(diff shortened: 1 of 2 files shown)


schisamo:
I know this is more of an implementation detail, but it make sense to collapse the Ohai code base into `chef/chef` like we have done for `chef-config`. This would ensure `rake version` revs chef, chef-config and ohai's version numbers at the same time.

If we do decide to go this route grouping Chef's components in a `src/` directory would help with the overall organization of the `chef/chef` repo. This also follows the pattern we use for other projects like chef-server: https://github.com/chef/chef-server

lamont-granquist:
In the past we've actually taken the liberty of bumping a major version of ohai in a minor dev cycle of chef-client.  This allows us to make more major changes outside of the at least 18 month cycle of chef-client major releases.  This would constrain us to only being able to do that kind of ohai work during the period where we're also working on major version bumps for chef-client

btm:
I'm a little nervous about this. Need to think about why.

There may be situations where we bump the minor version of Chef solely for Ohai. I guess we're okay with that.

lamont-granquist:
If we go down this road I'm :+1: on merging ohai into chef/chef.  It'll make CI/CD much easier.

I'm also insanely :-1: on any scheme which results in us maintaining a compatibility matrix for chef + ohai + chef-config.  We need to be able to release breaking changes to the internal APIs between the three of them so that master of chef and ohai and chef-config may not work at all with last months versions.  And it is this very requirement which is fundamentally driving this PR (the issues caused by ohai having a hard dependency on an unreleased version of chef-config in chef/chef master).

thommay:
totally agree that the compat matrix approach would be a horror show.

mcquin:
Fine with having the ohai version follow chef. Do we need to be concerned with or have a plan for keeping these versions in sync?

-1 to merging repos. It'd be a nightmare from a maintainer wanting to easily find ohai-specific issues/prs.

coderanger:
My vote for how to play this out:

1. Leave the repos separate, too much stuff makes sense to be project-level and the world now defines "project" as a GitHub repo.
2. Make the policy for Ohai that master should be shippable at all times. If we need to we could create a new `release` branch but I don't think that would help anything.
3. Whenever we do a Chef build, it would tag Ohai with the same version as the Chef build.
4. The Chef gemspec would have `s.add_dependency "ohai", s.version`
5. When we release a build of Chef to the stable channel, in addition to the doing the gem push for Chef we also do a gem push for ohai.

I think that covers the major test-matrix-y concerns of this RFC while still allowing some operation as an independent project.

tas50:
I think that merging the repos increases the bloat of chef-client and reduces the usefulness of ohai outside of core chef.  I'd really like to see us not do that.

lamont-granquist:
:+1:

coderanger:
We should decide if the CI bot will be extended to grok that changes in ohai trigger a build (the logical way) or if we wait for the next change in chef (the easy way).

thommay:
I think it would not be super hard for changes in ohai to trigger a bump in chef, too - @sersut ?

btm:
I think we go with the easy way, ohai doesn't trigger builds. Chef is active enough we get daily builds anyway.

btm:
:+1:

adamhjk:
Once the edits from the IRC meeting are landed, this is approved.

thommay:
As per the meeting, I updated RFC 48 to reflect the process. @chef/rfc-editors please do your thing.

PR #194, opened 2016-03-18
Add information on AIX lifecycle

juliandunn:
The motivation here is to capture that we can stop building and supporting AIX 6.1 on April 30, 2017.


coderanger:
One can only hope. /cc @lamont-granquist for approval.

lamont-granquist:
:+1:

PR #196, opened 2016-04-19
Clarify that AIX 7.2 is supported.

juliandunn:
This is to clarify that AIX 7.2 is supported by virtue of binary compat with AIX 7.1 (and 6.1 too, since we actually build on 6.1)


lamont-granquist:
👍

thommay:
I think we should just merge this. 👍

PR #197, opened 2016-04-27
do not deprecate exit code 3

majuscule:
The exit code 3 is useful to prevent service management systems from
marking the service as failed when they are explicitly stopped.

An example using this functionality is here:
https://github.com/chef-cookbooks/chef-client/pull/389

### DIFF
--- a/rfc062-exit-status.md
+++ b/rfc062-exit-status.md
@@ -60,7 +60,7 @@ Exit Code        | Reason             | Details
 42               | Audit Mode Failure | Audit mode failed, but chef converged successfully.
 1                | Failed execution   | Generic error during Chef execution.
 2                | SIGINT received*   | Received an interrupt signal
-3                | SIGTERM received*   | Received an terminate signal
+3                | SIGTERM received   | Received an terminate signal
 -1               | Failed execution*   | Generic error during Chef execution.  On Linux this will show up as 255, on Windows as -1
 
 * \*Next release should deprecate any use of this exit code.
--- a/rfc062-exit-status.md
+++ b/rfc062-exit-status.md
@@ -59,7 +59,7 @@ Exit Code        | Reason             | Details
 0                | Successful run     | Any successful execution of a Chef utility should return this exit code
 42               | Audit Mode Failure | Audit mode failed, but chef converged successfully.
 1                | Failed execution   | Generic error during Chef execution.
-2                | SIGINT received*   | Received an interrupt signal
+2                | SIGINT received    | Received an interrupt signal
 3                | SIGTERM received   | Received an terminate signal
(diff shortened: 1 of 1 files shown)


rfc062-exit-status.md
```diff
@@ -60,7 +60,7 @@ Exit Code        | Reason             | Details
 42               | Audit Mode Failure | Audit mode failed, but chef converged successfully.
 1                | Failed execution   | Generic error during Chef execution.
 2                | SIGINT received*   | Received an interrupt signal
```
lamont-granquist: i kind of think we should keep SIGINT as well.

rhass:
Does keeping those process signals prevent implementing the reboot exit codes in any way? I would imagine a SIGTERM or SIGINT you probably wouldn't want a reboot to happen, but I am just trying to get context to find out if it was deemed necessary to deprecate them to add the reboot signaling.

tbunnyman:
I agree that SIGTERM and SIGINIT should not be depreciated and reboot flags should not override these.

If you SIGINIT the run you wouldn't want it to leap in and signal a reboot. Especially with a SIGTERM, the process needs to die and be done with it, no reboot.

rhass:
@carpnick

carpnick:
So I think SIGTERM a process cannot event catch, so I dont think we would be worried about that.  Same with SIGINT, this is an interrupt, where someone is outright stopping the process.

I think the reboot logic only applies when there is a natural ending of the chef run.  Exception (whether it be bad code or in this case on purpose for reboot purposes), or clean exit.   Technically we could get killed because of out of space or memory as well, and I would say those use cases are undefined since you have an almost hung system at that point.

Reboot should be best effort, if you interrupt a process  (SIGINT or SIGTERM) I think the process should cease immediately, and not try to play games with those signals.  There is usually a good reason for those.

lamont-granquist:
According to `man signal` i think the uncatchable signals are only SIGKILL and SIGSTOP unless I'm reading them wrong...  e.g.

```
     [EINVAL]           An attempt is made to ignore or supply a handler for SIGKILL or SIGSTOP.
```

We already have some goofy handling for SIGINT and SIGTERM around our daemonization that I'd prefer we not touch as part of this RFC (its all fairly subtle and brittle, but absolutely necessary).

tbunnyman:
You can indeed catch these, allowing your process to cleanup after itself before halt.
You should not however abuse this and override.

lamont-granquist:
We already do what we do for $REASONS, this is the wrong RFC to argue about those behaviors.

thommay:
This needs updating for `SIGINT` and is then approved. (Maybe just do that update during merge) @chef/rfc-editors

PR #199, opened 2016-05-17
add some guidance about RFCs, update for reality, prep for scope definition

btm:
I realized listing a bunch of projects that fall under the chef-rfc umbrella would in fact look a lot like a list and should probably be a special low-numbered RFC.

### DIFF
--- a/rfc000-rfc-process.md
+++ b/rfc000-rfc-process.md
@@ -12,6 +12,8 @@ community use RFCs to document, discuss, and plan improvements to Chef and
 other Chef ecosystem tools. You can always find the list of accepted RFCs at
 https://github.com/chef/chef-rfc.
 
+The specific projects covered by this process are detailed in RFC002 (Scope).
+
 ## RFC Types
 
 * A **Standards Track** RFC describes a new feature or improvement for Chef or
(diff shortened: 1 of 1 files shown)


rfc000-rfc-process.md
```diff
@@ -12,6 +12,8 @@ community use RFCs to document, discuss, and plan improvements to Chef and
 other Chef ecosystem tools. You can always find the list of accepted RFCs at
 https://github.com/chef/chef-rfc.
 
+The specific projects covered by this process are detailed in RFC002 (Scope).
```
coderanger: I assume you are writing this now?

btm: next, now is :beer:. 

This didn't feel like the right document for a long list, especially once I started pondering grouping by exceptions. For instance chef-server and test-kitchen have their own maintainers file. Most the RFCs here are applicable to chef-server, but not so much test-kitchen. However I don't want more than one chef-rfc either.

Obviously early feedback welcome.

thommay:
The additional wording around process is much better, thanks!

nathenharvey:
👍

lamont-granquist:
👍

thommay:
I decide this, and also i merge this!

PR #130, opened 2015-05-19
Propose moving OpenBSD 5.7+ i386 and amd64 into Tier 2 support

joemiller:
I'd like to propose moving OpenBSD 5.7+ (i386, amd64 platforms) from `Not Supported` into `Tier 2 support` group.

I have done some work to the current `openbsd_service` resource to support the new `rcctl(8)` daemon control script functionality released in OpenBSD 5.7 as well as implement missing functionality (namely `flags/parameters` support): https://github.com/chef/chef/pull/3372

This PR is a rewrite of the existing `openbsd_service` resource and would not be compatible on OpenBSD < 5.7. It is possible to support two different providers and use ohai to choose the appropriate one based on version, but it may be simpler to only support 5.7+.

Additionally I have initial omnibus builds of the chef client on OpenBSD 5.7 with this PR: https://github.com/chef/omnibus-software/pull/426 .. It probably needs further work in the verification phase, I'm not sure, and could use some feedback from the release engineering team on what would be needed for "good" omnibus builds.

I have been an OpenBSD user for about a decade but only in a personal/hobby sense. I do not use it at work. I'd be willing to spend time to support the basics of OpenBSD for chef, in particular: omnibus builds, package and service providers.

Additional work I have in mind for OpenBSD:

1. Decide on supporting OpenBSD < 5.7 in the `openbsd_service` provider or not.
2. Improve package selection with certain package "flavors" in the `openbsd_package` provider.
3. Complete, approved, omnibus chef builds... Verification tests? Integration into the build pipeline?

Feedback appreciated. thanks.

PR #200, opened 2016-05-17
Drop support for platforms we don't(never) support(ed)

lamont-granquist:
i don't believe we ship omnitruck artifacts for any of these, and that
is the definition we're using for tier 2.

### DIFF
--- a/rfc021-platform-support-policy.md
+++ b/rfc021-platform-support-policy.md
@@ -48,14 +48,10 @@ Ubuntu Linux | 12.04, 14.04 | x86, x86_64 | deb | Ubuntu 12.04
 
 Tier 2 supported platforms are those on which Omnitruck will serve packages, but those packages may not have been built on that OS variant. Additionally, we may or may not do post-build verification on these platforms. A platform must have at least one Maintainer before being added to this tier.
 
-* SUSE Linux Enterprise Server 10, 11
+* SUSE Linux Enterprise Server 11, 12
 * Scientific Linux 5.x, 6.x and 7.x (i386 and x86-64)
-* Gentoo Linux (rolling release)
-* Arch Linux (rolling release)
 * Fedora (current non-EOL revisions)
-* OpenSUSE 13.1/13.2
-* OmniOS stable and LTS releases
-* OpenBSD 5.7/5.8 (i386 and amd64)
+* OpenSUSE 13.1/13.2/42.1
 
 ### Not Supported
 
@@ -65,12 +61,16 @@ Tier 2 supported platforms are those on which Omnitruck will serve packages, but
 * AIX 5.1L
 * FreeBSD 8
 * NetBSD
+* OpenBSD 5.7/5.8 (i386 and amd64)
 * Windows 2003R2, Windows 2003, Windows 2000
 * RHEL/CentOS/Oracle/Scientific 4.x or older
 * RHEL or SLES on POWER (ppc64) or System/z
 * HP-UX
 * Mac OS X < 10.9, anything ppc
 * Debian < 7.0, anything ppc or arm
+* Gentoo Linux (rolling release)
+* Arch Linux (rolling release)
+* OmniOS stable and LTS releases
 
 ## Chef Server
 
--- a/rfc021-platform-support-policy.md
+++ b/rfc021-platform-support-policy.md
@@ -136,6 +136,11 @@ OpenBSD 5.8 | Added February 2016
 openSuse 13.2 | Added February 2016
 Ubuntu 10.04LTS | Removed February 2016 | https://wiki.ubuntu.com/LTS
 Mac OS X 10.8 | Removed February 2016 |
+SUSE Linux Enterprise Server 10 | Removed May 2016 |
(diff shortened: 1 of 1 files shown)


tas50:
OmniOS just uses the solaris omnibus installer. Works fine in Test Kitchen runs

rfc021-platform-support-policy.md
```diff
@@ -48,14 +48,11 @@ Ubuntu Linux | 12.04, 14.04 | x86, x86_64 | deb | Ubuntu 12.04
...
 Tier 2 supported platforms are those on which Omnitruck will serve packages, but those packages may not have been built on that OS variant. Additionally, we may or may not do post-build verification on these platforms. A platform must have at least one Maintainer before being added to this tier.
 
-* SUSE Linux Enterprise Server 10, 11
+* SUSE Linux Enterprise Server 11, 12
 * Scientific Linux 5.x, 6.x and 7.x (i386 and x86-64)
-* Gentoo Linux (rolling release)
-* Arch Linux (rolling release)
 * Fedora (current non-EOL revisions)
-* OpenSUSE 13.1/13.2
+* OpenSUSE 13.1/13.2/42.1
```
btm: Weird. I wonder if anyone is using/testing chef on OpenSUSE Leap 42.1, as I wonder how different it's lsb-release data or whatever is and how Ohai handles that.

rfc021-platform-support-policy.md
```diff
@@ -65,12 +62,15 @@ Tier 2 supported platforms are those on which Omnitruck will serve packages, but
 * AIX 5.1L
 * FreeBSD 8
 * NetBSD
+* OpenBSD (i386 and amd64)
```
btm: https://github.com/chef/chef-rfc/pull/130
https://github.com/chef/omnibus-software/pull/426

The questions on OpenBSD are 1) where do platforms live that are working toward tier 2, tier 1.5? and 2) is this just waiting on Chef Software to either build the packages or provide a way for someone else to do so?

rfc021-platform-support-policy.md
```diff
@@ -65,12 +62,15 @@ Tier 2 supported platforms are those on which Omnitruck will serve packages, but
...
 * Windows 2003R2, Windows 2003, Windows 2000
 * RHEL/CentOS/Oracle/Scientific 4.x or older
 * RHEL or SLES on POWER (ppc64) or System/z
 * HP-UX
 * Mac OS X < 10.9, anything ppc
 * Debian < 7.0, anything ppc or arm
+* Gentoo Linux (rolling release)
+* Arch Linux (rolling release)
```
btm: These aren't mentioned in install.sh, not sure why they were in T2 to start with. :+1:

rfc021-platform-support-policy.md
lamont-granquist: If we're working towards having build pipelines for openbsd then it should be tier 1 IMO.  I think the doc should reflect where we want to get to, not necessarily exactly where we are, which means that plans that are currently in-flight should be included.

We do have a long-standing issue with the fact that we're the gateway to teir 1 status and nobody can contribute external packages.  IMO I don't think we should bikeshed on that problem here.  So if we're not delivering the support for OpenBSD then I think it has to stay off of tier 1.  I would love to see that change, but its currently our harsh reality.

rfc021-platform-support-policy.md
lamont-granquist: @tas50 has looked into it.  I'm pretty sure it fits the standards for this tier.  I think it reports something like "opensuseleap" right now, and that should probably be changed to "opensuse" which is an open issue.  I don't think that means we don't support it, because the packages we ship probably run on it find otherwise.

rfc021-platform-support-policy.md
lamont-granquist: Yeah, I have no idea how those snuck in.

rfc021-platform-support-policy.md
tas50: Since it's a hybrid of Suse and Opensuse we went with identifying it as opensuseleap in ohai. It's still suse platform family though. We don't yet have a bento image so that keeps us from testing cookbooks.

thommay:
👍

rfc021-platform-support-policy.md
```diff
@@ -48,14 +48,11 @@ Ubuntu Linux | 12.04, 14.04 | x86, x86_64 | deb | Ubuntu 12.04
 
 Tier 2 supported platforms are those on which Omnitruck will serve packages, but those packages may not have been built on that OS variant. Additionally, we may or may not do post-build verification on these platforms. A platform must have at least one Maintainer before being added to this tier.
 
-* SUSE Linux Enterprise Server 10, 11
```
nathenharvey: Please update Appendix 2: Timeline of Platform Addition/Removal with the removal of 10.

nathenharvey:
👍

btm:
:+1:

nathenharvey:
@chef/rfc-editors this was approved during the May 28, 2016 meeting.  Please take it from here.

PR #201, opened 2016-05-20
Mode-Agnostic Data Collection

adamleff:
New ability to collect and report run statistics in any Chef Client run (solo, local mode, client-server) to a separate "Data Collector" server.

### DIFF
--- a/new/mode-agnostic-data-collection.md
+++ b/new/mode-agnostic-data-collection.md
@@ -1,0 +1,562 @@
+---
+RFC: unassigned
+Author: Adam Leff <adamleff@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Mode-Agnostic Run Data Collection
+
+In order to provide increased clarity into a user's fleet operations, and to support Chef deployments that do not necessarily include the use of a Chef Server, this change aims to provide an easy and consistent way for the Chef Client to report run statistics to a data collection system. The initial statistics we look to collect include node state, converge start/end times, and converge details.
+
+## Motivation
+
+    As a Chef user who uses both Chef Client Mode and Chef Solo Mode (including the mode commonly known as "Chef Client Local Mode"),
+    I want to be able to collect data about my entire fleet regardless of their client operation type,
+    so that I may better understand the impacts of my changes and may better detect failures.
+
+## Definitions
+
+To eliminate ambiguity and confusion, the following terms are used throughout this RFC:
+
+ * **Chef**: the tool used to automate your system.
+ * **Chef Client Mode**: Chef configured in "client mode" where a Chef Server is used to provide Chef its resources and artifacts
+ * **Chef Solo Mode**: Chef configured in a mode that utilizes a local Chef Zero server. Formerly known as "Chef Client Local Mode" (run as `chef-client --local-mode`) before [RFC 031](https://github.com/chef/chef-rfc/blob/master/rfc031-replace-solo-with-local-mode.md) was implemented, available in Chef version 12.10.54 and later.
+ * **Chef Solo Legacy Mode**: Chef in the former Solo operational mode (run as `chef-solo`) before [RFC 031](https://github.com/chef/chef-rfc/blob/master/rfc031-replace-solo-with-local-mode.md) was implemented (in Chef versions earlier than 12.10.54), or Chef run as `chef-solo --legacy-mode` in Chef version 12.10.54 and later.
+
+## Specification
+
+Similar to how data is collected and reported for Chef Reporting, we expect to implement a new EventDispatch class/instance that collects data about the Chef run and reports it accordingly. Unlike Chef Reporting, the server that receives this data is **not** running on the Chef Server, allowing users to utilize this function whether they use Chef Server or not. No new data collection methods are expected to be implemented as a result of this change; this change serves to implement a generic way to report the collected data in a "webhook-like" fashion to a non-Chef-Server receiver.
+
+The implementation must work with Chef running in any mode:
+
+ * Chef Client Mode
+ * Chef Solo Mode
+ * Chef Solo Legacy Mode
+
+### Protocol and Authentication
+
+All payloads will be sent to the Data Collector server via HTTP POST to the URL specified in the `data_collector_server_url` configuration parameter. Users should be encouraged to use a TLS-protected endpoint.
+
+For the initial implementation, transmissions to the Data Collector server can optionally be authenticated with the use of a pre-shared token which will be sent in a HTTP header. Given that the receiver is not the Chef Server, existing methods of using a Chef `client` key to authenticate the request are unavailable.
+
+### Configuration
+
+The configuration required for this new functionality can be placed in the client.rb or any other `Chef::Config`-supported location (such as a client.d or solo.d directory).
+
+#### Parameters
+
+ * **data_collector_server_url**: required. The full URL to the data collector server API. All messages will be POST'd to this URL. The Data Collector class will be registered and enabled if this config parameter is specified.
+ * **data_collector_token**: optional. A pre-shared token that, if present, will be passed as an HTTP header named `x-data-collector-token` to the Data Collector server. The server can choose to accept or reject the data posted based on the token or lack thereof.
+ * **data_collector_mode**: The Chef mode in which the Data Collector will be enabled. For example, you may wish to only enable the Data Collector when running in Chef Solo Mode. Must be one of: `:solo`, `:client`, or `:both`. The `:solo` value is used for Chef operating in Chef Solo Mode or Chef Solo Legacy Mode. Default: `:both`.
+ * **data_collector_raise_on_failure**: If true, the Chef run will fatally exit if it is unable to successfully POST to the Data Collector server. Default: `false`
+
+### Schemas
+
+For the initial implementation, three JSON schemas will be utilized.
+
+#### Action Schema
+
+The Action Schema is used to notify when a Chef object changes. In our case, the primary use will be to update the Data Collector server with the current node object.
+
+```json
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Data Collector - action schema",
+  "properties": {
+    "entity_name": {
+      "description": "The name of the entity",
+      "type": "string"
+    },
+    "entity_type": {
+      "description": "The type of the entity",
+      "type": "string",
+      "enum": [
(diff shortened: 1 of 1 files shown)


coderanger:
Do you have a prototype implementation in an external gem/script? That's where I would start at least.

coderanger:
Also hugely 👎 on putting this in core. At most it should be an extra gem that is shipped in the chef-client installers. We already have 2 different systems for event handling and 3 different plugins like this in varying states of goodness.

adamhjk:
The intent here is to wind up with a single, simpler implementation. The current reporting mechanisms are tightly coupled to the Chef Server API - we want something that is much closer to a simple webhook, which enables not just our software but other peoples to ingest the data easily. While I respect your opinion about where it will live, I think it belongs in core chef, and will make everyones lives easier moving forward.

coderanger:
Maybe eventually, but if this is going to be coupled with the development of a new server-side component to receive the data (and if it isn't, then this is even weirder but I think this is related to the new Insights work?) it's probably going to be under heavy churn for a while. If you want to commit to keeping to this spec then fine, but that seems like a big ask for the server-side team and both updating this document for every schema change and doing the client-side work in Chef (meaning that it can't be tested except by running Chef nightlies) seems like it isn't worth it.

lamont-granquist:
So its entirely opaque to me what this achieves that the existing standard for the ResourceReporter in the chef-client lacks.

Its also written sounding like its going to be a new class bolted onto the Chef client without bothering to address migrating the existing ResourceReporter API over to it.  That should definitely be addressed in the text of the RFC because we're now into the third or fourth iteration of people who don't normally touch the client looking at its reporting and finding it lacking (no argument) and then bolting something on to the side of it.  That problem needs to be addressed up front.  I'd expect to see some language around using the server versioning API to handle the start of run and end of run communication where the next "V N+1" in the server API would support the new schema.

What's also lacking here is a prose description of the motivation.  If we just change the on-wire JSON format of the pre-run and post-run communication that isn't very compelling.  I'm assuming there's something else here, but I'm not clearly seeing it, and the point is probably getting buried somewhere in the middle of the JSONSchema rather than highlighted by it.

And as a concrete example of the text that I object to:

```
Similar to how data is collected and reported for Chef Reporting, we expect to implement a new EventDispatch class/instance that collects data about the Chef Client run and reports it accordingly
```

If the intent is to wind up with a single simpler implementation, then that text needs to be changed, because that text reads like creating a brand new thing to bolt onto the client right next to the ResourceReporter.

I would like to see something more like:

```
The ResourceReporter class will be replaced by a new EventDispatch class.  Its will fundamentally differ from the ResourceReporter in that it will hook into...  <some kind of new hawtness> ...."  The initial run_started POST request will be versioned...
```

Actually the initial run_started POST request is initiated by the client before it has any information about anything, before it does the node GET request, so there's no way for it to know about the version of the server that it is talking to.  That's a fairly big issue to address how the protocol downgrade happens.  I'd like to see that sketched out along with what the responsibilities for the classes are going to be along with when Chef::ResourceReporter gets put out its misery.

And there is some precedent for all of this in the Chef::User and Chef::UserV1 classes.   But the important point there is that depending on which version of chef server you're talking to, only one of those classes gets used, and we're eventually going to remove the OSC11 backcompat one.

lamont-granquist:
Also if the idea is to eventually support something like having updates pushed to the server on every change of a resource then it should really get spelled out clearly.   There's nothing here that isn't in the resource reporter, so I suspect there's more.

adamhjk:
Two parts here. One is having an endpoint that is more generic than the reporting endpoint, usable as a simple web hook. The other is schema change.  

Lets talk about the endpoint first, as it's part one, and the most important. The current reporting API is tightly coupled to the Chef Server API, making it much more difficult to get useful reporting data out of Chef Solo environments (no authentication). One part of this addresses that - we wind up with essentially a web-hook with multiple payloads. Since the existing reporting API, well, exists, it's something we would not remove from the client without a major revision, and feels like injecting needless breakage. Yes, we will ship an implementation that ingests the data from that web-hook - I expect others will too.

Second, the schema. I honestly have less energy about the specifics here. Obviously it should be versioned, but it should be in the payload (as it's a web-hook, it needs to be in the payload rather than the headers, for example). It also means delivery of multiple entity types to a single endpoint (much like the github webhook distributor does).

So that's the story. 

1) Solo gains the ability to send statistics to an easily implementable REST API that's authed with a token, rather than the Chef Server API.

2) The payloads are different, because webhooks need more context, since by definition the event is the thing, vs the Reporting API, where the server endpoint encoded entity-type.

We'll support both APIs for some (probably long) period of time.

new/mode-agnostic-data-collection.md
```diff
@@ -0,0 +1,549 @@
...
+
+## Motivation
+
+    As a Chef user who uses both Chef Server and Chef Client local mode,
+    I want to be able to collect data about my entire fleet regardless of their client operation type,
+    so that I may better understand the impacts of my changes and may better detect failures.
+
+## Specification
+
+Similar to how data is collected and reported for Chef Reporting, we expect to implement a new EventDispatch class/instance that collects data about the Chef Client run and reports it accordingly. Unlike Chef Reporting, the server that receives this data is **not** running on the Chef Server, allowing users to utilize this function whether they use Chef Server or not.
```
lamont-granquist: It would be good to spell out here that since we're on solo we're not using chef-server authentication, so that requires a fundamental change.  Its sort of implicit and should be as easy as "2+2=4" to figure out, but my brain went running down other avenues and entirely missed it.

adamhjk:
This isn't a new way of gathering statistics internally, either - it's just about schema and wire protocol.

adamleff:
Thank you for the conversation and feedback, gentlemen. I've updated the text of the RFC to offer some more clarity on the motivation and addressed some specific points brought up in this conversation, and spelled out some potential future work in an effort to eliminate technical debt.

new/mode-agnostic-data-collection.md
```diff
@@ -0,0 +1,559 @@
...
+Type: Standards Track
+---
+
+# Mode-Agnostic Run Data Collection
+
+In order to provide increased clarity into a user's fleet operations, and to support Chef deployments that do not necessarily include the use of a Chef Server, this change aims to provide an easy and consistent way for the Chef Client to report run statistics to a data collection system. The initial statistics we look to collect include node state, converge start/end times, and converge details.
+
+## Motivation
+
+    As a Chef user who uses both Chef Server and Chef Client local mode,
```
nathenharvey: I recommend "chef-solo" instead of "Chef Client local mode"

new/mode-agnostic-data-collection.md
```diff
@@ -0,0 +1,559 @@
...
+    "user_agent"
+  ],
+  "dependencies": {
+    "parent_name": ["parent_type"],
+    "parent_type": ["parent_name"]
+  },
+  "title": "ActionSchema",
+  "type": "object"
+}```
+
```
nathenharvey: GHFMD isn't treating this the way you intended.  May need a line break before the ```

coderanger:
Looks like work on this was started before this RFC was filed, just in case people want to see a concrete implementation https://github.com/chef/chef/compare/adamleff/data_collector

new/mode-agnostic-data-collection.md
```diff
@@ -0,0 +1,560 @@
+---
+RFC: unassigned
+Author: Adam Leff <adamleff@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Mode-Agnostic Run Data Collection
+
+In order to provide increased clarity into a user's fleet operations, and to support Chef deployments that do not necessarily include the use of a Chef Server, this change aims to provide an easy and consistent way for the Chef Client to report run statistics to a data collection system. The initial statistics we look to collect include node state, converge start/end times, and converge details.
```
btm: We should better differentiate between the tool, **Chef** and the modes, e.g. **Chef Client** and **Chef Solo**. There's a clear bias in this document for the world in which Chef is a client, which reads strange in some examples here.

Can we agree on some definitions, maybe put them somewhere?

Chef - The tool you use to automate your system.
Chef Client [Mode] - when using Chef in client mode
Chef Solo [Mode] - when using Chef in solo Mode in >= 12.10.54 / post RFC 031
Chef Solo Legacy Mode - when using Chef in solo mode in < 12.10.54 / pre RFC 031, or using `--legacy-mode`.

Chef Local Mode - Mostly the same as Chef Solo Mode in 12.10.54 / post RFC 031?

new/mode-agnostic-data-collection.md
```diff
@@ -0,0 +1,554 @@
...
+
+ * Chef Client Mode
+ * Chef Solo Mode
+ * Chef Solo Legacy Mode
+
+### Protocol and Authentication
+
+All payloads will be sent to the Data Collector server via HTTP POST to the URL specified in the `data_collector_server_url` configuration parameter. Users should be encouraged to use a TLS-protected endpoint.
+
+For the initial implementation, transmissions to the Data Collector server can optionally be authenticated with the use of a pre-shared token which will be sent in a HTTP header. Given that the receiver is not the Chef Server, existing methods of using a Chef `client` key to authenticate the request are unavailable.
```
lamont-granquist: the pre-shared http header key concerns me a bit.  might be better to leverage oauth or something even though its more complicated?

lamont-granquist: and i think paging @stevendanna to see if i'm worrying about this for nothing or not...

stevendanna: What is the goal of this token?  Assuming it is a single shared token, you don't get much in the way of authentication (since any node could pretend to be any other) but you do at least get a bit of authorization (only token holders can make requests).  If you plan on having a token-per-node then some public key encryption feels like it would make more sense since you already have to manage N tokens.

adamhjk:
If it's run like a web hook, this should be a shared secret that gets hashed based on the payload.

adamleff:
This was conditionally approved at the 5/26/16 community IRC meeting:

```
16:39 < nathenharvey> holoway: weigh in here?  approved with the mentioned edits?  or should we revisit next week after the edits are in place?
16:39 < holoway> approved with the mentioned edits, but wait on merging until the schema is final
16:40 < holoway> at which point just update the RFC and merge
```

adamleff:
Schemas finalized, versions set to 1.0.0, PR accepted as RFC 77.

PR #204, opened 2016-06-02
Update RFC021 to clarify support

btm:
Support means many things to different groups, users, customers, developers.

This RFC is about builds and testing on different platforms and platform versions, so we let's reduce the scope of the phrase support here.

### DIFF
--- a/rfc021-platform-policy.md
+++ b/rfc021-platform-policy.md
@@ -6,32 +6,30 @@ Type: Informational
 ---
 
 
-# Chef Platform Support Policy
+# Chef Platform Policy
 
 The purpose of this RFC is to clarify:
 
-* What specific operating system platforms and platform versions are supported by the software produced by Chef Software, Inc.
-* What is the meaning of "supported platform"
-
-This RFC does *not* address "What is the product lifecycle of Chef Software, Inc.'s software". That is covered in a separate RFC.
+* What specific operating system platforms and platform versions will work with the software produced by Chef Software, Inc.
+* What are the meanings of "foundational platform", "secondary platform", and "other platforms"
 
 ## Chef Client
 
-A Chef Client supported platform means:
+To be included as either a Foundational Platform or a Secondary Platform, the following must be true for that specific platform and version:
 
 * Omnitruck won't fail when confronted with the platform and version
 * The most important core resources (package, service, template) work out of the box
 * Ohai attributes for ```platform```, ```platform_family```, ```platform_version``` and ```kernel.machine``` are correct
 
-Chef Client support policies also apply to Ohai, since that is a dependency.
+These policies also apply to Ohai, since that is a dependency.
 
-### Tier 1 Support
+### Foundational Platforms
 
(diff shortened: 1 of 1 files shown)


rfc021-platform-policy.md
```diff
@@ -6,32 +6,30 @@ Type: Informational
...
+Foundational platforms are those for which native binary "Omnibus" (full-stack installer) packages are built and available via Omnitruck for every release. For each platform, Chef performs some post-build verification on them or their equivalents. For example, we may elect to do post-build verification for Oracle Enterprise Linux using the same test results as Red Hat Enterprise Linux, since they are so similar. Only platform versions that receive automated post-build verification are listed here, even when newer or intermediate versions may be known to work. A platform must have a Lieutenant as defined in RFC 030 before being added to this tier.
 
 Platform | Versions | Architectures | Package Format | Built on
 --- | --- | --- | --- | ---
-AIX | 6.1, 7.1, 7.2 | ppc64 | bff | AIX 6.1
+AIX | 6.1, 7.1 | ppc64 | bff | AIX 6.1
```
scotthain: We should put AIX 7.2 in the Secondary Platforms section

btm: Is there anything special we do for 7.2, or does it just happen to work?

rfc021-platform-policy.md
```diff
@@ -76,38 +74,37 @@ Tier 2 supported platforms are those on which Omnitruck will serve packages, but
...
+###  Other or non-viable
 
 * Windows Vista, XP, 2000
 * Mac OS X < 10.9, anything ppc
 
-## Appendix 1: Guiding Principles for Operating System Version Support
+## Appendix 1: Guiding Principles for Operating System Version Adoption
 
-Once Chef Software, Inc. decides to support an operating system, we will also develop rules to determine under what upstream vendor lifecycle we will continue to support products, and they will be documented in this section. Vendors have various terminology to describe support lifecycles ('standard support', 'extended support', etc.) and it is useful to clarify what those mean in the context of Chef's products.
+Once Chef Software, Inc. decides to adopt an operating system, we will also develop rules to determine under what upstream vendor lifecycle we will continue to build and/or test on our products on that platform, and they will be documented in this section. Vendors have various terminology to describe support lifecycles ('standard support', 'extended support', etc.) and it is useful to clarify what those mean in the context of Chef's products.
```
scotthain: This sentence is kinda awkward, maybe rephrase as something like `...we will also develop rules to determine what upstream vendor lifecycle applies, and will build and/or test our products according to these rules. These rules will be documented in this section.`

I don't know if that muddies up the underlying meaning, but it seems more clear to me.

thommay:
I'm not in love with the names but I can't think of better ones, so 👍

rfc021-platform-policy.md
thommay: think it should be "build and/or test our products" (drop the `on`).

rfc021-platform-policy.md
scotthain: Nothing special, it just works. It's different enough from 7.1 that it's worth mentioning (IBM has a big thing about "forward binary compatibility" on AIX).

thommay:
The discussion at the 2016-06-02 IRC meeting left us wondering if an RFC was the most appropriate place for this document.

btm:
We're currently planning on deprecating/replacing this RFC and making https://docs.chef.io/supported_platforms.html the definitive guide to which platforms/versions:

1) Have builds that are believed to work + are tested
2) Have builds and are believed to work
3) Are believed to work

@thommay is planning to open a separate PR. Feedback on this can be dropped here until then.

btm:
https://github.com/chef/chef-rfc/pull/206

PR #198, opened 2016-05-17
add Ubuntu 16.04

lamont-granquist:

PR #206, opened 2016-06-15
Move platform support policy to docs.chef.io

thommay:


thommay:
@josephrdsmith please review

josephrdsmith:
@juliandunn

lamont-granquist:
closes #198?

rfc021-platform-support-policy.md
```diff
@@ -5,109 +5,17 @@ Status: Accepted
...
 
-# Chef Platform Support Policy
-
-The purpose of this RFC is to clarify:
-
-* What specific operating system platforms and platform versions are supported by the software produced by Chef Software, Inc.
-* What is the meaning of "supported platform"
+The purpose of this RFC is to clarify how Chef Software, Inc choose to adopt or remove a platform.
 
 This RFC does *not* address "What is the product lifecycle of Chef Software, Inc.'s software". That is covered in a separate RFC.
```
btm: Remove "That is covered in a separate RFC" as I don't believe that RFC exists.

rfc021-platform-support-policy.md
```diff
@@ -5,109 +5,17 @@ Status: Accepted
...
 
+# Chef Platform Policy
 
-# Chef Platform Support Policy
-
-The purpose of this RFC is to clarify:
-
-* What specific operating system platforms and platform versions are supported by the software produced by Chef Software, Inc.
-* What is the meaning of "supported platform"
+The purpose of this RFC is to clarify how Chef Software, Inc choose to adopt or remove a platform.
```
btm: *chooses

btm:
Not thrilled with "adopt" as we don't do anything that seems like adopting?

Could we find a word like 'validated' or 'tested' or 'verified'?

juliandunn:
Overall 👍 modulo the copy-edits that others have provided.

coderanger:
I don't think I agree with this. There is a world of platform support outside of what Chef Software chooses to provide and we should be open to that as a project. Chef Software can choose how to allocate engineering time, but what the project declares it's desired support policy to be is independent of the company's support policy.

btm:
The bulk of this RFC has always been a public reference of what happens in
CI and gets pushed to Omnitruck. I don't really care of that goes here or
there, but for the foreseeable future that is something only Chef Software
can change.

You're talking about what we think Chef works on, omnibus packaging aside.
That list can still be here.

coderanger:
What we decide to support should be a proactive decision, and the right place for those seems like an RFC to me. Maybe if you explained the motivation to move this to another document?

btm:
I'll let @thommay defend that, I don't feel strongly about how we split it either way, but there are already two documents. The goal is to clarify what exactly this document intends to assure, and who provides that assurance.

coderanger:
Basically I don't think this document should exist as written, what Chef Software Inc wants to support is up to them and can be documented on their website. This should be distinct (though likely a subset) of what Chef as a project wants to support and how we define "supported". This change is going to far towards crossing the streams of project and company.

coderanger:
(though I realize I might be the only human being on the planet that really cares about this distinction O:-)

rfc021-platform-support-policy.md
```diff
@@ -5,109 +5,17 @@ Status: Accepted
...
-* Mac OS X 10.9, 10.10, 10.11
-* Ubuntu 12.04, 14.04
-
-ChefDK bundles Chef Client. Therefore, Chef Client is supported, by extension, on the foregoing client platforms, if not already mentioned explicitly in the Chef Client support matrix.
-
-### Unsupported
-
-* Windows Vista, XP, 2000
-* Mac OS X < 10.9, anything ppc
+The [chef docs site](https://docs.chef.io/) has [details of current adopted platforms](https://docs.chef.io/platforms.html).
```
btm: I'd say something like: 

The Chef Docs site contains the list of platforms that Chef Software currently builds packages for and tests.

The list of working platforms maintained by the Chef community is here:

blah blah blah raspbian blah blah.

thommay:
So I guess the question is - should we actually define here what platforms are compliant with the language? And say that adding a new platform requires becoming an LT and updating this list?

rfc021-platform-support-policy.md
```diff
@@ -7,48 +7,25 @@ Type: Informational
...
-OpenBSD 5.8 | Added February 2016
-openSuse 13.2 | Added February 2016
-Ubuntu 10.04LTS | Removed February 2016 | https://wiki.ubuntu.com/LTS
-Mac OS X 10.8 | Removed February 2016 |
-SUSE Linux Enterprise Server 10 | Removed May 2016 |
-Gentoo Linux | Moved to Tier 3 May 2016 | This was never adopted by the Tier 2 definition
-Arch Linux | Moved to Tier 3 May 2016 | This was never adopted by the Tier 2 definition
-OpenSUSE 42.1 | Added to Tier 2 May 2016 |
-OpenBSD 5.7/5.8 | Moved to Tier 3 May 2016 | This was never adopted by the Tier 2 definition
+The purpose of this RFC is to clarify which platforms the chef community chooses to work on, and how those platforms are maintained.
```
btm: *Chef

rfc021-platform-support-policy.md
```diff
@@ -7,48 +7,25 @@ Type: Informational
...
+
+This RFC does *not* address "What is the product lifecycle of Chef Software, Inc.'s software".
+
+The [chef docs site](https://docs.chef.io/) has [details of the platforms Chef Software, Inc validates for release](https://docs.chef.io/platforms.html).
+
+# Adding a new platform
+
+Community maintenance of a platform requires that core chef resources,
+such as `package`, MUST have support for the platform. For example, on Debian
```
btm: ... including `package`, `service`, and `file`, MUST have working providers for the platform.

PR #207, opened 2016-06-27
Added myself as Reddit Advocate

jjasghar:
- I created the subreddit, and have moderator ability.

### DIFF
--- a/ADVOCATES.md
+++ b/ADVOCATES.md
@@ -63,4 +63,5 @@ More information can be found in the [Chef Community Guidelines](https://github.
 * Vacant - Submit a PR to become a Chef Community Advocate
 
 #### Chef_Opscode Reddit ([https://www.reddit.com/r/chef_opscode/](https://www.reddit.com/r/chef_opscode/))
-* Vacant - Submit a PR to become a Chef Community Advocate
+* [JJ Asghar](https://github.com/jjasghar)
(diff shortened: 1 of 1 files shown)


adamleff:
:+1:

lamont-granquist:
👍

btm:
RFC020 specifies a majority of existing advocates, but isn't clear about how we handle new areas of advocating. That said, JJ is already an advoate, this is just a new area. We got +1's from three existing advocates which is about half, and everyone else in the irc meeting was in favor including some maintainers.

That'll do. thanks @jjasghar.

ADVOCATES.md
```diff
@@ -63,4 +63,5 @@ More information can be found in the [Chef Community Guidelines](https://github.
 * Vacant - Submit a PR to become a Chef Community Advocate
 
 #### Chef_Opscode Reddit ([https://www.reddit.com/r/chef_opscode/](https://www.reddit.com/r/chef_opscode/))
```
lamont-granquist: ^ that should probably be renamed to something non-opscode, but r/chef obviously doesn't work and i have no better idea...

jjasghar: Yep, I'd love to, but I don't know if it's possible, i chose `/r/chef_opscode` three years ago because `/r/chef` private.

PR #209, opened 2016-06-29
Supermarket Prefix Reservation

coderanger:
A short RFC to document and formalize what I think is already common practice.

### DIFF
--- a/new/supermarket-prefix.md
+++ b/new/supermarket-prefix.md
@@ -33,6 +33,16 @@ to if a cookbook is officially associated with the prefix name or not.
 
 ## Specification
 
+A group or company will have exclusive control to prefixes registered to them.
+These prefixes are subject to approval by Chef Software as part of their operation
+of the public Supermarket application. Private deployments of Supermarket can
+define their own governance structure for allocating prefixes.
+
+If a prefix is registered, the user or users that are responsible for it will
+be the only ones allowed to register new cookbooks or tool with that prefix.
+The responsible user or users can authorize other users to create cookbooks in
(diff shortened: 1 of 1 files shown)


new/supermarket-prefix.md
```diff
@@ -0,0 +1,63 @@
...
+The responsible user or users can authorize other users to create cookbooks in
+their prefix, but this requires explicit permission from the prefix holder.
+
+In the short term, prefix management can be handled ad hoc via existing cookbook
+rename/removal tooling in Supermarket. Established prefixes can be added to this
+RFC by pull request.
+
+In the long term, it would be good to see native functionality added to
+Supermarket to enable this, though care will have to be taken at that point to
+ensure abuse is not a problem.
```
lamont-granquist: we might want to specify something like a prefix is at least two letters followed by an underscore, and that when registering a prefix there shall be no cookbooks not owned by the registeree that exist with that prefix or something.

in the case of someone registering a prefix after someone else has already trampled on the prefix that would require some kind of administrative action.

that way nobody can register compat_ (compat_resource) or chef_ (chef_hostname).

lamont-granquist:
:+1: fine with not doing the future thing in this RFC

btm:
as this week's decider, this is accepted contingent on @coderanger adding a little more information on how to contact the supermarket admins in case of any namespace encroachment.

adamleff:
👍

adamleff:
Accepted/assigned RFC 078 in 779cb6f. Thanks!

PR #210, opened 2016-07-14
Add draft supermarket release RFC

danielsdeleo:
🚧 🚧 🚧 🚧 🚧 🚧 🚧 🚧 🚧 🚧 🚧 🚧 🚧 🚧 🚧 🚧 🚧 

Add RFC for release channels in supermarket. There is still a lot of implementation detail to be decided.

### DIFF
--- a/new/supermarket-release-channels.md
+++ b/new/supermarket-release-channels.md
@@ -72,10 +72,10 @@ remain available after a newer version is promoted to that channel.
 There are some use cases where it is desirable to only allow one version of an
 artifact in a channel. Users of Chef Automate, for example, are encouraged to
 only use one version of an artifact in a given environment. Thefore, the
-user may specify that an artifact (TODO: is this config on the channel, or an
-option passed in to the promotion API call?) is _exclusive._ When this option
+user may specify that an artifact is _exclusive._ When this option
 is active, promoting a version of an artifact to a channel removes all other
-versions from that channel.
+versions from that channel. The exact mechanism by which this happens, e.g.,
+configuration or API call, will be determined when implemented.
 
 #### Filtering Artifacts by Channel
 
--- a/new/supermarket-release-channels.md
+++ b/new/supermarket-release-channels.md
@@ -1,0 +1,137 @@
+---
+RFC: unassigned
+Author: Daniel DeLeo <dan@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Supermarket Release Channels
+
+Add the ability to assign supermarket artifacts (cookbook versions) to a named
+channel, and to filter the universe endpoint data by specifying a channel name.
+
+## Motivation
+
+    As a Supermarket User,
+    I want to promote my artifacts through named channels,
+    so that I can release them according to my testing and release policy.
+
+Many users implement a testing pipeline to automate testing and release of chef
+artifacts. An example of this kind of system is the workflow component of the
+Chef Automate product. Frequently it is necessary or convenient to publish an
+artifact to a Supermarket service at an early pipeline stage, so that the
+artifact can be distributed to other systems for testing. For example, when
+using Chef Automate, the default behavior is to publish the artifact in the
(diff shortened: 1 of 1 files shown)


coderanger:
Just for the sake of having the discussion, another option would be to do this like Rubygems and PyPI. No explicit channels, just a single set of versions and an option to install "prereleases" or not. Might be less work on some fronts, but Chef doesn't support those version number addons so probably just a non-starter.

danielsdeleo:
The downside of that is you have to re-package the artifact to change the version number. And it would probably end up being more work, as you say.

schisamo:
For consistency's sake would be OK to use the naming of `view` instead of `channel`? This is the naming the Habitat Depot uses for the same sort of functionality:
https://www.habitat.sh/docs/continuous-deployment-overview/

/cc @juliandunn

danielsdeleo:
I have no attachment to "channels," it's just what I'm most familiar with. "view" would be fine as long as there's a consensus that it's grokable

tyler-ball:
I am okay with switching from `channel` to `view`.

@danielsdeleo Are you purposefully not defining the available (like `unstable`, `current` and `stable`) views in this RFC? It seems like that is an important detail. LMK if that is more of an implementation detail - I'm out of date on RFC scoping.

new/supermarket-release-channels.md
```diff
@@ -0,0 +1,93 @@
...
+
+#### Artifact Versions and Channels
+
+Channels may have any number of versions of an artifact, so that old versions
+remain available after a newer version is promoted to that channel.
+
+There are some use cases where it is desirable to only allow one version of an
+artifact in a channel. Users of Chef Automate, for example, are encouraged to
+only use one version of an artifact in a given environment. Thefore, the
+user may specify that an artifact (TODO: is this config on the channel, or an
```
tyler-ball: Exclusivity seems weird to me. I understand the idea behind it, but I don't know of an existing artifact storage system that implements this. I am familiar with the idea that "you get the latest version of the requested artifact unless you specify a version constraint".

Thinking more about it, I can understand the use case in a private supermarket. I don't have enough context around private supermarket use cases to answer this TODO. But the simplest way to support this feature seems to be to specify exclusivity on a channel level, so thats what I would vote for if we do not know of another customer use case.

new/supermarket-release-channels.md
```diff
@@ -0,0 +1,93 @@
...
+versions from that channel.
+
+#### Filtering Artifacts by Channel
+
+Filtering is implemented by modifying the response from the `/universe`
+endpoint to include only the artifacts in the requested channel.
+
+TODO: figure out how this works.
+
+For compatibility, there must be a mode of operation where Supermarket works
```
tyler-ball: I like the idea that the Supermarket administrator specifies the 'default' channel. And when upgrading Supermarket to the version that supports Views, it defaults the 'default' channel to the one containing all the existing artifacts.

danielsdeleo:
@tyler-ball

danielsdeleo:
@tyler-ball 

> Are you purposefully not defining the available (like unstable, current and stable) views in this RFC? 

Yeah, I think they need to be user-selectable. I think that if you're using the Chef Automate workflow, you'll probably map the view to the Automate Workflow environments, but if you're using your own Ci pipeline, you may do something that maps more closely to that.

new/supermarket-release-channels.md
danielsdeleo: My primary motivation for this is to make private supermarket usage with Chef Automate make more sense, but still have it be useful for private supermarket usage with a bespoke Ci pipeline. I haven't really figured out what benefit there would be for the public supermarket or how it would be configured.

thommay:
I'm 👍

thommay:
approved, @chef/rfc-editors

nathenharvey:

thommay:
Accepted and merged as RFC 79

PR #211, opened 2016-07-28
update Ruby RFC to be more general

lamont-granquist:
The practical background here:

Pinning activesupport to `~> 4.0` and rack to `~> 1.0` is creating a fairly substantial workload to run around and touch every gem in our world and update them all and we need to be able to drop support for ruby 2.1.0 entirely.  This problem is only likely to get worse, it certainly won't get any better.

The bikeshed of what ruby version we should ship in the packages and/or support is not something that be discussed in this RFC.  If anyone cares to hear about why that is a hard problem we can have an after-meeting about it (TL;DR is that it's very easy to write a policy we should follow when you're not the one who has to do the work to bludgeon ruby into submission in our windows, solaris and aix omnibus build environments).

The filename of this RFC should probably change to reflect the changed scope.

### DIFF
--- a/rfc034-ruby-193-eol.md
+++ b/rfc034-ruby-193-eol.md
@@ -5,9 +5,10 @@ Status: Accepted
 Type: Standards Track
 ---
 
-# Ruby 1.9.3 EOL
+# Ruby Version Deprecation Policy
 
-Chef 12.0 will require Ruby 2.0 or greater.
+Establishes a rolling version policy for Chef and ChefDK and the
+assocated ecosystem (berkhelf, test-kitchen, etC).
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


danielsdeleo:
Only question for me is if this needs a new RFC or updating the old one is okay. The old one was kinda silly I thought because it just confirmed that we were gonna do a ruby version bump in Chef 12, which was allowed by the other versioning RFC

lamont-granquist:
Yeah, the old one is kind of so out of date now that keeping it around seems like keeping around cruft.  I'd rather replace it.  I don't think anyone cares right now about why we dropped 1.9.x support back then.

danielsdeleo:
I'm 👍 on having this text. Though even if the old RFC is obsolete and not relevant to anything, it still feels weird to rewrite it with a very different meaning and context. I guess I'm -0 on that issue.

ranjib:
:+1:  im fine with updating existing one. But I'm not sure if thats a legit approach

rfc034-ruby-193-eol.md
```diff
@@ -5,9 +5,10 @@ Status: Accepted
 Type: Standards Track
 ---
 
-# Ruby 1.9.3 EOL
+# Ruby Version Deprecation Policy
 
-Chef 12.0 will require Ruby 2.0 or greater.
+Establishes a rolling version policy for Chef and ChefDK and the
+assocated ecosystem (berkhelf, test-kitchen, etC).
```
tas50: Extra capital C here

lamont-granquist: "berkhelf" and "assocated" as well

i think i typed this out when i was high on too much coffee and gem pinning rage...

lamont-granquist:
I also very deliberately didn't bother mentioning ruby 3.x because we don't have time travel yet, and i don't want to constrain our future hands.

thommay:
👍

thommay:
(although i am disappointed that you're not inventing time travel as a pre-req of this rfc, lamont)

coderanger:
👍

btm:
:+1:

thommay:
i deciderate, by the power of greyskull!

thommay:
approved, @chef/rfc-editors

nathenharvey:

PR #212, opened 2016-08-02
Restrict Solaris 10 minor version support to latest

jkeiser:
Solaris 10 support is presently set at 10u7 and above. This change clarifies that 10u11 is the minimum minor version of Solaris 10 supported.

This is important because IPv6 and networking support was greatly improved post-10u7, and we presently have to apply patches (and disable ipv6) to compile for older Solaris versions. It will allow Solaris Ruby to compile without patches and mapfiles as well, and enable compiler optimizations which are currently disabled, leading to a faster Ruby, and more frequent releases (due to not messing with patches, older Solaris and mapfiles).

### DIFF
--- a/rfc021-platform-support-policy.md
+++ b/rfc021-platform-support-policy.md
@@ -33,16 +33,16 @@ platform that the manufacturer supports.
 
 Platform | Versions | Architectures | Package Format
  ---- | --- | --- | --- | ---		
-AIX | 6.1, 7.1, 7.2 | ppc64 | bff 
-CentOS | 5, 6, 7 | i386, x86_64 | rpm 
-Cisco IOS XR | 6 | x86_64 | rpm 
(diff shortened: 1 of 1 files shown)


lamont-granquist:
👍  it is indeed a massive PITA in our build process and is a leading reason as to why we can't have nice things (e.g. ruby 2.2, ruby 2.3)

thommay:
👍

johnbellone:
:+1:

scotthain:
👍

thommay:
decided, approved, take it away @chef/rfc-editors

MarkGibbons:
It would be nice if ohai reported the update level to make it simple to identify the servers that need a solaris update.  I'll probably submit something.

nathenharvey:

PR #214, opened 2016-08-05
Update RFC template in RFC 000

nathenharvey:
It now matches the [rfc-template.md](new/rfc-template.md) in the `new` directory


coderanger:
Merging under my authority as an editor :)

PR #220, opened 2016-09-12
Add Title metadata

thommay:
This makes it easier to automatically generate statuses and such

cc @chef/rfc-editors


adamleff:
💯  and thanks for taking the time to make the title display consistent in each of the PRs.

lamont-granquist:
👍

PR #221, opened 2016-09-12
Generate GitHub pages for RFC

thommay:


jjasghar:
:metal: great idea.

coderanger:
For anyone that wants to see the output https://chef.github.io/chef-rfc/ 👍

coderanger:
@thommay Would it be possible to parse out the first `<h1>` from the markdown as a title if one isn't given explicitly?

ranjib:
:+1: 
Its hard to tell which of the RFCs are done. If state is "final" does that mean the RFC is done? If so, I'll go ahead and update couple of them at least

thommay:
@ranjib yeah, not sure what the right answer is there.

@coderanger it's annoying to do, and I'd rather make someone do an extra line of work than rely on something fragile :)

nathenharvey:
👍  I will add this to future agendas

PR #222, opened 2016-09-12
Adding Google to allowed prefixes

nelsonjr:


coderanger:
You might want to use a different prefix, as there are a large number of cookbooks already using `google[-_]`. Check out https://supermarket.chef.io/cookbooks?utf8=%E2%9C%93&q=google&platforms%5B%5D=.

nelsonjr:
Thanks for catching that. Updated to google-g<product> so we can publish with the same look-n-feel.

coderanger:
Hmm, so this would be the first prefix no being split on word boundaries. The current prefixes are all of the form `poise-foo` or `fb_bar`. I think you're asking for the prefix `google-g`? Not sure I love that but I defer to @robbkidd and @nellshamrell.

nelsonjr:
Well, the namespace is squatted already (not by the squatters fault).

Now what we are trying to do is workaround the obvious "google-<product>" expected name into something else. If you look at Microsoft you see ms<product>, microsoft-<product>, msazure_... Looking at these I have no idea which ones belong to Microsoft and which ones does not. 

I talked with JJ about this a little bit. Other than google it seems there is not much we can use for that. I'm not sure doing something like goog- or gg- would make it any better.

Note that google-cloud is also taken, so calling it googlecloud- might be as well confusing IMHO.

lamont-granquist:
what about just `gc_` similar to `fb_`?

nelsonjr:
"FB" is how people usually abbreviate Facebook, so it seems a good compromise. I'm not aware of a similar to Google. "gc" would mean more "garbage collection" than "Google Cloud" IMHO.

coderanger:
I would recommend you start with `gcloud` and register prefixes for other products as you want them.

nelsonjr:
`gcloud` seems to be a good choice. I'll run this internally and update the PR accordingly.

lamont-granquist:
`gcloud` sounds fine to me

nelsonjr:
We'll prefer gcloud for our work. Would it be okay if we keep google so it discourages new modules that would confuse with Google made ones, encouraging prefixes?

coderanger:
I think it would be better for you to pick one. This is not intended to allow groups to do a land-grab. Brand continuity and recognition implies you would only use one anyway. I don't think it's reasonable to squat all of `google` given how many of the existing users are reasonable. That said, @robbkidd and @nellshamrell get the final say.

nelsonjr:
SGTM, just asking :-). I'll keep gcloud only then.

nelsonjr:
Kept only gcp for now.

coderanger:
:+1: Confirmed that no one is using that currently. @robbkidd @nellshamrell what say ye?

robbkidd:
👍  Confirmed the confirmation that the prefix is not currently in use on public Supermarket.

PR #225, opened 2016-09-20
Update rfc062-exit-status.md

tas50:
Add an exit code for chef upgrades. This would initially be used by the omnibus_updater cookbook when the client exits so that users an write kitchen configs which expect that code and run again. This allows users to test their upgrade process locally before pushing it out to even a dev environment, where a failure could impact the business.

### DIFF
--- a/rfc070-chef-upgrade.md
+++ b/rfc070-chef-upgrade.md
@@ -29,7 +29,7 @@ Users should be able to automatically upgrade to the desired version of chef wit
 
 At the beginning of each chef run, the client should decide whether or not to upgrade, based on configuration supplied by the administrator.
 
-If the administrator has specified a version of chef different to the one currently running, the chef client should check with an update service to ensure the specified version is available. If so, the client will download the specified version, and install it appropriately. The client run should then exit, using an appropriate exit code to signal an upgrade.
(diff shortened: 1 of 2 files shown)


jeremymv2:
I like this idea.  It shores up the strategy around chef-client upgrades using omnibus_updater cookbook here: https://github.com/chef-cookbooks/omnibus_updater/pull/130

thommay:
I think the Upgrade RFC also needs updating to include this. But 👍

johnbellone:
+1

rfc070-chef-upgrade.md
```diff
@@ -29,7 +29,7 @@ Users should be able to automatically upgrade to the desired version of chef wit
 
 At the beginning of each chef run, the client should decide whether or not to upgrade, based on configuration supplied by the administrator.
 
-If the administrator has specified a version of chef different to the one currently running, the chef client should check with an update service to ensure the specified version is available. If so, the client will download the specified version, and install it appropriately. The client run should then exit, using an appropriate exit code to signal an upgrade.
+If the administrator has specified a version of chef different to the one currently running, the chef client should check with an update service to ensure the specified version is available. If so, the client will download the specified version, and install it appropriately. The client run will then exit with exit code 213 to signal an upgrade has occurred as defined in RFC 62.
```
thommay: 👍

ranjib:
:+1:

thommay:
Approved!

PR #226, opened 2016-09-21
Calendar Versioning for Chef

coderanger:
Okay, let's lead off with some disclaimers:

1. I'm way over quota on pending RFCs, sorry everyone. But this is less about work I want to do and more that we need to at least discuss this so should be okay.
2. I like SemVer overall and will keep using it in (much smaller) projects.
3. Even though I think this is a good move overall, it makes me deeply uncomfortable.
4. There are some numbers in the Specification section that need tuning: when we would start using CalVer (i.e. would there be a Chef 13 and then CalVer, switch to CalVer "before" Chef 13, etc), what deprecation timelines should look like.

With that, have at it. This would be a really big change, but we've survived version scheme changes before (`0.10.x` -> `10.x.y`).

### DIFF
--- a/new/calver.md
+++ b/new/calver.md
@@ -100,7 +100,7 @@ minor release. This either leads to sad users in the former case, or stressed
 maintainers in the latter. Having a support policy like this wouldn't preclude
 cookbooks from defining their own policy, but it would be a good starting point
 for most community authors. If combined with date-based version numbers above,
-it can make it very easy to tell if a release is LTS (eg. `x.0` or `x.6`).
+it can make it very easy to tell if a release is LTS (eg. `x.1` or `x.7`).
 
 The specifics of what constitutes "long term support" is left to another RFC.
 
--- a/new/calver.md
+++ b/new/calver.md
@@ -1,19 +1,19 @@
 ---
 RFC: unassigned
-Title: Calender Versioning for Chef
+Title: Calendar Versioning for Chef
 Author: Noah Kantrowitz <noah@coderanger.net>
 Status: Draft
 Type: Standards Track
 ---
 
-# Calender Versioning for Chef
+# Calendar Versioning for Chef
 
 Since its inception, Chef has followed the structure of [Semantic Versioning](http://semver.org/)
 (SemVer). This way of expressing API compatibility has been
 popular in the software world for many years, and is a widely understood and
 well-liked standard. Unfortunately as Chef has grown both in scope and user
 base, this is increasingly difficult to manage. This RFC proposes migrating to
-a [Calender Versioning](http://calver.org/) (CalVer) standard instead.
+a [Calendar Versioning](http://calver.org/) (CalVer) standard instead.
 
 ## Motivation
 
--- a/new/calver.md
+++ b/new/calver.md
@@ -55,8 +55,8 @@ right move for the time, we now have a looming issue: Chef 13 _(cue spooky music
 Chef 12.0.0 was released in December 2014, meaning at the time of writing it is
 rapidly approaching two years old. In those two years we have made many improvements
 to the codebase and moved to deprecate numerous old mis-features and bad APIs.
-We are not faced with the prospect of Chef 13 representing an even bigger
-compatibility break than Chef 12 was, and even that was a long slog to get the
+We are now faced with the prospect of Chef 13 representing an even bigger
+compatibility break than Chef 12 was, and that was a long slog to get the
 community to upgrade. While it may be too late to avert this Chef 13 calamity,
 if something isn't done it is likely the cycle will repeat and Chef 14 will have
 all the same problems.
@@ -117,10 +117,11 @@ option to allow testing for future compatibility with a specific release.
 
 ## Specification
 
-Starting with either the next major release (i.e. Chef 13) we will
+Starting with the next major release (i.e. Chef 13) we will
 change Chef version numbers to a `year.month.build` format. The `year.month` for
-a release is the date it is released, not when work s begun. If a monthly release
-is missed, we will move on to the next version number.
+a release is the date it is released, not when work is begun. If a monthly release
+is missed, we will move on to the next version number. The `build` number retains
+the same auto-bump semantics as it currently has.
 
 Feature deprecations will be classified as either high-impact or low-impact.
 High-impact deprecations will take place over 6 months, low-impact over 2 months.
--- a/new/calver.md
+++ b/new/calver.md
@@ -1,0 +1,161 @@
+---
+RFC: unassigned
+Title: Calender Versioning for Chef
+Author: Noah Kantrowitz <noah@coderanger.net>
+Status: Draft
+Type: Standards Track
+---
+
+# Calender Versioning for Chef
+
+Since its inception, Chef has followed the structure of [Semantic Versioning](http://semver.org/)
+(SemVer). This way of expressing API compatibility has been
+popular in the software world for many years, and is a widely understood and
+well-liked standard. Unfortunately as Chef has grown both in scope and user
+base, this is increasingly difficult to manage. This RFC proposes migrating to
+a [Calender Versioning](http://calver.org/) (CalVer) standard instead.
+
+## Motivation
+
+    As a Chef maintainer,
+    I want to improve Chef over time,
+    so that previous design decisions do not cause an undue burden on maintainers.
+
+    As a Chef user,
+    I want to upgrade Chef,
+    so that new features and bug fixes can be used.
+
+## What Is CalVer?
+
+The likely-biggest user of CalVer is the Ubuntu project. Here we can see some
+of the defining characteristics of CalVer. They release on a relatively fixed
+schedule (twice a year) and feature deprecation plans are expressed in terms of
+timelines instead of major version bumps. Ubuntu also uses the timestamp of the
+release to form the version number, though other CalVer projects forgo this in
+favor of a more traditional-looking version number.
+
+## Why CalVer?
+
+SemVer is based on the idea that a release is either "compatible" or
+"incompatible" with the prior release. If it's compatible then it's a minor or
+patch bump, otherwise it's a major bump. Major bumps are generally scary because
+they mean a user may need to do some kind of porting work before upgrading. The
+root problem is that as a project grows, it is increasingly hard to say what is
+an API change and what isn't. The surface area of a project grows rapidly,
+especially in something like Chef with a mix of DSLs, internal APIs, and command
+line tools (without even starting in on the Chef Server APIs). Major version
+bumps and the ensuing porting work make users unhappy, so as a project we try
+to minimize compatibility breaks.
+
+## The Chef 13 Problem
+
+While all the decisions Chef has made with respect to versioning and compatibility
+have come from the best of intentions and I would stand behind all of it as the
+right move for the time, we now have a looming issue: Chef 13 _(cue spooky music)_.
+Chef 12.0.0 was released in December 2014, meaning at the time of writing it is
+rapidly approaching two years old. In those two years we have made many improvements
+to the codebase and moved to deprecate numerous old mis-features and bad APIs.
+We are not faced with the prospect of Chef 13 representing an even bigger
+compatibility break than Chef 12 was, and even that was a long slog to get the
+community to upgrade. While it may be too late to avert this Chef 13 calamity,
+if something isn't done it is likely the cycle will repeat and Chef 14 will have
+all the same problems.
+
+## How Does CalVer Help?
+
+Chef has already largely moved to a calendar-based release cycle, specifically
+monthly releases. Currently all deprecation warnings are phrased that we will
+drop the hammer in the next major version in accordance with SemVer. With a
+CalVer standard we could instead create a timeline for deprecations, where a
+given feature will be maintained for X releases (months) and then removed along
+with a slowly ratcheting warning scheme (passive warning, active warning, error).
+This would help spread the workload of upgrades out over a longer period of time,
+avoiding the "major version cliff" we are currently facing. The downside is,
+unfortunately, the same as the upside; this would mean that moving from one
+monthly release to another would involve more risk than it currently does. I
+think this can be reduced with good testing tools which I'll cover below, but
+not every Chef user writes tests and this will bite a non-zero number of them
+no matter how much we advertise deprecations.
+
+### Date-Based Versions
+
+While not required to follow a CalVer-style deprecation workflow, date-based
+version numbers provide a lot of useful information in an otherwise-opaque
+integer field. The most common form for date-based version numbers is to use
+`year.month` as the prefix, often expanded to `year.month.patch` to allow for
+more than one release in a month if a bug fix is needed. This allows clear
+understanding of how old any given release is, as well as knowing when a
+specific future release will happen. This also aids in planning deprecation
+cycles.
+
+### LTS Releases
+
+A possible side benefit of CalVer would be the ability to designate certain
+releases as "long term support" in a similar way to the Ubuntu and RedHat projects.
+This would help cookbook authors by providing a much more limited set of
+releases that we expect community cookbooks to support while not leaving users
+who need to be able to go longer between upgrades in the dark. Currently most
+cookbooks either only test with the latest version of Chef or test with every
+minor release. This either leads to sad users in the former case, or stressed
+maintainers in the latter. Having a support policy like this wouldn't preclude
+cookbooks from defining their own policy, but it would be a good starting point
+for most community authors. If combined with date-based version numbers above,
+it can make it very easy to tell if a release is LTS (eg. `x.0` or `x.6`).
+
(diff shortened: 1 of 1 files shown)


miketheman:
Verrrrry interesting!

This appears to not only impact the versioning scheme, but also the deprecation policy cycle.

For clarity, Django has not adopted CalVer - it appears the intent from upcoming version 2.0 to have LTS releases going forward. The _do_ have an desired 8 month release cycle cadence - and that appears to have little to do with version numbering.
https://docs.djangoproject.com/en/dev/internals/release-process/

I guess a lot boils down to: How often does a user want things to change, in a risk-vs-reward scenario?

To use the Ubuntu examples provided, I can say with some certainty that most places I've worked with production versions, we had tested and pinned to an LTS - sometimes much longer than the 3 year cycle - because the base-level distro packaging provided all the base-level dependencies we needed.
At no point had we ever considered that "there's a new LTS out, I'm sure `apt-get dist-upgrade` will work from our current servers!" - fully expecting breakage.

And those are 2 years apart.

To say nothing of the packages (cookbooks) that are not part of the distributed component (Chef) - if Chef packaged all cookbooks together with the server, client, etc - then this would make more sense as a versioning scheme to me.
But we don't do that, because we've all had that moment where we've been on an LTS distro, and wanted a package that exists in a distro release after our current one - so decoupling cookbooks from the execution environment was a good choice.

One might ask - is the monthly release cadence of Chef too "fast" for the users of Chef? And my answer is: Usually not, when the expectation of "I have a version of 12.x - anything added to 12.x that I don't use won't matter, as long as everything I use in 12.x remains working" - aka SemVer.

But if a release has a deprecation in Month 1, and a release in Month 7 "breaks" it by removing the deprecated code, I completely see a world where a user doesn't update for Months 1-7, never sees the deprecation notice, installs the new date-versioned Chef, and has their world broken.

Maybe the deprecation cycle cadence is the part I'm most resistant to - using Django as a pattern, I've taken their deprecation cycle example and annotated it with imaginary dates on an 8-month release cycle:

```
2001-02 X.0
2001-10 X.1
2002-06 X.2 LTS
2003-02 Y.0: Drop deprecation shims added in X.0 and X.1.
2003-10 Y.1: Drop deprecation shims added in X.2. 
2004-06 Y.2 LTS: No deprecation shims dropped (while Y.0 is no longer supported, third-party apps need to maintain compatibility back to X.2 LTS to ease LTS to LTS upgrades).
2005-02 Z.0: Drop deprecation shims added in Y.0 and Y.1.
```

From here, we can predict a 2-year LTS support cycle (similar to Ubuntu's release cycle), meaning the user really should consider that time frame as a decent amount of time to benefit from the software before sinking in the upgrade testing efforts for the new version.

With Chef monthly release cycles introducing and then removing deprecated features, the timeline in which a user may hit a problem is greatly reduced, as you mention yourself:
> ... this would mean that moving from one monthly release to another would involve more risk than it currently does ...

Does the risk outweigh the reward for the developer? Certainly no developer likes to carry around deprecated code for longer than absolutely necessary. Minimizing the window of time that a deprecation lives makes authoring much nicer, as you get to clean up things in under 2 years.

Does the risk outweigh the reward for the end-user? Probably not - as an end-user of Chef, the _last_ thing that I wanted to do on my never-ending list of maintenance tasks was to routinely upgrade components that are in a working, debugged, profiled state.
I'd posit that most operators want to focus on building, maintaining, debugging the parts that make their business unique, not worry as much about the gas that powers it, that should remain a known quantity for as long as possible.

Do we know how often people upgrade their Chef versions today?

johnbellone:
We tend to upgrade our version quarterly. There is no real reason we can't do it quicker, but it's typically only been for fixes in the client and Ohai.

jonlives:
Personally I'd be against making each release more risky vs the compatibility cliff. Speaking personally, a Chef major version upgrade right now is often quite time consuming and painful *but* it is also something we can plan around and allocate a chunk of resource to at a specific time. We're still able to upgrade to minor & patch releases for bug / security fixes.

My personal feeling is that calver would make each individual upgrade of Chef client to *any* version potentially painful and result in for example having to deal with a breaking upgrade just to get a security patch or new feature.

The general case I guess of this would be that installing Chef upgrades is typically just one routine task in a broad spectrum of planned work for ops teams (or whoever does it). Making that routine task more frequently risky instead of "generally safe but with a big chunk of work at major upgrade time" has the potential to make the general baseline of dealing Chef more painful than it is now, which feels bad to me.

lamont-granquist:
When Chef releases a minor version bump that has a backwards incompatibility, why does this "break" anyone that consumes Chef?  You were running your business fine on the prior minor release, presumably.  You shouldn't be "broken" by it.  Its true that if there's a backcompat break you'll be blocked from upgrading and getting bugfixes and patches until you do some work on your end -- but nobody should be "broken" by a chef-client upgrade.  If they are, then their business is exposed to all the mistakes that we make, along with the possibility of disgruntled employees or something like state-sponsored hackers uploading malicious code.

To me this really comes down to the contract that we have with existing users as to how much work they're required to do in order to obtain the work that we're doing.   Right now there's an assumption that existing users are allowed to do zero work in order to get all the bugfixes for up to 2 years.  That is pretty horrible for software development, and for anyone who gets stuck on a major-version breaking change bug that we can't fix outside of bumping the major version.  Its also bad for new users, and it prioritizes existing customers and the fact that they've invested time into figuring out baroque aspects of the Chef API over new users who will encounter that confusion for the first time.

We are also getting more and more users show up in issues being annoyed at open bugs not being fixed for years, e.g.:

https://github.com/chef/chef/issues/2514

So when you're considering this issue you need to put yourself in the frame of mind of someone who is blocked on a bug where people have raised backcompat fears over fixing it (and we probably now have quite a few backcompat bugs which are blocked on fear, but we could probably fix and either nobody would notice or maybe we'd have to tell that one person that they're doing something horrible and they'd have to change their code).   The result is that because we're all clever enough to consider all kinds of _possible_ ways that people _might_ be affected by changes that changes become extremely hard to do.  We also wind up considering all kinds of defensive ways to program around issues -- and often the solution becomes overly complicated and either we simply move on and don't do it because the solution has become too hard -- or we plow forwards and cause disruption because the overly-complicated, but (on paper) backwards-compatible solution has hard to reason about edge cases and unintended consequences.

So I'm going to take more of a position that I don't care as much about your ops burden.  Right now I think we're trying to be too stable and it's hurting the software project and its allowing old bugs to fester and creating more hard to maintain code that we can't ever delete.  What we need is a framework that is not constructed around the agenda of minimizing the disruption around upgrades, but one that is constructed **fundamentally** around answering the question about how code changes and how code gets removed.

If we want to go with a cliff, then I think that's fine, but we need to do it more often and have a yearly major release cadence -- this 2 year cycle between 12 and 13 I think has very clearly been too long.  It also should be made clear that we don't have the resources to support the N-1 version of chef-client outside of security upgrades -- so once you hit that major release cliff all bugfixes pretty much end (there were two notable exceptions to that for Chef 11 support after Chef 12 -- one release involved a dumptruck load of money and the other one was to patch ca-certs so that chef-11 could talk to AWS again, the rest of the upgrades were all security fixes for openssl).  For Chef 10 there was a corporate sponsor (facebook) that was responsible for backporting fixes from 11 to 10.  Chef 11 has not had that kind of sponsorship (facebook got straight onto 12 and has vowed to never let that happen again) and the cliff for future major version upgrades likely will be a similar concrete wall.

poliva83:
I had some questions that came to mind while reading this RFC.

- Would using  CalVer over SemVer allow me ability to get me a pre-release chef-client 13 package months in advance of official "LTS" release? 
- If chef-client 13 is expected to come out Jan 1st, 2017 should I expect a pre-release package be available on `https://bintray.com/chef/current` Sept 1st, 2017? 
- Would moving to CalVer mean the code for major version breaking changes are available to community sooner? (Right now when I look over `https://github.com/chef/chef` branches I don't see a feature branch with version 13.x.x)

coderanger:
@poliva83 Pre-releases would work as they do now. After each monthly release we would bump the "minor" component and start work on the next month's release. All automatic builds would be pushed to the `current` channel as they are today, and when we reach a build for the month we are happy with, we promote it to `stable` and start over. I think the answer to all three of your questions is "No" as those processes wouldn't change compared to what they are now.

coderanger:
Basically the big change here is proposing that deprecated code be removed on a fixed schedule relative to when it was deprecated rather than when the next major release happens. Adding more semantic info to the version number (in the form of a date) is a side benefit that dovetails nicely, but isn't the most important part. I'm not proposing any changes to when deprecations happen, just how they are handled after that point.

danielsdeleo:
Personally, I would like the "rolling breakage" cycle. I think part of the problem with major version upgrades is that all the breaking changes get packed into one release, so everyone experiences at least one problem. I also dislike that very small and sensible, yet technically breaking, changes get blocked for a very long time. For example, checking GPG signatures of SUSE packages.

That said, I think we need to define what the timelines look like in order to have an honest debate about this. I think we might need different categories of timeline based on the estimated impact of the change, but maybe this could be the standard timeline:

1. months 0-5: opt-in to new behavior. No messaging in the software (but we can advertise it on discourse, blog, etc.).
2. months 6-11: opt-in to new behavior, user gets a warning if detected using old behavior (detection should be as accurate as possible, we accept that 100% accuracy isn't always possible).
3. months 12-23: opt-out of new behavior. If user selects the opt-out, they get a warning about when the opt-out will go away.
4. month 24: old behavior goes away.

This assumes a change is feature-flaggable, which I think is okay as a standard requirement for a breaking change. In some cases (upgrading ruby version in the packages, etc.) this won't be possible (or if it's possible, it's not worthwhile), but most cases should be flaggable.

For high impact changes like attributes or something, maybe the timeline is roughly doubled?

coderanger:
@danielsdeleo I don't think feature flags really make sense given how we deprecate things is currently always a new API or parameter? I guess that's a kind of feature flag, but I don't think we need to leave time for that in the schedule as its own category. 24 months also seems _way_ too long, as Lamont mentioned before, I think we should set an absolute maximum for the breaking-est of changes as 12 months because past that the maintenance burden is too high (as it is now). Most changes should be less than that though, I put 2 months for minor breaks (i.e. something that's very simple to fix) or 6 months for more complex changes. That kind of deprecation workflow seems like something that makes more sense for a hosted application, rather than an API.

danielsdeleo:
So as one example, we put the deadline for `node.set` at Chef 14 which would be probably 2 years from Chef 13 at the given release cadence.

coderanger:
Okay, fair, "special terms and conditions may apply at the discretion of the maintainers" always applies, but that's a bigger break than almost anything we're done in the past 3 or 4 years :)

ryancragun:
Long deprecation cycles increase stability, increase upgrade cliff, increase developer maintenance burden.
Short deprecation cycles decrease stability, decrease upgrade cliff, decrease developer maintenance burden.

To reduce the negative impact of a shorter deprecation cycle I propose adopting a CalVer `year.month.patch` version scheme where breaking changes are allowed on a yearly basis.
This makes it clear that every year there could be breaking changes that I'll have to work through,  makes it easy to write cookbooks that test against current year stable and n-1 stable, reduces the impact of upgrade cliffs because fewer breaking changes will be lumped into a single upgrade, and reduces the super long deprecation maintenance burden on developers.

coderanger:
@ryancragun The whole point of CalVer is to not have a "yearly basis", that would be SemVer and is what we do now. It would be a rotating year, so if we say the biggest deprecations take 12 months that means its 12 months from the first release with the warning until we remove the old code. That helps avoid having big "breaking change" releases with the pent-up workload of 14 (or more by the time we get there) monthly revs.

coderanger:
re: shorter cycles, I think that's pretty universal now. Even if we don't go with this RFC and stick on SemVer I don't think any maintainers want to go more than 12 months between major revs again.

ryancragun:
@coderanger We have SemVer but we lack an actual major bump cadence. My proposal is exactly what you describe: SemVer with a fixed major bump cadence and version scheme that makes it sensible.

ryancragun:
@coderanger As a user of Chef I would much prefer a single smaller "breaking release" event once a year (or maybe every 6 months) then potentially many smaller "breaking releases" throughout the year.

coderanger:
Sure, that's a competing option. How would you solve the upgrade cliff problem though? 11->12 was about a year and we're only just now getting community cookbooks to drop 11.x support because people are still on it.

ryancragun:
We should have gone to Chef 13 a year ago. Then the only supposed versions would be 12 and 13, with 14 and deprecating 12 on the way.

coderanger:
To be clear, I don't think SemVer with a more rapid (and maybe more rigid so that we stick to it) major rev cycle of every 6 or 12 months is not a ludicrous suggestion. If that's the way that we end up going I'll be happy. I just think that the rolling deprecation windows give us a lot more flexibility for less user pain in the aggregate.

coderanger:
As an example, if we do yearly major revs (say in January because easy math but srsly lets not do it in January) and I put in a "big-ish" deprecation in the November release, that means you get only 3 months to deal with it instead of 6. The SemVer-y releases make the dates of breakage more predictable but overall you have a less consistent deprecation experience.

ryancragun:
I wish there was actual data on how often people upgrade their chef-clients. Rolling deprecations help those that regularly upgrade the client by giving them warnings ahead of time but could hurt people that only upgrade periodically because the deprecations could be completely removed by the time they upgrade their clients. If you're an infrequent upgrader you no longer have the benefit of SemVer to know that breaking changes have even occurred.

You're absolutely right that if we went with a yearly cadence that come major release time some major-ish thing(s) could have been deprecated towards the end of cycle and now users have less time to fix them. Yup, either they don't get to upgrade to the new release until they fix their cookbooks or the developer that's doing the deprecating can choose to maintain backwards compat for another year. If the change was major enough I would assume that we'd maintain 15 months rather than hose people with 3.

Perhaps a 6 month cadence of major releases make sense. All deprecations would have to exist for a minimum of one full major release before they're removed. Now users get at least 6 months to fix things between every major release, breaking changes are only introduced twice a year instead of potentially every month, and developers don't have to maintain a deprecation for longer than a year.

danielsdeleo:
@coderanger I'm mostly concerned with getting agreement on what the "standard" deprecation cycles would be, the numbers I put for each stage are changeable. As for the feature flag issue, it depends on the specific circumstance, but I think it can be figured out in terms of:

1. new behavior available for use
2. old behavior triggers a warning
3. old behavior must be turned back on explicitly
4. old behavior gone

Sure, there are some cases where it's impossible or just not worth the effort to have both behaviors available simultaneously, but I don't think those are the majority.

coderanger:
Fair, to put that in more concrete terms if we imagine that the new deprecation API looks like `Chef.log_deprecation('nodeset', '2017.3', 'node.set is deprecated etc etc more message here')` so we have both some kind of name for each deprecation (also helps for docs and displaying things) and the target version I could imagine some kind of config options like the existing treat_deprecations_as_errors, but instead of a boolean it takes an array of names, and have a similar one for treat_deprecation_errors_as_warnings with a similar array. Set up some rules for when a thing transitions from "shows warning in dev" to "shows warning in prod" to "throws error" to "goes away".

thommay:
@ryancragun the last time I ran the numbers on Hosted Chef (just before the AWS migration, so ~6(?) months ago), 4 out of the top 5 versions were Chef 11, and 7 or 8 of the top 10. (There was also a non-trivial amount of Chef 10 showing up.

btm:
@thommay did you save that data? Could you grab it again from HC so we could compare?

thommay:
I got @paulmooring to run the numbers for Hosted Chef again:

Version | Percentage
-------- | ------------
10.16.4 | 22.96
11.12.4 | 20.72
11.18.6 | 13.25
12.11.18 | 7.52
12.13.37 | 7.34
12.6.0 | 6.54
12.10.24 | 6.28

and so on. On aggregate, it looks like about 25% of the total is chef 10, 35% is 11, and 40 is chef 12. So the numbers are getting better, but we still have a massive problem getting people to upgrade - and I really think we need to get that fixed before we make the gaps even larger.

lamont-granquist:
I don't think those numbers are fixable via any policy short of setting a hard date when support will be shut off on hosted.

btm:
We should reach out to some people and ask why they haven't upgraded before we assume that.

nathenharvey:
Adding the `on hold` label to indicate that we plan on deciding on this **after** the [Chef Community Summit in Seattle](https://summit.chef.io/seattle/).

thommay:
I think the summary of our sessions at the London and Seattle summits is that most of our users are much more comfortable with a single, "big bang" hit of deprecations on a well known schedule, rather than a rolling deprecation set. The big bang approach is also much less scary from a security/bug fix perspective since you do not have to be concerned that the critical security fix you need also contains a bunch of deprecations.

coderanger:
Closing this out for the reasons @thommay mentioned. There might be some value is discussing CalVer (without rolling deprecations) for ChefDK but I don't feel strongly enough about that to champion it.

PR #227, opened 2016-09-21
Add Chef Deprecation Policy

tas50:
This gets us on a predictable schedule for shipping major releases and gives our customers large amounts of warning before we introduce major breaking changes.

Signed-off-by: Tim Smith tsmith@chef.io

### DIFF
--- a/new/release_and_deprecation_policy.md
+++ b/new/release_and_deprecation_policy.md
@@ -1,20 +1,20 @@
 ---
 RFC: unassigned
-Title: Chef Release and Deprecation Policy
+Title: Chef Major Version Release and Deprecation Policy
 Author: Tim Smith <tsmith@chef.io>
 Status: Draft
 Type: <Process>
 ---
 
-# Chef Release and Deprecation Policy
+# Chef Major Version Release and Deprecation Policy
 
-A policy for shipping major Chef releases on a predictable schedule along with the communication of breaking changes to users.
+A policy for shipping major Chef releases on a predictable schedule along with the communication of breaking changes in those releases to users.
 
 This RFC is two fold:
 
-It states that Chef will ship a new major release of the chef-client on a yearly schedule.
+It states that Chef will ship a new major release of the chef-client on a yearly schedule, while continuing to ship minor releases on the previous monthly schedule.
 
-It also sets policy for the for when breaking changes can occur, and how Chef will communicate those changes to users. Chef will provide a clear road map of upcoming compatibility breaks, including remediation steps users can take to update their codebase. This allows for the project to progress while also providing users with the information necessary to scope future upgrades. It avoids what has been called the Chef 13 problem, in which the large time between major releases creates a mass of breaking changes that hinders adoption of new major releases.
+It also sets policy for the for how and when breaking changes will be included and communicated to users. Chef will provide a clear road map of upcoming compatibility breaks, including remediation steps users can take to update their codebase. This will allow the project to progress while also providing users with the information necessary to scope future upgrades. It avoids what has been called the Chef 13 problem, in which the large time between major releases creates a mass of breaking changes that hinders adoption of new major releases.
 
 ## Motivation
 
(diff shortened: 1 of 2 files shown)


shortdudey123:
Thanks for writing up a formal deprecation process!

One thought though, if versioning will stay semver then maybe it would be better to increase the frequency of major releases as opposed to keep adding up the minor releases with breaking changes since this is against the semver thought process.

tas50:
@shortdudey123 I purposfully wanted to stay away from SemVer vs. CalVer here, but I think this sort of process would lend itself nicely to SemVer with much more frequent releases. @ryancragun proposed an idea of just releasing on a 6 month or yearly scheduled with major version bumps then. I think that sort of idea fits nicely with this RFC since the identified future release would just be one of those major releases. I think how exactly we do releases is a conversation we need to have and I hope we can get some more RFCs going for that.

shortdudey123:
> stay away from SemVer vs. CalVer here

fair enough

> releasing on a 6 month or yearly scheduled with major version bumps

I was thinking something along that lines too

lamont-granquist:
for both this process and for #226 how are we supposed to handle things like:

https://github.com/chef/chef/pull/3778

that's considered a breaking change because we can anticipate that the output of that would be a public API and that it may be fed into brittle scripts that would break with any output format change.

we can include a chef config switch to change the behavior, but i really worry that Chef::Config is already growing out of control, and having more knobs to control behavior that are mostly useless for the vast majority of users is not good for the product.

its also not clear that we could introduce any kind of deprecation warning here since the output is by definition the API, so including a deprecation warning might break scripts as well (and also, just "ugh" -- a deprecation warning that a blank space will go away in the future?).

tas50:
We don't necessarily have to warn the users. I specifically said we'd warn when technically possible, because a lot of deprecations don't lend themselves to warnings. In this case we'd still go through the process of adding it to the list, stating the impact, warning people that in X months we would deprecate it, and then finally we'd make the change with a notice in the release notes. This RFC is all about giving as much warning as possible and a paper trail to make the upgrade process more predictable for users. In this case as much warning as possible is simply a notes in release notes and on a docs page.

nathenharvey:
Adding the `on hold` label to indicate that we plan on deciding on this **after** the [Chef Community Summit in Seattle](https://summit.chef.io/seattle/).

lamont-granquist:
I'm basically :-1: on any set 6 month deprecation policy in the absence of any kind of classification system for breaking changes.

Some breaking changes are entirely theoretical, or not even SemVer-violation.  Because of ruby users can do literally anything they like.  So one again here is a breaking change we shipped to a user in a minor release which is due to the User poking around very deep in the internals of the Chef API:

https://github.com/chef/chef/issues/4053

Also the stability of the software for end users needs to be balanced against the bugginess of the software.   We get numerous complaints in old tickets about bugs not being fixed, and I think those customer voices are largely lost when we're considering breaking changes.  The fact is that breaking changes are generally done in an attempt to benefit other end users -- we don't purposefully
set out just to mess with people who have running code.

I can guarantee that if we fix bugs and clean up code in the major release and do some entirely necessary housecleaning work that has been needed for 18-24 months since 12.0-12.6 that we will wind up breaking someone entirely accidentally.  I don't want to be playing "GOTCHA!" and rolling back those changes in 13.1.x because someone did something we didn't anticipate.

And that's the other point, which is that we're human software developers and don't fully understand our own software and simply cannot predict how people are using our own software.  And this cuts both ways.  I do think that occasionally problems can stream in unanticipated and we need to figure out how to unbreak and change that we never anticipated would be breaking.  At the same time I think that needs to be a CONVERSATION and not POLICY.

I'd point to these two issues as examples.  In this case we chose to fix the backcompat break even though the cookbooks were using broken code:

https://github.com/chef/chef/issues/4124

In this sequence of tickets we chose to not fix the backcompat break and instead introduce a different API:

https://tickets.opscode.com/browse/CHEF-3964
https://tickets.opscode.com/browse/CHEF-5016
https://tickets.opscode.com/browse/CHEF-5032

I guess I'd be okay with issuing deprecation warnings in advance by policy as long as customers have fully notified us of every use case of Chef they're using so I can play "GOTCHA!" in return, but hopefully its obvious that I'm being ridiculous...

thommay:
I think the summary of our sessions at the London and Seattle summits is that most of our users are much more comfortable with a single, "big bang" hit of deprecations on a well known schedule, rather than a rolling deprecation set. The big bang approach is also much less scary from a security/bug fix perspective since you do not have to be concerned that the critical security fix you need also contains a bunch of deprecations.

thommay:
My feeling is that we should move forward with this RFC, selecting the <bikeshed> first Wednesday in April </bikeshed> to be our yearly major bump point. We should stick with SemVer rather than CalVer, simply because I'm not sure it matters that much; we should also ensure that ohai changes version to match chef's. We'd continue to support N-1 chef versions - ie, next time we'll support 13.x and 12.last for a year going forward, and then we will drop 12 when 14 releases.

thommay:
To summarise, the bits of this that deal with release cadence should be removed, since they're contained in #232 . Once that's done this RFC is approved.

new/release_and_deprecation_policy.md
```diff
@@ -1,20 +1,16 @@
...
-
-It also sets policy for the for how and when breaking changes will be included and communicated to users. Chef will provide a clear road map of upcoming compatibility breaks, including remediation steps users can take to update their codebase. This will allow the project to progress while also providing users with the information necessary to scope future upgrades. It avoids what has been called the Chef 13 problem, in which the large time between major releases creates a mass of breaking changes that hinders adoption of new major releases.
+This RFC sets policy for how and when breaking changes will be included in Chef and how they will be communicated to users. Chef will provide a clear road map of upcoming compatibility breaks, including remediation steps users can take to update their codebase. This will allow the project to progress while also providing users with the information necessary to scope future upgrades. It avoids what has been called the Chef 13 problem, in which the large time between major releases creates a mass of breaking changes that hinders adoption of new major releases.
```
rhass: Making reference to "the Chef 13 problem" is inside-baseball. Perhaps the following rewording would be more clear:

`This alleviates issues with adoption of new major releases which creates a mass of breaking changes caused by large time delta between upgrading major releases.`

thommay: Hmm. I think the real world example is useful here. People who aren't involved at the time can look at what happened in the chef 12 cycle and understand the problems.

thommay:
Accepted and merged as RFC 82, thanks!

PR #228, opened 2016-10-14
Fix markdown for Cookbook Segment Deprecation RFC

thommay:
@chef/rfc-editors there are no content changes here, just markup.


adamleff:
:+1: Merging as obvious fixes.

PR #229, opened 2016-10-18
We've migrated to slack

nathenharvey:


nathenharvey:
@chef/rfc-editors nothing to review/approve here, just updating to reflect current status so this can probably be safely merged, though I'll leave that to you.

PR #230, opened 2016-10-25
Update RFC018 to update flag from -F to -S

tduffield:
The `-F` flag is currently associated with `--format` (i.e. `-Fj`). Instead, we shall use `-S`, the shortened form of `--field-separator`.


coderanger:
:+1: @thommay or @adamhjk want to deciderate this out of cycle since it's a small (and required) change?

thommay:
Oops. Yeah, this is accepted.

PR #232, opened 2016-10-27
Release cadence RFC

coderanger:
Split out the release cadence bits from #227 and add a bit more specific bits. Short and sweet.

### DIFF
--- a/new/release-cadence.md
+++ b/new/release-cadence.md
@@ -14,15 +14,15 @@ versions (eg 12.1 -> 12.2) will include new features and bug fixes but will be
 backwards-compatible to the best of our ability. Patch versions are governed
 by [RFC 47](rfc047-release-process.md).
 
-Chef feature releases are promoted to the stable channel once per month, in the
-second week of the month unless delayed by the Chef Software Release Engineering
-team. Additional patch releases for a given feature release may be promoted if
-critical issues are found.
+Chef feature releases are promoted to the stable channel once per month. It is
+expected that this occur during the second week of the month unless
+circumstances intervene. Additional patch releases for a given feature release
+may be promoted if critical issues are found.
 
-ChefDK is released once per month, in the fourth week of the month unless delayed
-by the Chef Software Release Engineering team.
+ChefDK is released once per month. It is expected that this occur during the
+fourth week of the month unless circumstances intervene.
 
-Both Chef and ChefDK will prepare the release candidate in the week prior to the
+Both Chef and ChefDK will prepare a release candidate in the week prior to the
 target release date.
 
 The Chef release in April of each year is a major version release, which will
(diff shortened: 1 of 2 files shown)


shortdudey123:
Awesome!  If accepted, the major bump would start w/ Chef 13.0 in April 2017 correct?

thommay:
@charlesjohnson FYI

new/release-cadence.md
```diff
@@ -8,14 +8,34 @@ Type: Informational
...
-Chef is released once per month, in the first week of the month unless delayed
-by the Chef Software Release Engineering team.
+Chef follows [Semantic Versioning](https://semver.org/) for releases. Major
+versions (eg. 11.x -> 12.x) will include backwards-incompatible changes, minor
+versions (eg 12.1 -> 12.2) will include new features and bug fixes but will be
+backwards-compatible to the best of our ability. Patch versions are governed
+by [RFC 47](rfc047-release-process.md).
+
+Chef feature releases are promoted to the stable channel once per month, in the
+second week of the month unless delayed by the Chef Software Release Engineering
```
thommay: not sure we need to specify the chef internal structure; maybe just say that we'll aim to do a release a month unless the universe prevents us?

coderanger: Also happy for a better way to say "Nth week unless shit happens" that sounds reasonable, don't want to throw releng under the bus, something that balances standard expectations with reality happening would be good.

thommay: Ah, I meant that I don't think we need to talk about RelEng. Also I would like to eventually get at least the concept of any maintainer doing releases…

```
Chef feature releases are promoted to the stable channel once per month. It is expected that this occur during the second week of the month unless circumstances intervene.
```
Something like that?

coderanger: Much better, thanky!

charlesjohnson: "Chef targets feature releases for the second week of every month" gets it across, but ^^^ is fine.

charlesjohnson:
@shortdudey123 That's correct.

new/release-cadence.md
```diff
@@ -0,0 +1,49 @@
...
+
+Chef feature releases are promoted to the stable channel once per month. It is
+expected that this occur during the second week of the month unless
+circumstances intervene. Additional patch releases for a given feature release
+may be promoted if critical issues are found.
+
+ChefDK is released once per month. It is expected that this occur during the
+fourth week of the month unless circumstances intervene.
+
+Both Chef and ChefDK will prepare a release candidate in the week prior to the
```
charlesjohnson: This is already covered in RFC-047. Can we delete the language here so it's not covered in two places?

thommay:
The language about when an RFC is released shall be moved here from RFC 47. Once that's done, this RFC is accepted.

PR #233, opened 2016-10-31
chef run

jjasghar:
- This is a rfc that helps design a new feature for our ChefDK.
- The ability to run "one off" recipes and have a tool chest of system
administrative scripts will make chef for admins more delightful.

Signed-off-by: JJ Asghar  <jj@chef.io>

### DIFF
--- a/new/rfc-chef-run.md
+++ b/new/rfc-chef-run.md
@@ -59,8 +59,10 @@ $ chef run <ssh|winrm> <list-of-hostnames.txt|yml> recipe.rb
 
 ## Downstream Impact
 
-- There is a `knife` plugin called `knife-solo` that does this type of work, but
-having it baked into the chefdk would allow for a natural adoption.
+- There is a `knife` plugin called [knife-solo][solo] that does this type of work
+but has not been adopted by the ChefDK.
+- There is a `knife` plugin called [knife-zero][zero] that does this type of work
+but has not been adopted by the ChefDK either.
 
 ## Copyright
 
@@ -68,3 +70,6 @@ This work is in the public domain. In jurisdictions that do not allow for this,
 this work is available under CC0. To the extent possible under law, the person
 who associated CC0 with this work has waived all copyright and related or
 neighboring rights to this work.
+
+[solo]: https://matschaffer.github.io/knife-solo/
+[zero]: https://github.com/higanworks/knife-zero
--- a/new/rfc-chef-run.md
+++ b/new/rfc-chef-run.md
@@ -1,0 +1,70 @@
+---
+RFC: unassigned
+Title: chef run command
+Author: JJ Asghar <jj@chef.io>, Edmund Haselwanter <me@ehaselwanter.com>
+Status: Draft
+Type: Process
+---
+
+# chef run
+
+The ability to run "one off" recipes to remote machines would be able to empower
+a system administrator to leverage Chef and a single Chef recipe to validate a state
+of a machine. Having command built into the ChefDK would allow for quick adoption
+of Chef's main builtin resources in turn helping with education and the "15 minute"
+story of chef.
+
+A good example recipe would be something like `install-and-stop-nginx.rb`:
+
+```ruby
+package "nginx" do
+  action :install
+end
+
+service "nginx" do
+  action [ :stop ]
+end
+```
+
+With this recipe you can ship it off to your clients, see if Chef is installed,
+if not, run the `curl -L https://chef.io/chef/install.sh | sudo bash`
+then would pragmatically run `chef-client -z install-and-stop-nginx.rb` and have it
+us it's idempotent way to verify than nginx is installed but down.
+
+With this tooling the inverse would be possible with something like the following
+with a `start-nginx.rb`:
+
+```ruby
+service "nginx" do
+  action [ :start ]
+end
+```
+
+With the same workflow, allowing you to bring up your nginx web servers.
+
+## Motivation
+
+    As a system administrator,
+    I want to leverage chef resources in a disposable manner,
+    so that I can run disposable scripts on remote machines.
+
+## Specification
+
+The general user experience would be the following:
+
+```
+$ chef run <ssh|winrm> <hostname> recipe.rb
+$ chef run <ssh|winrm> <list-of-hostnames.txt|yml> recipe.rb
+```
+
+## Downstream Impact
+
+- There is a `knife` plugin called `knife-solo` that does this type of work, but
+having it baked into the chefdk would allow for a natural adoption.
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.


coderanger:
Just to make sure I understand, this is kind of a mashup of `knife ssh` and `chef-apply`? I'm kind of on board but the use of Chef for that kind of one-off stuffs has always been kind of wonky because don't have the full metadata structures of a cookbook and much of our flow is built around that (ex. dependencies, template source files). Given the workflow complexity it might be prudent to prototype this outside of ChefDK so it isn't bound to ChefDK's release cycle in the very short term, and as a side benefit it will be easier to see the workflow when we do decide to move it into ChefDK itself.

shortdudey123:
Can you elaborate a bit more on the use case?

I am not sure i follow why you would want to use a recipe outside the context of a cookbook.
I sometimes use `chef-client -o recipe[cookbook1:recipe1]` but thats still in the context of a cookbook

jjasghar:
@shortdudey123 I guess the another use case would be to use the built-in Chef Resources to turn on every web server in your fleet. Imagine you had to bring up `httpd` on every single node in a list, and you could easily do a `chef run ssh <LIST.txt> install-and-stop-httpd.rb` to install `httpd` on every machine.

Where `LIST.txt` would be:
```
web01.example.com
web02.example.com
web03.example.com
...
web24.example.com
```

and your `install-and-stop-httpd.rb` would be:
```ruby

package "httpd"

service "httpd" do
   action [ :stop ]
end
```

After writing this out I realize it's the same example from the RFC, but you should see how you can do "one-off" recipes to do simple tasks on remote machines. This would only require the ChefDK to be installed and a text editor giving you the power of `chef` so you can be empowered to do other things.

It's a way to help introduce someone to chef without having to show the full infrastructure also, this is extremely advantageous, so you can gain a way to have a script like above instead of something like the infamous bash for loop like the following:

```bash
for i in `cat LIST.txt` do ; ssh $i -c "sudo yum install -y httpd && service httpd stop" ; done
```

All system administrators have written something like the above in their career, and I'm trying to make it so they could just leverage `chef` instead. Not to mention leveraging `chef` here if you don't have `yum` based distros, someone threw in a `deb` based, in theory you could write logic in the `.rb` and not have to create such a brittle `for loop`.

Does this answer your question?

ehaselwanter:
The main idea circles around a chef native, declarative command & control. 
And with this RFC and contributions to it we would like to explore this very idea.

@coderanger your are absolutely spot on with `knife ssh && chef-apply`

and my main concerns are definitely with the usefulness outside the actual chef-client run (no access to server node attributes /  cookbook attributes)

on the other hand: the very same issue (not knowing the parametrised values of the original declarative system, be it chef/puppet) is true for all other c&c tooling out there.
rundeck/mcollective come to mind.

 but being able to leverage build in chef resources like

```
service "httpd" do
   action [ :stop ]
end
```

still looks like a win to me. We can build an arsenal of transportable one-off chef-apply style snippets.

so let us explore the idea :-)

jjasghar:
@ehaselwanter and @coderanger I think it should be noted that this RFC would be to use `chef-client -z` not `chef-apply`. I had some internal discussions at Chef about this, and after some discourse, leveraging `chef-client -z` seemed like the most delightful option.

coderanger:
@jjasghar You mean `chef-solo` because `client -z` can now return to the memory pile? ;-)

Either way, my concern is that production Chef code rarely looks like that. I agree it would be lovely, but the `httpd` cookbook currently has 7 different providers in all for `httpd_service` all of which have to be more verbose than that. And because the "single flat recipe" style has no room for dependencies, we can't tell people to just use `httpd_service` itself. Basically I'm not against the overall idea but this proposed implementation feels overly simplistic compared to what we know about production Chef code.

jjasghar:
@coderanger about this:
>  my concern is that production Chef code rarely looks like that.

I think this is where this RFC can come in and shine. Yes, "production chef cookbooks" never would be something like the examples, but this isn't cookbook development.

These are one-off scripts to leverage chef to do service management and the like. I wouldn't think a lot of Custom Resources would even fit here, it'll mainly be the built-in resources that Chef provides.

shortdudey123:
@jjasghar makes sense

Personally, i just use knife-ssh with one offs since i don't see a point to saving them since they will not be used again by definition.

jjasghar:
I'd like to mention we have this blog post:

https://blog.chef.io/2015/01/16/convert-your-bash-scripts-with-chef-apply/

That this would be the natural progression from this blog post too. /cc @randomcamel

coderanger:
So fully laid out, `chef-apply` fits a bunch of niches. One is learning, for people that are just starting out and for that the existing `chef-apply` works fine. Next is true one-offs, `knife ssh 'somequery' 'chef-apply -e "some recipe"'` works, but is arguably pretty ugly. For a one-off, it's probably enough though. Then the currently under-served bit is runbook automation. There is definitely a void for a shim to make Chef look more like a procedural system while still having the convergence power we all know and love. The place I personally see this the most is with application deployment, but it's got a lot of little mini-niches where you want some kind of procedural shim.

My concern is that most of those places where you want runbook automation or a similar thing, I really don't think the plain Chef recipe DSL is going to be enough. Even that kind of service management example, the service resource doesn't cover things like runit or supervisor. Or even further down, on Ubuntu 14.04 you often need to manually select debian vs upstart providers depending on the service.

Our recommended solution to that whole mess has been custom resources and high-quality community cookbooks, so I think we shouldn't just leave those features on the sidelines. I see a few options for this. One would be to make `chef run` consume whole cookbooks or folders of cookbooks. That is probably limiting on the UX side given how much overhead there is to write a cookbook (related: we should fix that, but that's for a different day). Another option would be to allow putting some kind of markup in the recipe code to indicate dependencies and then have the system (_::jazzhands::_) handle that. As a possible implementation, generate a temp policyfile based on parsing the recipe comments and run `export` on that and then feed that in to `chef-solo`.

ehaselwanter:
yep. the headline should be 'runbook automation - chef style' and I like to quote your:

> There is definitely a void for a shim to make Chef look more like a procedural system while still having the convergence power we all know and love.

jjasghar:
> Next is true one-offs, knife ssh 'somequery' 'chef-apply -e "some recipe"' works, but is arguably pretty ugly

Wouldn't this also require the `"some recipe"` on the remote machine too? With what I'm proposing you'd be able to ship it in the command and have it from your local workstation to your remote machine.

I should mention we could in theory leverage machines like switches and networking gear with this type of workflow. Which is something I know a lot of us want ;) /cc @edolnx

thommay:
My concerns are, in no particular order:

 - that this looks like an unfunded mandate. My suspicion is that it'll be a fair chunk of work hooking the transport stuff up and supporting it, not to mention few people really use chef-apply and we definitely have seen some weird bugs there.
 - that it doesn't really address the "chef in 15 minutes" story. I'm actually worried it does the opposite, and introduces a "production cliff" into chef that we'd be hard pressed to get round. By ignoring cookbooks (and attributes, templates, etc) we give people a very half hearted impression of chef.

My preference would be, as Noah alludes to, a system giving one a lightweight language (yaml?) that would be enough to get started, that under the hood would generate a cookbook and use it. You could then policy -> dependencies -> export, as suggested. (I actually described this in the pub last week to @jaymzh, who looked suitably horrified).

coderanger:
Also while I recoil a bit from doing so, we should mention OpsWorks as prior art here. Their multiple callbacks system seems like it was intended to be vaguely in this direction (albeit in a limited fashion).

jaymzh:
The example of "I want to start a webserver on all my webservers" makes my recoil quite a bit... because, why isn't that in "normal" chef? It feels like this is promoting doing things not checked into code repo, and not actually making reproducible systems (which is what Chef is supposed to bring you into doing). This terrifies me quite a bit.

@thommay and I came to some decent compromises on some "lightweight chef" stuff, as he mentioned, and as long as it's actually getting converted to cookbooks under the hood I think it could be an easier sell than what we have today while still at least trying to show people the "right way".

I worry that what we're doing with stuff like this is taking people who don't understand the value of reproducible systems, telling them Chef gives that to them, then giving them a tool that **doesn't** do that, and thus they'll be angry when it doesn't.

I imagine something like this:

* I wrote some recipes to do some ssh tweaks
* I wrote some recipes to do some webserver setup
* I wrote some recipes to further tweak ssh
* I wrote some recipes to do some basic SA settings (maybe nsswitch or something)
* I loose some sometimes
* I build new ones and go "chef run <list of hosts> *.rb" and it fails miserably because there's no ordering, because some recipes assumed some manual-setup things, because whatever else
* I go "See Chef SUCKS!"
* Chef goes "Well that's not really how it's meant to be used"
* I go "... it's RIGHT THERE in the __chef__ command, wtf?"

I'm not -1ing this (yet?) but boy oh boy here there be dragons.

lamont-granquist:
Prior art:  https://github.com/chef/chef-dk/issues/452

Mostly I'd just ask that if we re-use `knife ssh` code that this be kept in `knife ssh` since that code is terribad and don't want it to infect everything else.

coderanger:
The specific use case I get asked for something like this a lot of webapp deployment (which I know more larger places will have some other system in place for). It's probably more of a `chef-client -o` type of thing though, like `chef run myapprecipe::deploy version=1.2.3` or something.

thommay:
Per the developer meeting, this RFC is withdrawn. When a plan follows for remote execution that will become a separate RFC, and should any work be desired on a more simple approach to Chef that should become another RFC.

PR #237, opened 2016-11-08
Added client-rb-logger.md

zealws:


thommay:
So I thought that we had a logger already by this point, and https://github.com/chef/chef/blob/db57131ad383076391b9df32d5e9989cfb312d58/lib/chef/config.rb#L28 seems to agree; whilst it might be a bit clunky, can you `ChefConfig.logger.warn("foo")` in your config?

new/client-rb-logger.md
```diff
@@ -0,0 +1,39 @@
...
+## Motivation
+
+    As a chef user,
+    I want to log output from client.rb into the same log file as the chef output,
+    so that client.rb output can be tracked in the same method as the chef run.
+
+## Specification
+
+1. Create a [Logger](https://github.com/chef/mixlib-log) instance before client.rb is initialized.
+1. Log messages on this logger will be written to stdout, and saved.
```
coderanger: "saved" is somewhat vague. Would you expect things to be buffered in memory? On disk? How do we handle an error during either config file loading or in core initialization soon afterwards?

new/client-rb-logger.md
```diff
@@ -0,0 +1,39 @@
...
+
+# Logging for client.rb
+
+Output from client.rb doesn't get logged to the log file specified with --logfile.
+If you want to save the output from client.rb along-side the chef logs, you need to redirect stdout, rather than using chef's built-in logging.
+I'd like to make Chef::Log accessible from the client.rb so messages logged during client.rb execution can be stored using chef's built-in logging.
+
+## Motivation
+
+    As a chef user,
```
coderanger: This isn't your real motivation I don't think. I'm guessing you are trying to do some debugging on values loaded during the config file? Whatever the real underlying motivation is, you should document it here.

new/client-rb-logger.md
```diff
@@ -0,0 +1,39 @@
...
+RFC: unassigned
+Title: Logging for chient.rb
+Author: Zeal Jagannatha <zeal@fb.com>
+Status: Draft
+Type: Standards Track
+Tracking:
+  - https://github.com/chef/chef/issues/5526
+---
+
+# Logging for client.rb
```
coderanger: Will this also apply to `solo.rb` and the knife config files?

coderanger (commented):
This feels very tailored to a Facebook use case, which isn't specifically a problem but I think we should zoom out a bit and make sure the implementation works beyond FB.

new/client-rb-logger.md
jaymzh: Nope. We want to log shit. We want to log which handlers we loaded. We want to log whether your in taste-tester or prod chef (and taste-tester is used various places outside of Facebook).

Today, we `puts "INFO: ..."` bla bla. and we have a wrapper to chef which we're looking to opensource which just runs `chef-client $options > /var/log/chef/chef.$DATETIMESTAMP` (among several other things).

I was discussing this with @lamont-granquist  at Summit and we agreed that about 90% of the things the wrapper does can/should be in core-chef. We started looking at the logging aspect first, and realized that we would lose any logging we make from client.rb which would be a bummer.

new/client-rb-logger.md
coderanger: What's the behavior if you log something in the config file and then later the config file raises an exception? I could totally buy either (major) approach (drop messages or try to get them to the real logger) but we should pick one :)

zfjagann: @coderanger The specific behavior I want is to buffer the log messages in memory, as well as printing them to stdout, just as they are done now with a `puts` statement. This way if the chef run dies before the real logger is initialized, the messages aren't totally lost.

new/client-rb-logger.md
```diff
@@ -0,0 +1,39 @@
+---
+RFC: unassigned
+Title: Logging for chient.rb
```
zfjagann: +1

new/client-rb-logger.md
zfjagann: solo.rb yes, since it's just a client.rb with a few special options.

I'm not sure about knife config files.

nathenharvey:
this should be ready for deciding at the next weekly developers' meeting.

nathenharvey:
We've agreed to put this RFC on hold until the spec is further fleshed out.

PR #238, opened 2016-11-15
Attributes dot Next

lamont-granquist:
This builds out a new API in parallel with the old API, with slightly different simplified semantics, emits deprecation warnings for cases where the semantics differ, and then gradually deprecates the old API.  I expect this is a 3+ year project at least, so the timing of when we actually deprecate the old APIs is not specified, and we should only pull the trigger there once the impact will be minimal.

This Expands/Replaces #77

### DIFF
--- a/attributes-next.md
+++ b/attributes-next.md
@@ -52,7 +52,7 @@ Deprecated Behavior:
   "force" behavior becomes the standard behavior.
 - Environment files consistently have a higher priority / later merge order than roles.
 
-Merged Precedence Levels Before:
+Merged Precedence Levels (Current API):
 
 | level | lowest  |  | | highest |
 | ---- | ------ | ------ | ---- | ---- |
@@ -63,7 +63,7 @@ Merged Precedence Levels Before:
 | force_override | attribute | recipe | | |
 | automatic| node |
 
-Merged Precedence Levels After:
+Merged Precedence Levels (New API):
 
 | level | lowest  |  | | highest |
 | ---- | ------ | ------ | ---- | ---- |
--- a/attributes-next.md
+++ b/attributes-next.md
@@ -52,6 +52,24 @@ Deprecated Behavior:
   "force" behavior becomes the standard behavior.
 - Environment files consistently have a higher priority / later merge order than roles.
 
+Merged Precedence Levels Before:
+
+| level | lowest  |  | | highest |
+| ---- | ------ | ------ | ---- | ---- |
+| default       | attribute | recipe | environment | role |
+| force_default | attribute | recipe | | |
+| normal | attribute | recipe | | |
+| override | attribute | recipe | role | environment |
+| force_override | attribute | recipe | | |
+| automatic| node |
+
+Merged Precedence Levels After:
+
+| level | lowest  |  | | highest |
+| ---- | ------ | ------ | ---- | ---- |
+| default       | attribute | role | environment | recipe
+| override | attribute | role | environment | recipe
+
 ## New Syntax for Ohai Attributes
 
 A new sugar for ohai attributes will be introduced:
--- a/attributes-next.md
+++ b/attributes-next.md
@@ -37,7 +37,7 @@ Overhauling Attributes to be simpler.
     We see consistent user confusion over what the point of force_default is.
 
     As a Chef Developer,
-    I've never seen anyone user o understand the role/environment precedence order swap between default and override.
+    I've never seen anyone use or understand the role/environment precedence order swap between default and override.
 
 ## Overall Design
 
--- a/attributes-next.md
+++ b/attributes-next.md
@@ -1,0 +1,107 @@
+---
+RFC: unassigned
+Title: Attributes v2.7182818284
+Author: Lamont Granquist <lamont@chef.io>
+Status: Draft
+Type: <Standards Track>
+---
+
+# Title
+
+Description and rationale.
+
+## Motivation
+
+    As a Chef User,
+    I want Attributes to be Less Confusing.
+
+    As a Chef User,
+    I dislike NoMethodError for NilClass Errors,
+    Because those are confusing and unhelpful.
+
+    As a Chef Developer,
+    The monolithic Node attributes API tries to do too much.
+
+## New Syntax for Ohai Attributes
+
+A new sugar for ohai attributes will be introduced:
+
+```ruby
+ohai("platform")
+ohai("cloud_v2", "public_ipv4_addrs", 0)
+```
+
+This will literally be sugar for the `node.automatic.read("platform")` API that has already been
+shipped.
+
+Write access to ohai data will continue through the existing `node.automatic` API.  In a future RFC
+that will be replaced.
+
+## New Syntax for Normal Attributes
+
+## New Syntax for Merged Default And Override Attributes
+
+## Removal of Method Missing Syntax
+
+A deprecation warning for method missing syntax access to the node attribute has already been added
+to Chef 12.  In Chef 13, method missing syntax will result in NoMethodError.
+
+## Deprecation of Array Merging
+
+Before the node.save the chef-client will walk the node object and warn on any array objects that are
+being merged with array merging.  Users who are actually using array merging will be encourage to replace
+array merging with hash keys instead.
+
+## Deprecation of Deep Merging with Normal Attributes
+
+Before the node.save the chef-client will walk the node object and warn on any normal attribute which
+also have default, override or automatic attributes set and are being merged.  In a future major release this
+deep merging will be removed.
+
+## Deprecation of Deep Merging with Automatic Attributes
+
+Before the node.save the chef-client will walk the node object and warn on any automatic attribute which
+also have default, normal or override attributes set and are being merged.  In a future major release this
+deep merging will be removed.
+
+## Deprecate Access to Automatic Attributes to the Node Object
+
+Rubocop rules will be added to cookstyle to warn when common automatic node attributes are being accessed
+through (e.g.) `node["platform"]` and will have an autofix rule to change them to (e.g.) `ohai("platform")`.
+
+On access the chef-client will issue deprecation warnings to any access to the node object through the
+deep-merged view of the node object.
+
+## NOTE: Favoring Strings Over Symbols
+
+Symbols as keys or values will not be deprecated, however, canonically the preferred syntax will
+be strings.  The internal representation of attributes will remain VividMash (autovivifying and
+converting-to-strings).
+
+## NOTE: Favoring Splat-Argument Syntax over Bracket Method Chaining
+
+Attribute APIs will favor splat-args syntax over method chaining because of the trainwreck issues that
+method chaining produces, which inevitably leads to the "NoMethodError on NilClass" error messages that
+are deeply tied to the ruby language.  The root cause being that all user created objects in the ruby
+language are truthy and that nil and false are the only objects which can be false and we cannot construct
+wrapper objects which are falsey.  As a result in order to give useful error messages, we need to call
+a method which is passed the entire path the user is requesting, and cannot support method chaining.
+
+We will favor this shape:
+
+```ruby
+something("foo, "bar", "baz")
+```
+
+Over this shape:
+
+```ruby
+something["foo"]["bar"]["baz"]
+```
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.
--- a/attributes-next.md
+++ b/attributes-next.md
@@ -66,8 +66,24 @@ Since normal attributes have different semantics they will be changed to use `st
 
 ```ruby
 store_set("wordpress", "passwd").to("sekret1")
+store_set("wordpress") do |s|
+  s["passwd"] = "sekret1"
+end
 store("wordpress", "passwd")
 ```
+
+## Note on why the New Syntax has split setter/getters
+
+It isn't possible to have a fused setter/getter and to use method chaining.  This API does not work with ruby:
+
+```
+default("foo", "bar", "baz").to("quux")
+default("foo", "bar", "baz")
+ => "quux"
+```
+
+The problem is going to be values like nil and false which would need to be wrapped to implement `#to` methods in
+a decorator to allow writing which would be a truthy return value for the getter and not falsey.
 
 ## Removal of Method Missing Syntax
 
--- a/attributes-next.md
+++ b/attributes-next.md
@@ -22,6 +22,9 @@ Description and rationale.
     As a Chef Developer,
     The monolithic Node attributes API tries to do too much.
 
+    As a Maintainer of Existing Cookbooks,
+    I need a slow and backwards compatibile transition to any new API.
+
 ## New Syntax for Ohai Attributes
 
 A new sugar for ohai attributes will be introduced:
--- a/attributes-next.md
+++ b/attributes-next.md
@@ -1,14 +1,14 @@
 ---
 RFC: unassigned
-Title: Attributes v2.7182818284
+Title: Attributes v.Next
 Author: Lamont Granquist <lamont@chef.io>
 Status: Draft
 Type: <Standards Track>
 ---
 
-# Title
+# Attributes v.Next
 
-Description and rationale.
+Overhauling Attributes to be simpler.
 
 ## Motivation
 
--- a/attributes-next.md
+++ b/attributes-next.md
@@ -82,6 +82,9 @@ Merged Precedence Levels (New API):
 | default       | attribute | role | environment | recipe
 | override | attribute | role | environment | recipe
 
+- automatic attributes are no longer merged so are not shown in the above table (they still exist)
+- `force_default/force_override` becomes the way attributes are applied in recipes
+
 ## New Syntax for Ohai Attributes
 
 A new sugar for ohai attributes will be introduced:
--- a/attributes-next.md
+++ b/attributes-next.md
@@ -27,6 +27,18 @@ Overhauling Attributes to be simpler.
     As a Ohai Plugin Developer,
     I want to be able to add ohai attributes without clobbering someone's cookbook attributes.
 
+    As a Chef Support Person,
+    We see normal attribute persistence cause consistent user confusion.
+
+    As a Chef Support Person,
+    We see array merging (sometimes) to cause consistent user confusion, with most users being surprised by merging behavior.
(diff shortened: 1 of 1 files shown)


lamont-granquist:
I haven't specified how this looks against the Chef::Node object as methods or how the syntax looks in attribute files because I haven't figured it out yet.  Right now this is just the methods injected into the Recipe DSL and trying to figure out how that looks.  I've been thinking that we should construct a proxy DSL object for attribute file to eval against so that we're not just injecting every Chef::Node method into that API.

There's a bit of a tension between keeping everything associated with the Chef::Node object and breaking it apart, and I don't know what works best with users.  I wrote this up looking at the Node object and finding it to be too much of a god-object, so I wanted to split it up into multiple distinct, non-overlapping APIs.  The problem that causes is it certainly makes it more difficult to understand what components ultimately go into the node object.

lamont-granquist:
Probably the scariest change in here is promoting all attribute setting in recipe files to force_default/force_override levels.  However by doing that you eliminate the need to have those as separate concepts, which is really huge for users.  The goal is that we don't get "wtf is force_default and force_override for?" questions any more, and users don't get surprised when setting attribute code in recipes turns out to the be the lowest precedence level.  The problem is that it is very difficult to gauge the amount of disruption that change will cause without actually doing it.

lamont-granquist:
I'll try to work up a couple tables in markdown of the attribute precedence graph before + after this change

attributes-next.md
```diff
@@ -0,0 +1,218 @@
...
+## New Syntax for Normal Attributes
+
+Since normal attributes have different semantics they will be changed to use `store` and will not deep merge with anything:
+
+```ruby
+store_set("wordpress", "passwd").to("sekret1")
+store_set("wordpress") do |s|
+  s["passwd"] = "sekret1"
+end
+store("wordpress", "passwd")
```
rhass: How about `property`?

lamont-granquist: `store` is actually a getter, part of the problem is that sounds like its very active verby.  `assign` has the same problem.  `property` collides with the resource concept.

thommay: `disinter`

jaymzh: why not store_get and store_set?

lamont-granquist: then you get`attr_get`, `default_get`, etc.  makes those kind of uglier.

jaymzh: OK, what about `persistent` and `persistent_set`?

lamont-granquist: that's getting better, its not as verby and "persistence" is right there in the name...

attributes-next.md
```diff
@@ -0,0 +1,249 @@
...
+mix with the default/override hashes written in attribute files.  Also note that env_default and role_default have been
+swapped for consistency (the ordering will consistently be attributes/roles/environments/recipes).
+
+# New Syntax for Accessing Default and Override Levels
+
+The writer uses `default_set` and `override_set` while the reader uses simply `default` and `override`.
+
+```ruby
+default_set("foo", "bar", "baz").to("quux")
+default_set("foo", "bar") do |n|
```
lamont-granquist: hmmm `default("foo", "bar", "baz") << 'more'` will have to work since we have to return a mash object and it won't be immutable.  we can also return a decorator for `default_set` which implements `#to` in the decorator and then allows for `default_set("foo", "bar", "baz") << 'more'` as well.

attributes-next.md
lamont-granquist: `sticky` and `sticky_set`?  (using a thesaurus)

attributes-next.md
```diff
@@ -0,0 +1,252 @@
...
+`store("wordpress", "passwd")`.
+
+On access the chef-client will issue deprecation warnings to any access to normal attributes through the
+deep-merged view of the node object.
+
+## Future:  Deprecate Access to Attributes Through the Node Object
+
+In a future version of Chef (no sooner than Chef 14) access to default and override objects through the node
+object APIs will be deprecated completely in favor of the new APIs.  In the following major release (no
+sonner than Chef 15) the old APIs will be disabled.
```
tas50: sooner

nathenharvey:
this should be ready for deciding at the next weekly developers' meeting.

attributes-next.md
shortdudey123: IMO this should be Chef 14 since it will probably be deprecated as of 13.  This gives people 1 year to remove the old API calls.

Counterpoint: this might require a bit of changes to many third-party plugins and tooling

attributes-next.md
jaymzh: Yes.

attributes-next.md
```diff
@@ -0,0 +1,252 @@
...
+
+Three Separated Namespaces:
+
+- Ohai attributes
+- Default and override attributes
+- Normal/storage attributes
+
+Deprecated Behavior:
+
+- Array deep merging is removed (must be manually replaced with hash-key merging)
```
jaymzh: Can you expand on what this means? with the change to force-by-default, I THINK this means:

```
default_set('foo', 'bar').to(['baz', 'tree'])
override_set('foo', 'bar').to(['thing', 'thong'])
attr('foo', 'bar')
```

would result in `['thing', 'thong']` ?

attributes-next.md
jaymzh: Please update the RFC with how you expect arrays to work, explicitly. This is critical for us and I don't want to end up in a place where we end up adding crazy crap like 'force_default!` stuff for the people using this use-case.

attributes-next.md
```diff
@@ -0,0 +1,252 @@
...
+- Deep merging of normal values with default and override is removed (users who were incorrectly using normal
+  as a third precedence level will need to manually fix)
+- Deep merging of automatic values with default/normal/override is removed (this is likely just namespace
+  collision mistakes and stright-up bugs?)
+- The force_default/force_override levels are now the standard behavior of setting attributes in recipe code, the
```
jaymzh: You say that, but do you mean `force_default` or `force_default!`?

The piece that was missing when we added these was:

```
node.default['foo']['bar'] = 'stuff'
# I can clear it!
node.override['foo'] = {}
node['foo']
=> {'bar' => 'stuff'}
# Fuck, what about this:
node.force_override['foo'] = {}
node['foo']
=> {'bar' => 'stuff'}
# Damnit! Let's try this!
node.force_override!['foo'] = {}
node['foo']
=> {}
# finally!
```

So if you drop force_default and force_override, you need to either **keep** force_default! and force_override! which sucks, or make `force_default!` and `force_override!` the default.

As a reminder the difference is:
* force_x - Nuke all instances in X precedence (roles_x, cookbooks_x, environments_x)
* force_x! - Nuke all instances in X precedence *and* all lesser precedences.

Please do not break our model again.

attributes-next.md
```diff
@@ -0,0 +1,252 @@
...
+| override | attribute | recipe | role | environment |
+| force_override | attribute | recipe | | |
+| automatic| node |
+
+Merged Precedence Levels (New API):
+
+| level | lowest  |  | | highest |
+| ---- | ------ | ------ | ---- | ---- |
+| default       | attribute | role | environment | recipe
+| override | attribute | role | environment | recipe
```
jaymzh: You're not clear here if you're having two storage locations (default and override) or 6 storage locations (default_attr, default_role, default_env, default_recipe, force_attr, force_role, force_env, force_recipe)...

If it's just 2 then all of the normal hash stuff will magically work again and regarding my comment above, force_x doesn't mean anything anymore and the default is force_x! However, if it's 6, then now you need both.

If it's just 2, then while I'm happy you will break the (IMO bad) use-case that Chef made all the Chef 11 changes for in the first place (which while I'm OK with it, others may not be)... which is users expect this craziness to work:

in cookbooks/foo/attributes/default.rb
```
default['foo']['dir'] = '/etc/foo'
default['foo']['file'] = 'foo.conf'
default['foo']['config_file'] = "#{node['foo']['dir']}/#{node['foo']['file']"
```
then in a role file:
```
default_attribute 'foo' => { 'dir' => '/etc/anotherfoo' }
```

And then expecting `node['foo']['config_file']` to be `/etc/anotherfoo/foo.conf` in some recipe. You can only make this work with multiple storage locations per precedence... but as I think we've seen... there be dragons.

So all this comes down to you need to **either**:

1. Call out in this RFC the above behavior is no longer supported and people will need to build `node['foo']['config_file']` later (and have only 2 storage locations)
**or**
2. Support a force_default! and override_default! behavior.

attributes-next.md
```diff
@@ -0,0 +1,252 @@
...
+- default (attribute files)
+- role_default
+- env_default
+- force_default (recipe files)
+- override (attribute files)
+- role_override
+- env_override
+- force_override (recipe files)
+
+Note (below) that attributes written in recipe files will go into the force_default/force_override hashes and will not
```
jaymzh: OK so you're calling out here the different storage facilities, but then you need to support force_override! (force_default! and force_default do the same thing since they are the lowest precedent)

jaymzh:
Maybe not, @nathenharvey  :)

attributes-next.md
lamont-granquist: Yeah, I'm :-1: on doing this in Chef 13 -- this is literally rewriting every use of every attribute in every cookbook, and a year is going to be too short.   I deliberately want "at least two years" to be the timeframe, allowing us the ability to revisit this in 2 years and then bump the time up longer if we need to.  I am trying to avoid a "Chef 11 attributes change" or "Python 3" event.  I'm also not planning on adding the deprecation notices to the old APIs until Chef 14.0.0 lands.  That means that throughout the lifetime of 13.0.0 both APIs are expected to be supported -- this is the "early adoption period".  Then the old one is deprecated in Chef 14.0.0, then Chef 15.0.0 is the first time we could drop the old APIs hard.

attributes-next.md
lamont-granquist: I need coffee before I fully tackle this, but the idea is still 8 locations.

I think it still works fine, because node.force_default!() setters will still obliterate all of the default precedence levels and then set the force_default level (those 8 precedence levels in the 'new' API still are *exactly* the same subcomponent VividMashes that underly the existing API -- same objects but with a different API to merge them).  I'll need to think about naming, but I think `node.force_default!()` will do exactly what you want with no changes.  We could even just keep that API there, since its something that most people don't use a lot.

lamont-granquist: And to maybe be a bit clearer with examples.  In *recipe* code:

```ruby
default_set("foo", "bar") = "baz"
```

That will literally boil down to `node.attributes.force_default.write("foo", "bar", "baz")`

Then when you do this:

```
node.force_default!("foo", "bar", "quux")
```

it will still delete that key out of `node.attributes.default/env_default/role_default/force_default` and set the `node.attributes.force_default` level.

we could introduce an alias for `default_set!()` to `node.force_default!`  (same with override)

attributes-next.md
shortdudey123: Ah gotcha, agree with it as written then

attributes-next.md
lamont-granquist: Yeah, this will still work:

```
default_set("foo", "bar").to "stuff"   # winds up in the force_default hash now
node.force_default!("foo", {})           # still clears default/env_default/role_default/force_default and sets force_default
node("foo")
=> {}
```

attributes-next.md
lamont-granquist: It already does.   Which is why this shit is so confusing.

Array merging only happens between "same" precedence levels -- default to default, or override to override:

attributes/default.rb:
```
default["foo"]["bar"] = [ 80 ]
```

recipes/default.rb:
```
default["foo"]["bar"] = [ 8080 ]
pp node["foo"]["bar"] #  => [ 80, 8080 ]
force_default["foo"]["bar"] = [ 8181 ]
pp node["foo"]["bar"] #  => [ 80, 8080, 8181 ]
node.override["foo"]["bar"] = [ 25 ]
pp node["foo"]["bar"] #  => [ 25 ]
```

lamont-granquist:
I'd like some feedback on an alternative syntax form.

We'd like to do this but cannot because $RUBY:  `default_set("foo", "bar") = "baz"` 

I also dislike this because of the need to mentally drop an equals sign before the last arg:   `default_set("foo", "bar", "baz")`

So we wound up here and have a need to create a decorator with a #to method to implement it and support it as a kind of getter to make `<<` and stuff work:  `default_set("foo", "bar").to("baz")`

But another possibility is named arguments to call out the last hash position:  `default_set("foo", "bar", to: "baz")`

IDK how this works with the array case, its possible that if the last argument is just a string that we make this work `default_set("foo", "bar") << [ "baz" ]` via just having it fall back to being a getter.

attributes-next.md
jaymzh: I'm fine with that, but this needs clarifying in the RFC.

attributes-next.md
jaymzh: Nope. So much nope.

You cannot in one breath say "everyone has to use the new syntax!" and then go "except here where you have to use node[...]"

jaymzh:
So why the `to` at all?

Why not have:

```
default_set('foo', 'bar') = "thing"
default_set('foo', 'bar') = ['a', 'b']
default_set('foo', 'bar') = {'a', 'b'}
default_set('foo', 'bar') << "stuff" # foo.bar is a string, we're appending
default_set('foo', 'bar') << "stuff" # foo.bar is an array, we're appending
default_set('foo', 'bar') += ['stuff'] # foo.bar is an array, we're appending.
```

In this cause if you then do:

```
default('foo', 'bar') = 'thing'
```

You can throw an error. default_set defines `=`, but default's `=` throws an error.

If that can't happen because $RUBY, then your `.to` syntax is way better then named args.

jaymzh:
To be clear, both cases of `<<` above need to work...

lamont-granquist:
you can't do `default_set('foo', 'bar') =` in ruby, there's no `:()=` operator.

jaymzh:
I'm fine with lack of `=` ... but there's a bunch of other questions here...

lamont-granquist:
The latter three cases should be entirely doable.  If the last arg isn't a hash with a `:to` key we'll just fall back to a getter in which case you just get a VividMash/Array and can operate on it normally.

jaymzh:
@lamont-granquist  what about all my other inline comments above?

attributes-next.md
jaymzh: @lamont-granquist  here.

attributes-next.md
jaymzh: @lamont-granquist  and here

attributes-next.md
jaymzh: @lamont-granquist and here

attributes-next.md
jaymzh: @lamont-granquist  and here.

attributes-next.md
jaymzh: @lamont-granquist  and here.

attributes-next.md
lamont-granquist: easily fixable by adding `default_set!()` to mean obliterate-all-default-level-values-and-then-set

lamont-granquist:
So the first part is we just need to add `default_set!` and `override_set!` then as aliases for `node.force_default!` and `node.force_override!` i can fix that up tomorrow.

> You're a bit contradictory about whether default_set() will work like force_default and clear the attributes_default data, or whether the attr() reader is just smart and tries to merge correctly.

`default_set()` will work like `node.force_default` (setting one precedence level)
`default_set!()` will work like `node.force_default!` (clearing everything from all default mashes and then setting the force_default level)

hopefully that's all clear.

> Because I'm pretty sure you stand to break us by splitting attribute and recipes into separate storage facilities (we tried to use force_default, but ended up not, and instead dropped all role data so that 'default' was the same storage location always).

This is a little curious.  And i can see where you're going with this because role and recipe shared hashes before and now they wouldn't and that has some subtle consequences.

Yep, not gonna solve that one tonight, I'm going to have to think about that one.  Its possible trying to kill off force_default/force_override is a bit of a bridge too far.

I think I finally *really* understand why you're so annoyed that attributes got broken up into a dozen different VividMashes as well...

(and no I don't think we can undo that)

We probably need `default_rm`, `override_rm` and `attr_rm` as well or something like that.   Again, that's just adding trivial aliases though.

attributes-next.md
jaymzh: Well, no because that's what you claim default_set already does - you said "force_default" is the default behavior in this world, and force_X says obliterate everything at my precedence and then set... and force_X! mmeans obliterate everything at my precedence and all lower precedences and then set... your document and argument are self-contradicting. You need to flesh this out.

jaymzh:
@nathenharvey - @lamont-granquist  and I have some more work to do to make this viable... let's push off reviewing this for a bit.

lamont-granquist:
So right now recipes have a lower precedence than roles or environments, and you have to reach for force_default if you want to fix that -- although now it has the side effect that you could overwrite attribute file settings completely with default attributes in recipes because that hash happened to not have been split up in Chef 11.

I'm unclear if there's any way to simplify this that doesn't mess everything up again.

We could still push the force_default VividMash as being the preferred one to set attributes in recipes, but then keep around the old method -- so `default_set()` in recipe code would be like `node.force_default[]` and then something like `attr_default_set()` in recipe code would be like `node.default[]`

Then attributes, roles, environments and recipe code would all have their own Mashes and recipe code would take precedence, but there would still be a way to edit the attribute-level mash from recipe code.   That would be conceptually somewhat simpler and the odd use case we have now would become the exception rather than the common path.  IDK if that makes things better or worse though.

I'm feeling kind of inclined now to drop trying to fight with eliminating force_default/force_override.

The thing that concerns me right now is that we've got a rule now that `node.default["foo"] = {}` will obliterate default attributes set in attributes files, but not in roles or environments, and which bypasses deep merging, which feels like a conceptual inconsistency to me, even though customers (like phil) might find it useful.

I'm open to suggestions, but I'm leaning more towards not trying to fix this problem.  That probably means we never fix this problem though.

jaymzh:
I don't hate that... but that means people who want do to the crazy
a = 'foo'
b = 'bar
c = a + b   # but I set b in a to 'sup' so I expect this to be 'foosup'

stuff that Chef11 attrs were designed around don't work magically unless they use `attr_default_set`. *Personally* I like this because then they have to invoke the special magic but I have a feeling naive customers will complain. I'm happy to sit down on a call with you to try and hack on ideas - after new years.

nathenharvey:
@lamont-granquist @jaymzh is this ready for review yet?  If not, may I suggest you add the 'on hold' tag so that we don't try to add it to the weekly agenda until it's actually ready for review?

PR #239, opened 2016-11-16
Shipping audit controls

thommay:

### DIFF
--- a/new/audit_phase.md
+++ b/new/audit_phase.md
@@ -29,15 +29,21 @@ Support audit and compliance checks in a technology agnostic way.
 
 ## Specification
 
-We will add a new directory, `audit`, to the top level of the cookbook.
-In this directory will be tests, typically but not necessarily written
-in Inspec, an operator can use to inspect the state of their system.
+As per RFC-35, audits will be run in the `audit` phase, which occurs
+after the `converge` phase completes.
+
+An audit handler will be introduced to allow cookbooks to register for
+the audit phase.
+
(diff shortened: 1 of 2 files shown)


chris-rock:
As discussed with @thommay:
1. we will have a audit phase in chef-client
2. we will remove the built-in audit mode, therefore the existing serverspec-based implementation will be externalized into a cookbook
3. we introduce a audit-handler to allow cookbooks to register for the audit phase. audit-handler are always executed in the audit phase and not during the converge phase
4. we should have the ability to upload the test directory with cookbooks to (chef-server), which is already covered in RFC 67 https://github.com/chef/chef-rfc/blob/master/rfc067-cookbook-segment-deprecation.md

jtimberman:
Per discussion in today's developer meeting, the new `audit` directory will be included in the cookbook artifact that gets shipped to, e.g., Supermarket. Tools that handle the upload / sharing of cookbooks will need to know about the audit directory. As one example, `stove` has a list of [acceptable files](https://github.com/sethvargo/stove/blob/master/lib/stove/packager.rb#L10-L25) that needs to be updated, and any versions of ChefDK that ship with this feature will need to ensure their stove version is updated too.

chris-rock:
@jtimberman I am not sure why we still need the audit directory. As discussed with @thommay (see result above) we fully opted into the AuditHandler approach instead of using a specific directory to identify a audit cookbook.

chris-rock:
@thommay Awesome. Thank you for the update. That perfectly reflects it.

thommay:

PR #240, opened 2016-11-16
Cookbook Metadata Lifecycle

charlesjohnson:
People have been running into more and more issues lately with cookbook sharing and distribution tools having different expectations around the presence of and preference for metadata.rb / metadata.json. 

This RFC seeks to clarify the behavior of Chef tools for handling these files. The goal is to provide a clear way to meet the needs behind the varying opinions expressed in the choices made by the tooling available.

### DIFF
--- a/new/cookbook_metadata.md
+++ b/new/cookbook_metadata.md
@@ -15,7 +15,7 @@ The purpose of this RFC is to clarify the handling of cookbook metadata files by
 
     As a person who writes and shares cookbooks,
     I want to specify metadata without hand-editing json,
-    so that I can write cookbooks without incurring the pain of hand-editing JSON.
+    because json is a serializable data format never intended for human interaction.
 
     As a person who uploads cookbooks to a Chef Supermarket server,
     I want to be able to upload my files without worrying about the format in which the cookbook metadata file was written,
@@ -26,24 +26,26 @@ The purpose of this RFC is to clarify the handling of cookbook metadata files by
     because this detail provides me with no tangible value.
 
     As a person who runs a Supermarket server,
-    I should be able to accept cookbook uploads without my server interpreting code as part of the uplaod process,
+    I should be able to accept cookbook uploads without my server interpreting code as part of the upload process,
     so that I can more easily provide a secure cookbook sharing service.
 
 ## Guiding UX Principles
-1. Machines should prefer `json` to `rb`
-2. Humans must never be forced to hand-edit `json` files.
+1. Machine interfaces should prefer `json` to `rb`
+2. Human interfaces must only require `rb` format metadata files.
+3. Humans must never be required to edit `json` files.
+4. Humans must never be required to care about the format of their cookbook metadata file.
 
 ## Specification
 
-- Supermarket, Chef server, depsolvers, et al. must only care about the metadata.json, and must ignore the metadata.rb. If no metadata.json is present, fail.
-- Tooling that takes cookbooks from machines and makes them available to humans for editing must generate a metadata.rb from the metadata.json, overwiting any metadata.rb that may already exist in the cookbook.
-
-- Tooling that takes cookbooks from humans and provides them to machines must generate a metadata.json from the metadata.rb, overwriting any metadata.json that may already exist in the cookbook.
+- Newly generated cookbooks must be created with only a metadata.rb file.
+- Newly generated cookbooks must chefignore the metadata.rb file, and gitignore the metadata.json file.
+- All server-oriented tools (Supermarket, Chef server, server-side depsolvers, et al.) must only care about the metadata.json, file, and must ignore the metadata.rb if present. If no metadata.json file is present, these tools must fail rather than trying to interpret the metadata.rb to generate a metadata.json.
+- Chef-client must only interpret metadata.json files. If a metadata.rb file is present, the chef-client may use the metadata.rb file to generate a metadata.json file at runtime.
+- Tooling that takes cookbooks from machines and makes them available to humans for editing must locally generate a metadata.rb from the metadata.json, overwiting any metadata.rb that may already exist in the cookbook.
+- Tooling that takes cookbooks from humans and provides them to machines must locally generate a metadata.json from the metadata.rb, overwriting any metadata.json that may already exist in the cookbook.
 
 - Processing of the metadata.rb to metadata.json must happen client-side, for security reasons.
 - Processing the metadata.json to metadata.rb can happen client-side or server-side, as there's no risk here.
-Add Comment
-
 
 ## Downstream Impact
 
--- a/new/cookbook_metadata.md
+++ b/new/cookbook_metadata.md
@@ -1,0 +1,61 @@
+---
+RFC: unassigned
+Title: Cookbook Metadata Files
+Author: Charles Johnson <charles@chef.io>
+Author: Thom May <tmay@chef.io>
+Status: Draft
+Type: <Standards Track>
+---
+
+# Cookbook Metadata Files
+
+The purpose of this RFC is to clarify the handling of cookbook metadata files by tooling in the Chef ecosystem.
+
+## Motivation
+
+    As a person who writes and shares cookbooks,
+    I want to specify metadata without hand-editing json,
+    so that I can write cookbooks without incurring the pain of hand-editing JSON.
+
+    As a person who uploads cookbooks to a Chef Supermarket server,
+    I want to be able to upload my files without worrying about the format in which the cookbook metadata file was written,
+    because this detail provides me with no tangible value.
+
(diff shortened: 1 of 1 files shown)


new/cookbook_metadata.md
```diff
@@ -0,0 +1,63 @@
...
+3. Humans must never be required to edit `json` files.
+4. Humans must never be required to care about the format of their cookbook metadata file.
+
+## Specification
+
+- Newly generated cookbooks must be created with only a metadata.rb file.
+- Newly generated cookbooks must chefignore the metadata.rb file, and gitignore the metadata.json file.
+- All server-oriented tools (Supermarket, Chef server, server-side depsolvers, et al.) must only care about the metadata.json, file, and must ignore the metadata.rb if present. If no metadata.json file is present, these tools must fail rather than trying to interpret the metadata.rb to generate a metadata.json.
+- Chef-client must only interpret metadata.json files. If a metadata.rb file is present, the chef-client may use the metadata.rb file to generate a metadata.json file at runtime.
+- Tooling that takes cookbooks from machines and makes them available to humans for editing must locally generate a metadata.rb from the metadata.json, overwiting any metadata.rb that may already exist in the cookbook.
```
lamont-granquist: this rule is caused by the original mistake of having tooling that strips the metadata.rb file out, and it feels like we're doubling down on that confusion.

we need a clearer rule here that metadata.json must always be favored, and metadata.rb must never be stripped (at least not by default) and that metadata.rb should be included in tarballs shipped to the supermarket.

the end users chefignore settings should really determine if they're pushing metadata.rb to their chef server or not (if they want an artifact repo were they can download cookbooks, mutate them and then re-upload then they'd want metadata.rb files uploaded, most other users would want them stripped by default).

stripping metadata.rb is an opinion and tooling should really be more neutral since not everyone shares the same opinion.

coderanger: I think this is a super bad idea, the JSON conversion is a lossy process and I don't think we should try and ignore that fact.

new/cookbook_metadata.md
```diff
@@ -0,0 +1,63 @@
...
+## Guiding UX Principles
+1. Machine interfaces should prefer `json` to `rb`
+2. Human interfaces must only require `rb` format metadata files.
+3. Humans must never be required to edit `json` files.
+4. Humans must never be required to care about the format of their cookbook metadata file.
+
+## Specification
+
+- Newly generated cookbooks must be created with only a metadata.rb file.
+- Newly generated cookbooks must chefignore the metadata.rb file, and gitignore the metadata.json file.
```
coderanger: This would make `knife download` unable to fully reconstruct the cookbook since the original `metadata.rb` would not be stored.

new/cookbook_metadata.md
```diff
@@ -0,0 +1,63 @@
...
+- All server-oriented tools (Supermarket, Chef server, server-side depsolvers, et al.) must only care about the metadata.json, file, and must ignore the metadata.rb if present. If no metadata.json file is present, these tools must fail rather than trying to interpret the metadata.rb to generate a metadata.json.
```
coderanger: Chef Server doesn't care about either ever, the only server-side thing that directly consumes the JSON as a file is Supermarket I think.

coderanger (commented):
Other than manual downloads from Supermarket lacking a metadata.rb sometimes (i.e. if uploaded via stove), is there another issue that needs to be addressed?

charlesjohnson:
I think the things that really matter here are:
- Current tooling is inconsistent in its handling of metadata.rb / metadata.json files.
- Because of these inconsistencies, humans are being forced to care and think about the format of their metadata file. 
- However, humans should never be forced to care or think about the format of their cookbook metadata file.
- Because one of the metadata file formats is json, humans are being forced to write and edit metadata.json files.
- However, humans should never be forced to write or edit JSON files.
- In addition, cookbook artifact repos should never interpret ruby metadata files to provide services.

Since the draft here doesn't seem to work for people, I'd love to hear thoughts about the simplest implementation that achieves the following goals:
- Humans don't have to care about the format of the metadata file in order to share and use cookbooks.
- Humans never have to directly edit or create json documents.
- Cookbook artifact repos should never interpret ruby metadata files to provide services.

Are there other things that matter in this conversation? Let's get them on the list, and work out how to meet all of those needs.

lamont-granquist:
* Tooling that generates metadata.json MUST not strip metadata.rb
* Tooling MUST favor reading metadata.json over metadata.rb when both are present
* Tooling MUST be able to read cookbooks with only metadata.rb
* Tooling MUST be able to read cookbooks with only metadata.json
* Authoring tools MAY regenerate metadata.json from metadata.rb for certain workflows (download->reupload)

coderanger:
@lamont-granquist To that I would add `Server side components MUST read only metadata.json` just to restate the obvious :)

lamont-granquist:
You mean the erlang?  Yeah, I guess that isn't ever going to parse metadata.rb, so there are some consumers that would never be able to meet the requirement of parsing that and MUST have metadata.json provided to them.

coderanger:
Also Supermarket or anything else you upload cookbooks to.

lamont-granquist:
updated

charlesjohnson:
Possible undesirable case:
- A person downloads a cookbook that contains both metadata.json and metadata.rb. 
- That person then customizes the cookbook, and modifies the metadata.rb accordingly, bumping the version.
- However, because the tooling favors the metadata.json, the changes are ignored by the tooling.

How can we avoid this and still keep:
> Tooling MUST favor reading metadata.json over metadata.rb when both are present?

shortdudey123:
I have seen the difference in the rb and json like @charlesjohnson described in several cookbook repos

coderanger:
@charlesjohnson The answer I've seen in some other tools is to favor the one with the newer modtime, but then that means tools need to preserve modtime and I don't think `knife download` currently does that :-/

robbkidd:
A thought, in addition to preserving mod-time of files within the generated cookbook artifact and when unpacking an artifact: 

"Tools should error when validating metadata if both `metadata.rb` and `metadata.json` are present and if `rb`'s mod-time is later than `json`."

Not unlike Gemfile/Gemfile.lock and Berksfile/Berksfile.lock.

lamont-granquist:
So I don't think we can fully prevent this issue.  The tools don't know that the user is going to tamper with the metadata.rb without regenerating the metadata.json, and have to assume that the metadata.rb cannot be evaluated in the current context to regenerate the metadata.json.  The user is responsible for modifying the metadata.rb and then calling the appropriate command to regenerate the metadata.json in the appropriate context.  The tools have to trust that any supplied metadata.json was derived from the supplied metadata.rb.

We probably should check the mod times, but we should only warn, which should be good enough.

charlesjohnson:
> The user is responsible for modifying the metadata.rb and then calling the appropriate command to regenerate the metadata.json in the appropriate context. The tools have to trust that any supplied metadata.json was derived from the supplied metadata.rb.

How could we validate that the supplied metadata.json was derived from the supplied metadata.rb, instead of trusting?

lamont-granquist:
We got into this whole problem because metadata.rb is pure-ruby and someone can go:

```
version MagicalVersionGenerator.gimme()
```

And then outside of that context, any other tool will likely not have that utility class and will explode with a syntax error.

Unless the user TELLS us that metadata.rb is cool to re-parse we can't reparse it for the user automatically.

Its also not sufficient to think that you can parse it correctly if it cleanly parses, I've seen stuff like:

```
version IO.read("/my/version.txt") rescue "0.0.0"
```

Which will not correctly regenerate the metadata.json if that file doesn't exist.

lamont-granquist:
https://github.com/chef/chef/pull/5622 has been merged and addresses the bug that really kicked this all off i think?

lamont-granquist:
So, for the record, I came around to being okay with mtime checking as long as its a WARN that metadata.rb was updated more recently than metadata.json.   It'll work most of the time, and give the user the tip to understand why they're getting weird results, and doing a WARN instead of a hard error will not block if the mtime's get garbled by something.

nathenharvey:
This is "on hold" until such a time that someone is interested in picking it back up and shepherding it  through the process.

lamont-granquist:
i think instead of mtime checking that we should most likely checksum the metadata.rb file that was used to create the metadata.json file and put that into the metadata.json, then it could be determined that the metadata.rb was 'tampered' with.

of course people who bump version information externally from the metadata.rb could still wind up in the situation where the external stuff has been bumped, but the metadata.json is being used instead of regenerated.  but the mtime approach has that same edge condition and more.  presumably people who are getting the version number from an external source of truth are going through a workflow which eliminates this risk (*cough* policyfiles *cough*).

PR #241, opened 2016-11-22
Add critical non-maintained cookbooks to RFC069

shortdudey123:
This adds provisions for transfering a cookbook on supermarket to a new owner in cases where the cookbook is major / critical and the owner is not reachable and not actively maintaining the cookbook.

### DIFF
--- a/rfc069-cookbook-adoption.md
+++ b/rfc069-cookbook-adoption.md
@@ -53,7 +53,7 @@ else, that is left to the cookbook's current owner.  See above for guidelines on
 when a cookbook has been placed up for adoption.
 
 We do recognize, however, that extreme circumstances do happen (i.e. death of the
-cookbook's current owner) and that sometimes a ciritical cookbook goes unmaintained.
+cookbook's current owner) and that sometimes a critical cookbook goes unmaintained.
 In either of these cases a community member who wishes to take ownership
 of the cookbook (or see ownership transferred to someone else with their
 consent) is welcome to file an [RFC](https://github.com/chef/chef-rfc).
@@ -62,7 +62,9 @@ In cases where a critical cookbook is no longer being actively maintained by the
 current owner (i.e. no commits on github and releases to supermarket for at least 6
 months with multple open issues and PRs), a best effort must be made to contact the
 author via at least 2 methods (i.e. email, slack, twitter) before opening an RFC to
-transfer ownership.
+transfer ownership.  If the current owner responds at any point before the transfer
+occurs, the transfer process must stop if it has been started.  It is then up to the
+current owner to decide if they with to proceed with the transfer themselves or not.
 
 The RFCs will be decided on by the Chef Community on a case by case basis.
 
--- a/rfc069-cookbook-adoption.md
+++ b/rfc069-cookbook-adoption.md
@@ -24,7 +24,7 @@ Type: Process
 ```
     As a supermarket user,
     When I believe a cookbook should be transferred away from the current owner,
-    Only in extreme cases (i.e. death of the current owner),
+    Only in extreme cases (i.e. death of the current owner, long outdated but a major cookbook),
     I should open an [RFC](https://github.com/chef/chef-rfc)
     And request that the cookbook ownership be transferred
 ```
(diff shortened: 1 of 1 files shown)


lamont-granquist:
👍

rfc069-cookbook-adoption.md
```diff
@@ -53,10 +53,17 @@ else, that is left to the cookbook's current owner.  See above for guidelines on
 when a cookbook has been placed up for adoption.
 
 We do recognize, however, that extreme circumstances do happen (i.e. death of the
-cookbook's current owner).  In this case a community member who wishes to take
-ownership of the cookbook (or see ownership transferred to someone else with their
+cookbook's current owner) and that sometimes a ciritical cookbook goes unmaintained.
```
josephholsten: `s/ciritical/critical/`

lamont-granquist:
Should probably clarify that this is for the case when the owner has gone completely radio silent.  If the owner responds and their response is 'no' then the transfer does not happen.

rfc069-cookbook-adoption.md
shortdudey123: Fixed

coderanger:
:-1: because this means Chef needs to decide what is or isn't maintained and what is "critical". It's open source, if you think something is unmaintained the fork it. Forced cookbook transfers are a nuclear option and shouldn't be used for inconveniences or disagreements.

shortdudey123:
Forking and adding the updated fork to supermarket under a slightly different name creates a lot of clutter and deplication

gsaslis:
@coderanger i agree with not needing Chef to decide. however, i also agree there is a problem here in the way supermarket currently works... it's really problematic to find some 'critical' cookbooks and really raises the bar for newcomers to Chef (I also faced this problem). 

perhaps the solution is not transferring cookbook ownership, but fixing the search in supermarket to (strongly) favour active cookbooks over inactive and 'unmaintained' ones in search results. at the moment it just seems to use text-based search, not scoring results using other parameters... ?

coderanger:
@gsaslis Yep, improved discovery for cookbooks is definitely the right solution IMO, and part of that is the quality metrics system.

@shortdudey123 It does create duplication, but that's the cost of being a neutral distribution platform. Supermarket is not intended to be a curated system in any real sense, it's a hosting platform. If you wanted to build out a list of curated cookbooks accounting for when forks happen and all that, :+1: but I don't think it should go in Supermarket.

shortdudey123:
Capturing some comments from the developer meeting:

-  "What is critical?" "By what criteria do we measure 'critical'?" "Who determines what is critical?"
- allow forced transfers in cases of negligence or security issues
- Who would maintain a list or idea of what is a critical cookbook?
- Unmaintained problem might be more of a supermarket metrics and discoverability issue
- cookbook adoption RFC as it stands covers the most serious issues of cookbooks being outdated/unmaintained

coderanger:
@shortdudey123 re: `allow forced transfers in cases of negligence or security issues` we already do allow that on a case-by-case basis as laid out in https://github.com/chef/chef-rfc/blob/master/rfc069-cookbook-adoption.md#when-a-cookbook-is-not-up-for-adoption-1. Basically if you think a specific cookbook is a big enough problem that it is worth writing an RFC, then we should talk about it :)

shortdudey123:
Ah, the way i was reading that, it only really covered the current owner dieing or being completely unable to maintain it vs just neglecting it.  So neglect / lack of maintenance would be covered under the existing RFC?

coderanger:
@shortdudey123 If it became an actual threat to the community, sure. You would have to show that it's a bigger deal than an annoyance and that forking would not fix the problem. For example if a cookbook was installing a database with a hard-coded backdoor password, that would be grounds for a forced transfer (or removal) as it is a threat to users of Supermarket. Someone not merging bug fixes is not a threat, its an annoyance but the code is not unsafe to use.

gsaslis:
@coderanger might I suggest changing: 
"(i.e. death of the cookbook's current owner)" to "(e.g. death of the cookbook's current owner)" to make it more clear that that's not all it's limited to?

coderanger:
@thommay You okay with the ^^ change? It's more than I would usually tweak with my Editor-y powers, but seems like a fix to me.

shortdudey123:
Closing since the overall view is the working in this is more trying to solve a supermarket discoverability problem

PR #244, opened 2016-12-02
Test Kitchen Multi-Server Testing

coderanger:
Okay so remember how I said last time that I would stop writing new RFCs that I don't have the bandwidth to implement? Well I lied. But I think this is worth writing down at least so we can start talking about something concrete.

This has been on the Test Kitchen to-do list for _years_ now. I think this design should cover all the use cases I've seen or heard of in the many many times this has been discussed at Summits, meetups, the mailing list, Slack, and probably plenty of other places.

I still really don't like the name `supersuites` but I'm purposefully leaving the terrible name there for now so people will suggest better ones. So far the suggestions I've seen: `multisuite`, `multi`, and `group`. There are also a bunch of `[ed:]` questions in the text, if you have opinions on any or all of them please leave line comments on them.

This is probably going to be a hell of a climb to implement all of it no matter how we do it, but every journey starts with a single step 👣

### DIFF
--- a/new/test-kitchen-multi.md
+++ b/new/test-kitchen-multi.md
@@ -294,6 +294,10 @@ All Test Kitchen drivers that want to support multi-server testing will need to
 be updated to include an implementation of the gossip phase, even if it might be
 a no-op in some cases.
 
+As the new functionality is only activated if one or more supersuites exist in
+the configuration, existing Test Kitchen users should be unaffected unless they
+opt in to the new features.
+
 ## Copyright
 
 This work is in the public domain. In jurisdictions that do not allow for this,
--- a/new/test-kitchen-multi.md
+++ b/new/test-kitchen-multi.md
@@ -1,0 +1,302 @@
+---
+RFC: unassigned
+Title: Test Kitchen Multi-Execution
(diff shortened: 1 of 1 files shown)


brandocorp:
`integrations` or `plans` might be good alternatives to the `supersuite`

cheeseplus:
`integration` as a term is already a bit loaded in the context of kitchen folder structure for tests. `plan` is a bit better but also don't want to folks to confuse with habitat.sh's plan.sh.

Naming things is still super hard but keep the comments and suggestions coming!

coderanger:
`plan` is probably the best I've heard so far. The overlap with Habitat is bad, but not the worst thing ever.

shortdudey123:
- `suite_plan`
- `test_plan`

(with or without `_`)

thommay:
Decided and approved. @chef/rfc-editors take it away!

coderanger:
To capture from the meeting, I will deal with picking a better name when I or someone else gets the cycles to implement this. Until then at least we have a plan on record and can use this for at least vague estimation on how long it will take.

adamleff:
Merged in 07f809d - thanks!

PR #245, opened 2016-12-06
RFC for removing unused metadata fields

thommay:


danielsdeleo:
Oh man, I've wanted this for a long time.

new/unused_metadata.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Specification
+
+Mark as deprecated and subsequently remove the `recommends`, `suggests`,
+`conflicts`, `replaces` and `grouping` metadata fields.
+
+## Downstream Impact
+
+Chef Server should be updated to not serve this metadata, docs need to
+be updated.
```
danielsdeleo: Generators, too.

lamont-granquist:
I have considered that we could start using `recommends` for cookbooks like `windows` and then have a config setting like `Chef::Config[:without_recommends] = [ "windows" ]` on non-windows machines to avoid synchronizing the files.  Although the direction we seem to be heading is towards baking that functionality into core and making cookbook synchronization faster.

lamont-granquist:
We've also had warnings in foodcritic to avoid recommends/suggests for some time now:  https://github.com/acrmp/foodcritic/commit/aa554ce2b8b64a2ddb122981a30126431770612b

tas50:
It's also been removed from a lot of the documentation so new users don't find it.

coderanger:
We had also talked for a while about making one of those (`suggests`?) be "not an error if isn't present but if it is present it will be treated as a dependency for load ordering" but I agree that has had no motion in years and no reason to block on it :)

lamont-granquist:
The complaining about that problem seems to have reduced in the past year or two.  I'm not sure if that is due to improvements in cookbook synchronization making it not as much of an issue, or if its just user's stockholm syndrome tho...

thommay:
Yeah, the noise around windows definitely seems to have gone down dramatically.

new/unused_metadata.md
```diff
@@ -0,0 +1,42 @@
...
+Mark as deprecated and subsequently remove the `recommends`, `suggests`,
+`conflicts`, `replaces` and `grouping` metadata fields.
+
+## Downstream Impact
+
+Chef Server should be updated to not serve this metadata, docs need to
+be updated. Also, generators such as `chef generate`, `berks cookbook`
+need to be updated to ensure they do not create metadata files with
+those fields.
+
```
marcparadise: Should chef server continue to capture these values when they're present, or silently drop them if they're provided?

thommay: i see no harm in capturing them.

coderanger: I think Chef Server should store and send back any arbitrary key/value gunk we send it since that makes things more flexible for the future, but `knife` and `berks` should probably stop sending these.

nathenharvey:
@chef/rfc-editors this was approved in the 2017-01-19 meeting.

thommay:
Merged as RFC 85.

PR #247, opened 2016-12-20
Fix README for generating GH Pages

adamleff:
The Rake task in the README is listed as `generage:pages`, but it's
actually `pages:generate`.


thommay:
thanks!

PR #249, opened 2017-02-16
add support windows for cookbooks and ruby versions

lamont-granquist:
We want to also establish a cadence for the support windows of ruby and cookbooks.

I felt like it made logical sense to me to put that here.

### DIFF
--- a/rfc081-release-cadence.md
+++ b/rfc081-release-cadence.md
@@ -98,6 +98,19 @@ month old Ruby versions -- long after the Ruby community had stopped support.  T
 sliding windows here means that the support windows for Chef and Ruby will be more
 synchronized.
 
+The combined sliding window of Chef Client support of Ruby and cookbook and ecosystem
+support of Chef Client needs to better match the Ruby language support window.  As
+external gem authors tend to drop support of Ruby versions that fall into security-only
+releases, we also plan on dropping support of Ruby versions once they fall into
+security-only releases.  For example, in Q1 of 2016 ruby-2.3.0 had been released,
+ruby 2.2.x was still being maintained and ruby-2.1.x fell into security-only releases,
+which caused many rubygems -- most notably rack-2.0 and activesupport-5.0 to later
+drop support for ruby-2.1.x.  This caused issues everywhere in the Chef ecosystem
+as Chef Client had stayed on ruby-2.1 until it became a crisis, and we were forced
+to work around supporting ruby versions that the rubygems community had dropped
+support for.  The Ruby, Cookbook and Ecosystem support cadence is design specifically
+to avoid this situation.
+
 ## Copyright
 
 This work is in the public domain. In jurisdictions that do not allow for this,
--- a/rfc081-release-cadence.md
+++ b/rfc081-release-cadence.md
@@ -49,7 +49,7 @@ elect to drop support for older chef-client versions.  This window does not rese
 a major version release so that the prior major version track is supported for a 6 month
 window.  Non-Chef managed community cookbooks are encouraged to follow this policy.
 
-As an example, in May we will typically drop 14.1.0 and both 14.0 and 14.1 will be
+As an example, in May we will typically release 14.1.0 and both 14.0 and 14.1 will be
 supported.  The version of 13 release 6 months prior should be 13.8 and will still
 be supported so that 6 versions will be considered current (13.8 through 13.11 plus
 14.0 and 14.1).  At that point community cookbooks may choose to start using 13.8
@@ -57,7 +57,7 @@ features and drop support for versions prior to 13.7
 
 ## Ruby Cadence
 
-Since the ruby language itself releases new minor versions over the Christmas holidays,
+Since the Ruby language itself releases new minor versions over the Christmas holidays,
 the April major release of Chef Client should include the minor revision of ruby which
 landed the prior Christmas.  Combined with the 6 month sliding window for cookbook
 support that also implies that when the prior major release of the client falls off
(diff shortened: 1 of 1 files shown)


lamont-granquist:
The rationale for including a cadence of ruby language releases can probably be summed up by this bug:

https://github.com/chef/cookstyle/issues/19#issuecomment-280218806

Since we dragged out bumping from ruby 2.0 to 2.1 and then 2.1 to 2.3 for as long as possible that also means that we now have to support those old ruby versions in cookbooks for long after they've moved off actively being supported by the core language team.  If we start bumping those with a better cadence then we can stay current in tooling like cookstyle better.

rfc081-release-cadence.md
```diff
@@ -41,6 +41,32 @@ potential hot fixes or follow-up.
 Major releases in April avoids releasing during winter holidays, summer
 vacations, ChefConf and Chef Summits.
 
+## Cookbook Support
+
+The latest version of community cookbooks are required to support only the latest 6
```
coderanger: Maybe say "encouraged to support at least the latest 6" since this is just a spec and we can't enforce adherence.

rfc081-release-cadence.md
```diff
@@ -41,6 +41,32 @@ potential hot fixes or follow-up.
...
+
+As an example, in May we will typically drop 14.1.0 and both 14.0 and 14.1 will be
+supported.  The version of 13 release 6 months prior should be 13.8 and will still
+be supported so that 6 versions will be considered current (13.8 through 13.11 plus
+14.0 and 14.1).  At that point community cookbooks may choose to start using 13.8
+features and drop support for versions prior to 13.7
+
+## Ruby Cadence
+
+Since the ruby language itself releases new minor versions over the Christmas holidays,
```
coderanger: Minor nit: `Ruby` because proper noun :)

rfc081-release-cadence.md
```diff
@@ -41,6 +41,32 @@ potential hot fixes or follow-up.
 Major releases in April avoids releasing during winter holidays, summer
 vacations, ChefConf and Chef Summits.
 
+## Cookbook Support
+
+The latest version of community cookbooks are required to support only the latest 6
+months of chef-client versions.  This window does not reset on a major version release
+so that the prior major version track is supported for a 6 month window.
+
+As an example, in May we will typically drop 14.1.0 and both 14.0 and 14.1 will be
```
iennae: Can we replace "drop" here for "release" or similar word. Drop can be used to mean multiple things and in this section we are talking about dropping support and then using it in a sentence to mean release.

lamont-granquist:
I updated the cookbook 6 month window to be more specific.

It has words now directed specifically at Chef-managed community cookbooks, which are expected to follow a 6-month support window for client versions.  I very deliberately chose SHOULD over MUST because I can forsee some exceptions to that rule -- mostly dealing with greenfield cookbooks.  For the section on dropping support after the 6-month window I also used MAY very deliberately because if a cookbook can easily support 12-month old chef there's no reason to break it just because of this RFC.

And then I just but in "encouragement" for other authors to follow this rule.

adamleff:
Given the soon-to-be-approved RFC detailed in #250, the developer community offered these suggestions during the March 2, 2017 community meeting:

 * since the RFC in #250 replaces RFC 81, the Ruby version support semantics should perhaps be copied into the RFC proposed in #250 
 * the cookbook support windows should become its own PR, or perhaps this PR simply is refactored to focus only on the cookbook support windows

lamont-granquist:
I also added more Rationale sections targeting the cookbook and ruby language cadences.

tduffield:
Since #250 aims to replace RFC-081, do we want to break some of this out into a new "Dependency Update Cadence" RFC and put the rest in #250? Something we can use to describe the method and rationale behind how we update and support dependencies?

lamont-granquist:
IDK i'd like to just accept this and then work on #250

rfc081-release-cadence.md
```diff
@@ -41,6 +41,76 @@ potential hot fixes or follow-up.
...
+
+The release of the new major version may be delayed if there are show stopping bugs
+in the released version of Ruby (we assume that 4 months will be enough time for
+major regressions in the core language to be addressed, but that is an external
+dependency).
+
+If the Ruby language version released over Christmas has a show-stopper bug then the
+next major Chef Client version may be released without it.  It can then be included
+in a subsequent minor version bump.  This RFC deliberately uses 'should' instead of
+'must', and [RFC-034](https://github.com/chef/chef-rfc/blob/b7bd9c53bf96235f9334e65bb5848f7843c81fed/rfc034-ruby-193-eol.md#specification)
```
thommay: this should link to https://chef.github.io/chef-rfc/rfc034-ruby-193-eol.html

lamont-granquist: i actually did that deliberately so that if we edited that RFC to say something different that we'd follow the link to what the text said when this was written.

i can see the point of referencing the current RFC, but if that RFC was updated with incompatible language, then with a link that tracked the current version it just becomes deeply confusing if you follow the link.  if you follow the link to an old RFC it becomes more apparent what happened.

lamont-granquist:
I think we merge this as-is and when #250 is merged then the top section of this becomes a link to that, and this RFC becomes the "Dependency Update Cadence" RFC.

thommay:
Approved with the proviso it gets turned into a shiny new RFC rather than an update to 81.

thommay:
Superseded by #267

PR #252, opened 2017-03-09
add some critical chef-sugar helpers

lamont-granquist:
Signed-off-by: Lamont Granquist <lamont@scriptkiddie.org>

### DIFF
--- a/new/rfcXXX-distro-sugar-helpers.md
+++ b/new/rfcXXX-distro-sugar-helpers.md
@@ -1,0 +1,186 @@
+---
+RFC: unassigned
+Title: Chef Sugar Platform Matchers in Core Chef
+Author: Lamont Granquist <lamont@chef.io>
+Status: Draft
+Type: Standards Track
+Tracking:
+  - https://github.com/chef/chef/pulls/< FIXME: TBD VERY SOON >
+---
+
+# Title
+
+Merging chef-sugar platform and `platform_family` helpers, along with higher level helpers.
+
+## Motivation
+
+    As a Cookbook Author,
+    I want the Chef Sugar helpers added to Core,
+    So that everyone can use them.
+
+    As a Chef Developer,
+    The platform_family abstraction is limiting,
+    A set of helper functions provides better abstractions.
+
+    As a Cookbook Author,
+    Helper methods produce cleaner case statement,
+    Which produces more readable and maintainable code.
+
+    As a Chef Developer,
+    Helper methods can provide cleaner abstractions,
+    Which should produce less need to break APIs.
+
+## Motivation
+
+There's considerable argument over what a `platform_family` should be, as evidenced by the
+discussion over the amazon `platform_family` issue.  At the root of the problem seems to be
+that every distro must belong to precisely one `platform_family` while the classification of
+different distros forms more of an inheritance hierarchy.  The argument seems to be over at
+which level of the hierarchy the `platform_family` should apply.
+
+Helper methods solve this problem by allowing a given distro to respond as true to different
+helper methods.
+
+The chef-sugar helper methods have also found widespread adoption in the community and its
+time to pull them into core.
+
+## Scope
+
+This RFC covers only the `platform` and `platform_family` helpers in chef-sugar along with
+some other badly needed helpers.  The `platform_version` style of helpers (`debian_before_squeeze`)
+are out of scope.  The limiting of scope has no implications as to if we eventually want more
+of chef-sugar into core or not (we certainly do want more of it in core), and is solely to limit
+the scope of this work and the discussion.  There do exist pieces of chef-sugar (like the
+`method_missing` style of attribute syntax) that we explicitly do not want, and each piece of
+chef-sugar needs to be assessed on its own merits.
+
+## Specification
+
+We will pull in most of the Architecture, Platform and Platform Family helpers from chef-sugar
+unmodified, along with the `systemd?`, `docker?` and `kitchen?` helpers.  Since they have been
+in widespread use, it is important to not modify their meaning.
+
+### Architecture Helpers
+
+* `_64_bit?`
+* `_32_bit?`
+* `i386?`
+* `intel?`
+* `sparc?`
+* `ppc64?`
+* `ppc64le?`
+* `powerpc?`
+* `armhf`?
+* `s390x`?
+
+### Platform Helpers
+
+The complete list of platform helpers is long, examples:
+
+* `amazon_linux?`
+* `centos?`
+* `linux_mint?`
+* `oracle_linux?`
+* `redhat_enterprise_linux?`
+* `scientific_linux?`
+* `ubuntu?`
+* `solaris2?`
+* `aix?`
+* `smartos?`
+* `omnios?`
+* `raspbian?`
+* `nexus?`
+* `ios_xr?`
+
+Note that this list deviates from the platform definitions in core chef, to eliminate ambiguity we will also
+add helpers which precisely match the field in core chef:
+
+* `linuxmint?`
+* `oracle?`
+* `scientific?`
+* `redhat?`
+
+Additional helpers will also be added to fill out the complete list of platforms, which will match the fields
+used by core chef (matching the code in ohai).
+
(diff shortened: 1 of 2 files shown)


new/rfcXXX-distro-sugar-helpers.md
```diff
@@ -0,0 +1,156 @@
...
+* `kitchen?`
+
+The `docker?` helper has already been merged as of Chef 12.11.18
+
+### Higher Level Helpers
+
+In addition to the chef-sugar helpers, higher level helpers will be added:
+
+* `rpm_based?` -- this will include the `platform_family` of `rhel`, `fedora`, `amazon`, and `suse`
+* `fedora_based?` -- this will include the `platform_family` of `rhel`, `fedora`, and `amazon`
```
thommay: This one definitely gives me the shivers.

thommay:
In this universe, what does:
```
case node[:platform_family]
```
become? Big ass `if` blocks? Or can we do better?

new/rfcXXX-distro-sugar-helpers.md
```diff
@@ -0,0 +1,159 @@
+---
+RFC: unassigned
+Title: My Title
+Author: Alan Smithee <asmithee@example.com>
```
schisamo: All this metadata looks wrong...

new/rfcXXX-distro-sugar-helpers.md
```diff
@@ -0,0 +1,159 @@
...
+Status: Draft
+Type: <Standards Track, Informational, Process>
+<Replaces: RFCxxx>
+<Tracking:>
+<  - https://github.com/chef/chef/issues/X>
+---
+
+# Title
+
+Merging chef-sugar platform and `platform_family` helpers, along with higher level helpers.
```
schisamo: Any reason we just don't bring chef-sugar into core chef instead of cherry picking some helpers?

lamont-granquist: i really hate the attributes method_missing syntax for one, that's a major :-1: because anything method_missing needs to die in a fire.

lamont-granquist:
thommay:

```ruby
case
when rpm_based?
when debian?
when solaris_based?
when gentoo?
when arch?
end
```

I'm slightly sick so I'm not positive that's the best case to use to illustrate, but the point there is that there's a flat case statement without nesting which mixes-and-matches at different 'levels' of the tree -- rpm-based is very high level, debian matches platform_family, gentoo and arch are picking out individual linux distros to special-case.

new/rfcXXX-distro-sugar-helpers.md
```diff
@@ -0,0 +1,159 @@
...
+The `docker?` helper has already been merged as of Chef 12.11.18
+
+### Higher Level Helpers
+
+In addition to the chef-sugar helpers, higher level helpers will be added:
+
+* `rpm_based?` -- this will include the `platform_family` of `rhel`, `fedora`, `amazon`, and `suse`
+* `fedora_based?` -- this will include the `platform_family` of `rhel`, `fedora`, and `amazon`
+* `solaris_based?` -- this will include the `platform` of `openindiana`, `opensolaris`, `nexentacore`, `omnios`, `solaris2`, and `smartos`
+* `bsd_based?` -- this will include the `platform` of `freebsd`, `netbsd`, `openbsd`, and `dragonflybsd`
```
coderanger: I agree with this list but we should have a format reason macOS isn't included.

new/rfcXXX-distro-sugar-helpers.md
lamont-granquist: and that's not just me, we shot that syntax down in this discussion:  https://github.com/chef/chef-rfc/pull/77

plus there's this work with largely replaces it (that i still need to finish very soon):  https://github.com/chef/chef-rfc/pull/238

lamont-granquist:
And I think we probably do eventually pull in the `debian_before_squeeze?` syntax so that we can do flat case statements that reach all the way from the umbrella-level of rpm_based down to matching on platform_family.

The nested syntax is the one that looks like this:

```
case node[:platform_family]
when 
  case node[:platform]
    case node[:platform_version]
    when
    end
  when
  end
end
```

there's also the really ancient hash-syntax of `value_for_node` and `value_for_platform_family` which this is vastly better than.

new/rfcXXX-distro-sugar-helpers.md
lamont-granquist: yeah.  i was trying to get at that with the latter statement.  IDK exactly how to write that right now since i'm coming down with a cold.

lamont-granquist:
Yeah, so the biggest namespace collision will be with chef-sugar itself, which is one thing this is trying to address.  I also plan on dropping it into 13.0 so if it blows anyone up who has done something similarly-named but different
from chef-sugar there's that dot-0 to hide behind.

I didn't address removing the helpers from chef-sugar and future-proofing against deviations between chef-sugar and core chef.  Obviously the intention is that at some point in the future chef-sugar drops support for chef < 13 and then deletes all these helpers.

new/rfcXXX-distro-sugar-helpers.md
lamont-granquist: good or bad?

lamont-granquist:
@coderanger @thommay, et al Open Question: when it comes to the XXX_based? helpers what should the `provides` syntax look like?

Right now it could look like this:

```ruby
provides :whatever do
  fedora_based?
end
```

But that gets the 'precedence-level' incorrect since `fedora_based?` is more general than any platform_family, or platform level matcher.

I would be open to the discussion that we need to eliminate the precedence matching and go back to last-writer-wins since mostly that seems to generate confusion (and at least one regression with the apt_repository chef version vs. cookbook version issue where core chef started binding tighter).  If we decided to do that, then this issue doesn't matter at all.

new/rfcXXX-distro-sugar-helpers.md
thommay: bad! bad shivers!

lamont-granquist:
Updated

`fedora_based?` is now `fedora_derived?` which hopefully in being slightly more vague fits the definition better

there's now a hierarchy of redhattiness:

`rpm_based?`  -- broadest category of stuff that simply uses RPM
`fedora_derived?`  -- that overarching category that I want which is RPM-things-that-aren't-SuSE
`redhat_based?`  -- distros that are strictly compatible with redhat-the-companies-standards (amazon distro fails this test)
`rhel?`/`amazon?`/`fedora?` -- once you get down into true special cases you're here

lamont-granquist:
In thinking about why we keep talking past each other one point to consider is that amazon not having dnf or not having systemd is something that I find completely irrelevant because I just write code like this:

```
if fedora_derived?
  if which("dnf")
    # late-model fedora/rhel stuff
  else
    # amazon and old-model yum stuff
  end
end
```

slicing that up between amazon-vs-redhat doesn't help me write code any and just makes it more complicated.  this is also likely future proof since i really doubt that amazon is going to stick with yum and will eventually be switching to dnf, and will almost certainly copy redhat/fedora when it does.  what you're likely seeing is just that amazon as a company uses the amazon distro and has a migration problem to dnf and systemd which it needs to tackle.  instead of staying on an old RH distro, though, they maintain their own distro and can pull in xen/aws-optimized newness while lagging on stuff that takes migration effort.

meanwhile, you can see that i really don't want to deal with suse in that block.

if you're thinking about the world in terms of managing systems via platform, platform_family and platform_version and not doing any duck-typing of the O/S (no interrogation of `which("dnf")` then you probably don't grok this model at all -- because you'll care very deeply about in which platform_version rhel, fedora and amazon switched to dnf.  i'd prefer to not have to maintain that complex if statement though and let the code sort it out.

thommay:
I would say that, yes, that's a great way of thinking about it, but we've done an awful job in educating users that you can/should think about platforms that way, and we've got zero documentation about doing so.

thommay:
Approved! let's get it done!

thommay:
@chef/rfc-editors let's merge this one please.

adamleff:
Rebased, assigning 87, and merging. Thanks!

PR #258, opened 2017-03-28
Propose Server Enforced Recipes feature

danielsdeleo:
Chef Server will provide an endpoint that MAY serve a Chef recipe file. Chef
Client will attempt to fetch the recipe during run context setup. If
no user action is taken to configure the feature, the endpoint returns 404
and Client behavior will be unaffected. When the feature is enabled, the
endpoint returns the configured recipe file. Chef Client will evaluate and
converge the recipe.

Signed-off-by: Daniel DeLeo <dan@chef.io>

### DIFF
--- a/new/server-enforced-recipe.md
+++ b/new/server-enforced-recipe.md
@@ -1,0 +1,104 @@
+---
+RFC: unassigned
+Title: Server Enforced Recipe
+Author: Daniel DeLeo <dan@chef.io>
+Status: Draft
+Type: Standards Track
+<Tracking:>
+<  - https://github.com/chef/chef/issues/X>
+---
+
+# Server Enforced Recipe
+
+## Description
+
+Chef Server will provide an endpoint that MAY serve a Chef recipe file. Chef
+Client will attempt to fetch the recipe during run context setup. If
+no user action is taken to configure the feature, the endpoint returns 404
+and Client behavior will be unaffected. When the feature is enabled, the
+endpoint returns the configured recipe file. Chef Client will evaluate and
+converge the recipe.
+
+## Rationale
+
+The motivation for this feature is to allow the operator of the Chef Server to
+enforce desired client-side configuration using Chef. Intended use cases include:
+
+* Allow cloud-based vendors to install an agent necessary for correct operation
+  of the service
+* Allow Chef Customer Development Partners to efficiently install experimental
+  client-side software during feature development
+* Allow organizations that operate as internal service providers to enforce
+  standard configurations
+
+Currently the only way to achieve these goals with Chef is to enforce the
+inclusion of recipes on a node's run list, which is not feasible in many cases.
+
+The enforced policy is limited to a single recipe instead of a full cookbook or
+secondary run list for several reasons:
+
+* Cookbooks are Chef Server objects that are organization-scoped and subject to
+  authorization restrictions. Allowing some cookbooks to be global requires
+  additional complexity which is not needed for the intended uses.
+* Cookbooks have versions and dependencies, which have to be solved. There are
+  several ways this could be addressed, but all options introduce unneeded
+  complexity into the solution.
+
+## Motivation
+
+    As a Chef Server Service Provider,
+    I want to enforce a recipe to run on client systems,
+    so that I can ensure client systems are correctly configured.
+
+## Specification
+
+### Enforced Recipe Endpoint
+
+Chef Server shall expose an organization-scoped endpoint for the enforced
+recipe. If the feature has not been configured by the Chef Server
+administrator, the endpoint shall return a 404 response. If the feature is
+enabled by the Chef Server administrator, the endpoint shall return a 200
+response with the recipe content as the response body.
+
+The endpoint shall authenticate the request via Chef Server's usual
+authentication mechanism.
+
+No authorization mechanism is provided. Any user or client with API access to
+any organization on the Chef Server will have read-only access to the enforced
+recipe.
+
+The URL path of the endpoint relative to the organization base path will be
+determined at a future time.
+
+### Chef Server Configuration
+
+The interface for configuring the feature is to be determined.
+
+Though the initial implementation will likely only support a standalone Chef
+Server deployment, the configuration interface will be written such that it can
+be extended to support tiered and HA configurations.
+
+### Chef Run
+
+During the setup phase of the Chef Client run, Chef Client shall make a HTTP
+GET request to the enforced recipe endpoint. If the Chef Server returns a 404
+response, Chef Client will continue the Chef Client run normally. If the Chef
+Server returns a 200 response, Chef Client will store the recipe file in its
+cache directory. Chef Client will then evaluate and converge the recipe using a
+mechanism to be decided.
+
+One possible implementation is to add the recipe to the list of
+`specific_recipes` which is currently populated only via CLI arguments to
+`chef-client --local-mode`. In this case, enforced recipes would be evaluated
+and converged after the primary run list.
+
+## Downstream Impact
+
+No downstream impacts are expected.
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.
--- a/new/server-enforced-recipe.md
+++ b/new/server-enforced-recipe.md
@@ -22,7 +22,8 @@ converge the recipe.
 ## Rationale
 
 The motivation for this feature is to allow the operator of the Chef Server to
-enforce desired client-side configuration using Chef. Intended use cases include:
+enforce limited desired client-side configuration using Chef. Intended use
+cases include:
 
 * Allow cloud-based vendors to install an agent necessary for correct operation
   of the service
@@ -31,8 +32,17 @@ enforce desired client-side configuration using Chef. Intended use cases include
 * Allow organizations that operate as internal service providers to enforce
   standard configurations
 
-Currently the only way to achieve these goals with Chef is to enforce the
-inclusion of recipes on a node's run list, which is not feasible in many cases.
+This feature is targeted at expert level practitioners who are delivering
+isolated configuration changes to the target systems, such as self-contained
+agent software. Users who wish to deliver more comprehensive configuration
+changes should not use this mechanism to deliver those changes directly, but
+could configure an additional Chef Client identity (i.e., node name, client
+key, organization/server url) to deliver those changes via this feature.
+
+As this feature is intended to be used in a manner that is as unobtrusive as
+possible, and in cases where the Chef Server is administrated by a vendor on
+behalf of the user, existing approaches to enforcing client-side configuration
+are not sufficient.
 
 The enforced policy is limited to a single recipe instead of a full cookbook or
 secondary run list for several reasons:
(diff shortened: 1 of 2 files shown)


new/server-enforced-recipe.md
```diff
@@ -0,0 +1,104 @@
...
+be extended to support tiered and HA configurations.
+
+### Chef Run
+
+During the setup phase of the Chef Client run, Chef Client shall make a HTTP
+GET request to the enforced recipe endpoint. If the Chef Server returns a 404
+response, Chef Client will continue the Chef Client run normally. If the Chef
+Server returns a 200 response, Chef Client will store the recipe file in its
+cache directory. Chef Client will then evaluate and converge the recipe using a
+mechanism to be decided.
```
coderanger: This should be decided here :)

new/server-enforced-recipe.md
```diff
@@ -0,0 +1,104 @@
...
+  of the service
+* Allow Chef Customer Development Partners to efficiently install experimental
+  client-side software during feature development
+* Allow organizations that operate as internal service providers to enforce
+  standard configurations
+
+Currently the only way to achieve these goals with Chef is to enforce the
+inclusion of recipes on a node's run list, which is not feasible in many cases.
+
+The enforced policy is limited to a single recipe instead of a full cookbook or
```
coderanger: It is worth noting this inherits a lot of the downsides of chef-apply, can't use `template` or `cookbook_file` resources, difficult to interact with attributes.

new/server-enforced-recipe.md
```diff
@@ -0,0 +1,104 @@
...
+mechanism to be decided.
+
+One possible implementation is to add the recipe to the list of
+`specific_recipes` which is currently populated only via CLI arguments to
+`chef-client --local-mode`. In this case, enforced recipes would be evaluated
+and converged after the primary run list.
+
+## Downstream Impact
+
+No downstream impacts are expected.
```
coderanger: Depending on implementation details, we might end up with resources in the collection which don't belong to any cookbook, which could impact some analysis tools (DataCollector, poise-profiler, etc). Should at least outline how those will be effected.

new/server-enforced-recipe.md
```diff
@@ -0,0 +1,104 @@
...
+    so that I can ensure client systems are correctly configured.
+
+## Specification
+
+### Enforced Recipe Endpoint
+
+Chef Server shall expose an organization-scoped endpoint for the enforced
+recipe. If the feature has not been configured by the Chef Server
+administrator, the endpoint shall return a 404 response. If the feature is
+enabled by the Chef Server administrator, the endpoint shall return a 200
```
coderanger: It would probably be good for future expansion to include some node details, maybe the node name? I could see an argument for platform+platform_version but conditionals like that should probably just be handled in the recipe code itself.

new/server-enforced-recipe.md
danielsdeleo: Lack of attributes is a feature in this case, since that way you're less likely to make attributes that can conflict with the attributes from the cookbooks in the run list.

new/server-enforced-recipe.md
danielsdeleo: We'll probably try using `specific_recipes` first and evaluate that. If it doesn't work out, then we'll try something else. I'm 90% sure `specific_recipes` will work, but would prefer not to be locked in to that for now.

new/server-enforced-recipe.md
danielsdeleo: Per the above, if these work with `chef-client -z path/to/recipe.rb` then they should work with this, assuming we are successful with the `specific_recipes` implementation.

coderanger: I mean by the nature of that usage, those tools usually aren't active whereas they would be here, so that doesn't really help :)

new/server-enforced-recipe.md
coderanger: Yeah, and the lack of templates isn't impossible to deal with, but it will mean this feature needs some specialized docs.

danielsdeleo:
So, about the trade-offs with recipe vs. cookbook. If we use a cookbook, then no matter what, we also have to come up with a solution that lets you create and update that cookbook, and control access to those operations. We _could_ make the "owning" org for the cookbook(s) part of the config, but then there's a kind of footgun situation if you use that org for real work instead of just a container for mandatory cookbooks. On top of that you have some options to work out with deps:

* Don't allow deps at all. In this case you're still kinda in a weird "strict mode" where you're limited in what you can do, which you need to learn about and code around.
* Do a normal non-policyfile dep-solve. Now you'd also have to consider how you can constrain the solution (i.e., environments), plus you're putting more load on the depsolver for simple cases.
* Make people use policyfiles for this feature. I love me some policyfiles, but it's not great if you have a non-policyfile based workflow you're used to.

In all of these cases, you're probably forced into implementing a deeper level of isolation for the other cookbooks within the Chef Client run, because folks will expect things like custom resources, libraries, etc. to work, but those could easily conflict with stuff in the main run list by defining the same classes and so on. That's probably a good thing for some use cases, but it's overkill for the use cases I'm focused on here. This feature existing doesn't preclude adding that if warranted in the future, but I think it makes more sense as part of an intentional multi-policy thing than a globally enforced recipe/cookbook thing.

coderanger:
Going with the first option (no deps, or at least you're not supposed to have deps and use them very very at your own risk) means that the special case bits are all in Chef Server rather than Chef Client though, from the client's PoV there is just an extra thing in the run list which isn't special. That allows for more rapid iteration and puts the onus of complexity in the hands of people that can more rapidly upgrade/iterate. I agree that it's going to be a slog to carve it out as a special case in the server though :-(

danielsdeleo:
I definitely do want to avoid slogging through adding the special case to Chef Server, but I also don't think it's warranted in terms of user benefit.

No matter what, you're gonna have to write the enforced code in some subset of Chef that's safe for this case. Providing just a recipe keeps you away from the most dangerous parts, and experts can deal with the limitations.

In any case Chef Client would need some special casing to handle an "extra" run list item. At the very least you'd need to remove that from the run list before the final `node.save` and be able to deal with the actual run list containing the same cookbook at a different version.

coderanger:
I would solve the different versions issue by saying that's simply not allowed (or at least if it breaks, you get to keep all the pieces), just as with a recipe, cookbooks in a magic prefix/suffix would still be somewhat special, just not _as_ special. If we can keep the client-side special cases to usage conventions rather than new systems, it helps piece this in to the broader ecosystem. For example, testing with ChefSpec and TK would me mostly off the table for this kind of single recipe unless we add special cases in those too. FoodCritic, same issue, and it just goes on. Our whole ecosystem is built around cookbooks :-/

lamont-granquist:
We're going to get users that want template and cookbook_file resources to work with this...

When we were pushing chef-apply as the first touch for learning chef we got that steady dribble of users filing bugs that those resources were broken with chef-apply.

lamont-granquist:
Is there a way we can flip this proposal on its head and lock down run_lists so that only the uber-admin can modify them, but then they could slurp up another object that unter-admins could control?  And it feels like we already have half of that proposal in the currently written desired-current state split proposal.

danielsdeleo:
@lamont-granquist 

re: comment 1, this isn't aimed at folks learning Chef. As I said before, no matter what implementation you pick, you're going to be restricted to some subset of Chef functionality, which is going to have to be documented for use by more advanced Chef users.

re: comment 2, that doesn't cover all of the intended use cases. The goal is that users of that Chef Server can use Chef normally and they just get some extra stuff as unobtrusively as possible.

@coderanger 

> I would solve the different versions issue by saying that's simply not allowed (or at least if it breaks, you get to keep all the pieces)

I think that's the wrong trade-off for these use cases. My goal here is to be as unobtrusive as possible to the end user. I think that once your needs expand beyond what this feature provides, you really need to find a way to collaborate via a Ci system. I want to keep the scope of this limited so we don't provide too much much incentive to avoid that.

martinisoft:
I'm not too keen on this. My concerns stem from this being a separate-but-seemingly-equal workflow especially with how it would work in client mode. I am guessing I can somewhat see the use case here, vaguely, but I am concerned none the less with how this may end being received and potentially abused. As @coderanger also pointed out, the testing side of this is also a concern. How I will I know it works? Do it live™️?

From @danielsdeleo's responses I can see that documentation will help with my first concern a bit (I hope), but the lack of testing feels like this should be documented with lots of red, bright flashing neon warning signs that you are literally playing with fire.

danielsdeleo:
@martinisoft this is definitely not intended to be a mainstream workflow, which is one of the reasons I think it makes sense to just give folks a single recipe instead of a full cookbook or run list snippet. Some folks have requested the ability to have "super admins" insert items into a node's run list at the Chef Server level, but this proposal isn't intended to address that use case directly. The solutions for that use case we have today are 1) create a Ci pipeline to merge and test code from teams with different areas of focus or 2) have separate Chef Client configurations (different client keys, run lists, possibly even different orgs) on a single machine. I believe that either of those solutions are better than having the Chef Server jumble all the code together at the last minute. The proposal as written could be used to set up option (2) by installing additional client configs, which I think would be fine (personal note, I vastly prefer (1) over (2), but sometimes organizations aren't ready for that yet cf. https://discourse.chef.io/t/segregation-of-duties-in-chef/10659/3 ).

As for the testing aspect, the intention is that this mechanism should only be used for installing isolated configuration on the target system, such as an agent service or some cron jobs to run Chef with a different node identity. The recipe file can be kept in a cookbook structure to test with kitchen or whatever and then exported via build process (this is how my team is planning to do it). Since this is not intended as a mainstream workflow option, the additional complexity is fine for the target audience.

coderanger:
To copy over my comments from the meeting, I'm still not hugely enthused about this but since I think I had the wrong user set in mind, I'm not strongly against this. It reads very much like this is the client side extension of an as-yet unannounced Chef Server/Automate feature, and as such the expectation is that any recipe code that will be used with it can be reasonably held to Very Very High Standards as it will either be written by the Chef Software team or directly vetted by them if coming from a Chef Software partner. With those restrictions in mind (give or take that it's hard to judge when it feels like I don't have the complete picture) these trade-offs make more sense to me. If this gets documented at all (and maybe it shouldn't be), we'll need some strongly worded warnings that this is not intended to be a "default base role" or "server-side node classifier" as both are sought after features and this could, in theory, be misused to those ends but it would probably go very poorly for those who try :)

adamhjk:
@coderanger for sure. We absolutely have a specific use-case in mind for this, and while I know its playing with fire, it's certainly something that if you knew exactly what you were getting into, could be useful for people beyond Chef Software. 

I agree with making its documentation very, very clear - that this is a super-advanced feature. :)

danielsdeleo:
Updated the rationale section to clarify the intended use cases and audience. Added a docs section to the spec.

stevendanna:
The proposed implementation of this feature is simple enough that I am 👍 on it as a minimal first step that doesn't prevent any of the more advanced features that one could imagine replacing it.

## On implementing a more robust feature

> I agree that it's going to be a slog to carve it out as a special case in the server though :-(

> I definitely do want to avoid slogging through adding the special case to Chef Server, but I also don't think it's warranted in terms of user benefit.

I don't think think that some of the more advanced versions of this feature would actually be a "slog".  Some thing to remember:

1) We now have a "server-admins" group that we can use to restrict access to global objects.
2) We can attach objects to the "hidden" global org_id without creating any sort of special org that people might be tempted to do.

I agree it would be more work than the proposed feature, but it doesn't seem like one we couldn't do in a clean, not edge-casy way with a medium amount of work.

## On the client side

> Going with the first option (no deps, or at least you're not supposed to have deps and use them very very at your own risk) means that the special case bits are all in Chef Server rather than Chef Client though, from the client's PoV there is just an extra thing in the run list which isn't special.

If we were to go with a feature that allows full cookbooks/run_list_items, I think we could go either way and either provide /extreme/ isolation (basically treat it as its own run) or /no/ isolation.  What would be problematic in my view is something in between.

Should failures in the required_recipe fail the run? I think we should specify that in the RFC.

lamont-granquist:
i'm guessing that if we head into a tardis and take a trip to the future that we'll find that some way of having segmentation by full cookbooks / run_lists without having completely separate runs / identities will likely happen, along with multi-stage chef runs...

thommay:
approved, take it away @chef/rfc-editors

PR #259, opened 2017-04-05
Add On Hold state and remove Rejected

coderanger:
We don't actually use the Rejected status as rejected RFCs never get past Pull Requests. Adding On Hold to more accurately communicate RFCs which are no longer planned for implementation by any of the Chef maintainers but could be in the future.


thommay:
Approved by: coderanger, thom, kallistec, nathenharvey, stevenmurawski, tas50, jhudson, lamont, robbkidd, jtimberman

PR #260, opened 2017-04-05
Mark a whole bunch of RFCs as implemented

coderanger:
I think all of these have been implemented. People more familiar with the actual features should confirm if the RFC actually matches the eventual implementation.


thommay:
Approved besides 67 which is not yet finished server side.

PR #261, opened 2017-04-05
Withdraw RFC 024

coderanger:
With the revival of `chef-solo` based around Zero mode, I would like to suggest we withdraw RFC 24 and instead make chef-client provide a clearer error message when the server URL isn't set.


lamont-granquist:
I think we should just note that the status doesn't necessarily preclude picking this up again and/or rescuing bits and pieces of it that make sense--but nobody is working on it, and its fallen completely off everyone's radars.

PR #263, opened 2017-04-06
server http head

jeremymv2:
New proposal for medium change to Chef-Zero, Chef Server.

### DIFF
--- a/new/server-http-head.md
+++ b/new/server-http-head.md
@@ -29,8 +29,14 @@ The meta-information contained in the HTTP headers in response to a HEAD request
 information sent in response to a GET request.
 
 The HEAD HTTP verb will be added to oc_erchef and chef-zero named object endpoints such that a client http HEAD request for
-an object name will result in a http 200 response code if it exists, 404 if it does not and 401 if the requestor does not
-have read authorization on the object.
+an object name will result in the following standard http response codes (no different than a `GET`):
+
+Code | Reason
+--- | ---
+200 | object exists
+401 | request signature is invalid
+403 | requestor does not have READ on the object
+404 | object does not exist
 
 example named endpoint: /nodes/NAME
 
--- a/new/server-http-head.md
+++ b/new/server-http-head.md
@@ -1,6 +1,6 @@
 ---
 RFC: unassigned
-Title: API HTTP HEAD support
+Title: Server HTTP HEAD
 Author: Jeremy Miller <jmiller@chef.io>
 Status: Draft
 Type: <Standards Track, Informational, Process>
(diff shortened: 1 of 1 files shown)


jeremymv2:
Related to: https://github.com/chef/chef-server/pull/1218

new/server-http-head.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Specification
+
+The HEAD method shall be identical to GET except that the server must not return a message-body in the response.
+
+The meta-information contained in the HTTP headers in response to a HEAD request should be identical to the
+information sent in response to a GET request.
+
+The HEAD HTTP verb will be added to oc_erchef and chef-zero named object endpoints such that a client http HEAD request for
+an object name will result in a http 200 response code if it exists, 404 if it does not and 401 if the requestor does not
```
stevendanna: 401 should be the returned code if the request signature is invalid, 403 if the requestor does not have READ on the object.

stevendanna:
Overall, I'm 👍  on this.  It is long overdue.

Do we want to specify which endpoints will now support HEAD or are we ok saying all endpoints that support GET will also get a HEAD request?  For some endpoints (`/nodes/`, `/clients/`), the HEAD request makes a bit less sense (I suppose hose should just always return 200?).

For those who might be thinking, "Isn't it relatively efficient to do a `GET /nodes` and then check for the name?".  The answer is, "yes, below a certain node count and in some situations", but for other situations (like trying to bulk upload data into a chef-server) the act of iterating over the list can actually start to dominate your runtime.

ctdk:
As far as which endpoints should support HEAD: I've been looking around, and there's an awful lot of endpoints that use GET that a HEAD response is not in any way meaningful. It's not just endpoints like `/nodes` or `/clients`, but also ones like `/environments/NAME/roles/NAME`.

A meaningless HEAD for some of these GET requests that short circuits most of the logic aside from needed perm checks is fine, but some decision ought to be made.

I've found a few areas where a HEAD request isn't as useless as it first seemed, however. I was looking at `/reports`, thinking that it wouldn't have much, but then realized that it would be possible to do things like make a HEAD request to a URL covering a date range to see if there were any reports at all in that range. The somewhat tenuous usefulness of that by itself is that said HEAD could use a simpler SQL query to ascertain the reports' existence.

That might work for some of these other endpoints, although I'm not quite sure what the HTTP status code should be for all of them. For something like `/environments/NAME/roles/NAME`, StatusOK might be good for indicating that there are indeed roles with an env run list for that environment, but I'm not sure that StatusNotFound is correct if the environment and role exist, but that role doesn't have an env run list for that environment. On the other hand, I can't think of anything better yet.

(As an aside, I think this is a wonderful idea.)

jeremymv2:
With respect to @ctdk's comments, I think it reasonable to establish a boiler-plate pattern for adding the HEAD method that can be re-used across most if not all endpoints where it makes sense, then cover the endpoints as time permits or where appreciative gains can be made (see existing PR above.)

lamont-granquist:
kind of feels like it should be added everywhere (that implements GET) no matter if it makes sense or not.

i could see where if it becomes difficult to add it to some endpoint where it wasn't appropriate that it could be skipped, but i find it hard to imagine how it becomes difficult for some endpoint...

thommay:
Approved by popular vote; take it away @chef/rfc-editors

adamleff:
RFC 090 assigned and accepted. Thanks!

PR #264, opened 2017-04-06
fix table formatting

moutons:
GFM's formatting required particular spacing at the end of the line in order to render the table & an additional "field" at the end of the divider line was mucking things up.


thommay:
Thanks for the PR, this was applied.

PR #269, opened 2017-06-15
Anonymous user behaviour data collection

thommay:
In order to provide greater insight into how our Chef users interact with Chef tools, this RFC introduces a framework for defining the policies for the anonymous collection (sharing), storage, and deletion of product usage data that can be applied to any Chef project.

### DIFF
--- a/new/telemetry.md
+++ b/new/telemetry.md
@@ -1,0 +1,201 @@
+---
+RFC: unassigned
+Title: Anonymous user behaviour data collection
+Author: Thom May <thom@chef.io>, Charles Johnson <charles@chef.io>
+Status: Draft
+Type: Informational
+---
+
+# Anonymous user behaviour data collection
+
+In order to provide greater insight into how our Chef users interact with Chef
+tools, this RFC introduces a framework for defining the policies for the
+anonymous collection (sharing), storage, and deletion of product usage data
+that can be applied to any Chef project.
+
+## Scope
+
+This RFC only covers Chef's open source ecosystem. It does not apply to proprietary,
+commercially available tools made by Chef Software.
+
+## Motivation
+
+    As a UX designer at Chef,
+    I want to understand how people use Chef tools,
+    so that I can improve their experience.
+
+    As a product owner at Chef,
+    I want to learn which Chef tools and commands are most used,
+    so I can prioritise development.
+
+## Specification
+
+This RFC defines the core of product usage data sharing. It does not define how
+individual apps will implement product usage data sharing; any app that
+wishes to do so MUST create a follow up RFC that details individual behaviour,
+and the types of data that will be shared with Chef Software.
+
+### Sharing by default
+
+In order to gather meaningful data, Chef tools will share anonymized usage data
+by default. Experience (with Habitat and Automate) has shown that disabling
+sharing by default provides sample sets that are too small (and potentially
+biased) to draw meaningful conclusions.  The Chef tools must provide clear
+information so that users always know how to easily change their data sharing
+preference. Once set, any data sharing preference saved on an individual host
+will be honored by all Chef tools run on that host.
+
+On the first run of a telemetry enabled tool, the tool MUST prompt the
+user if a decision about whether to participate has not been made. That
+decision will be recorded for future use. If the tool is unable to prompt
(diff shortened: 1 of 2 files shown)


thommay:
action items from 15/06 community meeting:
  * add back the language about public sharing
  * explicitly disallow chef-client
  * collect `Chef and other tool versions`
  * allow tools to detail what they're collecting
  * change "and it is expected that each app would create a follow up RFC that details individual behaviour" to be "each app MUST create a follow up RFC ..."

new/telemetry.md
```diff
@@ -0,0 +1,136 @@
...
+preference.
+
+### Privacy and Data Retention
+
+To ensure that sensitive data is not collected, Chef tools should never
+collect parameters passed to command-line options, and tools will be able to
+filter more aggressively if required.
+
+For example, knife may report that a `-P` option was invoked, but will never
+collect the password parameter that was passed to that option.
```
lamont-granquist: maybe they should also only collect data on successful commands?  unsuccessful commands are more likely to have accidentally typed passwords on them and such (`knife bootstrap -Passw0rd1`)

thommay: well, we want to know about people running unsuccessful commands, but maybe in that case we should just collect which command was tried.

davidski:
I've been a casual Chef user and a booster of the ecosystem for several years. I understand the desire to get telemetry on use of the products and the frustration with the poor uptake on opt-in mechanisms. At the same time, moving to an opt-out stance is a third rail issue for me, my organizations, and the clients I serve. I strongly urge Chef Co. to reconsider this RFC. If adopted, I will be forced to take a serious look at removing Chef products from my tool chain as I find such behavior completely unacceptable.

It's good to know that Habitat already has this turned on by default. I had been intending to delve into that product but will now strike that off my list.

Again, the desire for this information is certainly understood and I ascribe no ill intentions to anyone. Simply not a behavior I find acceptable in any of the software I use or support.

davidski:
Closing the loop on a Slack conversation, my objection is to any software that collects user behavior without my explicit, *opt-in* consent. As an open-source user and contributor, collecting user behavior without consent changes the implicit contract in a way with which I am not comfortable. Arguments that other software or orgs are doing similar items are not germane. I care about Chef and what Chef does here.

It's clear that my opinion is not a vocal one. I suspect, but have no data to prove, that there are many other casual users of the chef ecosystem of products, that also object to such a change. Consider this a voice of those users. Again, adopting this change will cause me to give pause to recommending Chef to clients large and small and certainly reduce the efforts I put into supporting the Chef ecosystem in general.

smith:
> It's good to know that Habitat already has this turned on by default. I had been intending to delve into that product but will now strike that off my list.

Data collection is _not_ enabled by default in Habitat. See https://github.com/habitat-sh/habitat/blob/master/components/hab/src/analytics.rs

stormerider:
This seems solid to me, personally.

thommay:
Further discussion on opting in/out:
 if there's no sentinel flag and we have a terminal, ask the question and be hard yes or hard no. if we don't have a terminal, leave it nil, which right now means "disabled", but in future and given warning and notification may change to "enabled"

metaxis:
Telemetry is "nice to have", and not in any way required to release and iterate on quality software.  Worse, as an opt-out default, it ends up creating an anti-demographic cohort, giving the most weight to users who care the least about their privacy and how their data is used.  

Explicit, opt-in consent is a moral requirement in my opinion.  Anything less is explicitly anti-user.  I would hate for Chef to become Yet Another piece of hostile software designed to run at the developer's convenience rather than the users.

lamont-granquist:
I could have used it here to understand the impact of removing the `berks test *` set of commands:

https://github.com/berkshelf/berkshelf/pull/1702

Could also use it here, to give information on how many people are still using `berks cookbook/berks init` instead of chefdk and which version of berks they're on:

https://github.com/berkshelf/berkshelf/pull/1729

Just two examples from the past couple months

The whole Chef-13 release was also a pile of deprecations.

ldesiqueira:
I am certainly in favor of explicit opt-in strategy for data collection especially on open source software.

thommay:
approved @chef/rfc-editors

adamleff:
Accepting, assigning RFC 094

PR #271, opened 2017-07-26
Stats endpoint for Chef Server

jaym:


coderanger:
Is this a replacement for `/_status`? Seems like some level of overlap going on, which is okay, but maybe just touch on that :)

jaym:
Not a replacement. This endpoint is to provide metrics. Similar to what https://github.com/facebook/chef-utils/blob/master/chef-server-stats/chef-server-stats provides but without mucking with the internals of chef server.

new/server-stats-endpoint.md
```diff
@@ -0,0 +1,115 @@
...
+    I want to understand the metrics I'm given,
+    so that I can easily monitor the health of my Chef Server.
+
+
+## Specification
+
+This RFC adds a series of `/_stats/$service` endpoints to Chef Server. Each service
+colocated on the box will be provided an endpoint, and with respond to GET requests.
+The response format will be a text format which is specified by 
+[Prometheus 0.0.4](https://prometheus.io/docs/instrumenting/exposition_formats/#text-format-details).
```
mivok: We might want to address the fact that prometheus also has a binary/protobuf format and explain why we do or don't want to support it.

coderanger:
Maybe at least combine the namespaces, `/_status/metrics` or something?

new/server-stats-endpoint.md
```diff
@@ -0,0 +1,115 @@
...
+This gives us the ability to use a few metrics types specified below along with the ability to give
+a description for any metrics which are provided.
+
+### Supported Metric Primitives
+The endpoints can provide metrics of following types, as defined by [Prometheus](https://prometheus.io/docs/concepts/metric_types/):
+- Counter - Monotonic cumulative metric. For example, number of requests served.
+- Gauge - Numeric measurement at a given point in time. For example, number of inflight requests.
+- Histogram - Distributes values into buckets. For example, response times.
```
mivok: This is something I don't think is provided right now by the statsd stuff we use, but it would be awesome to have buckets saying things like "m requests took 0-10ms, n requests took 0-20ms and so on", as it makes it possible to get percentile estimates with multiple frontend servers.

new/server-stats-endpoint.md
```diff
@@ -0,0 +1,115 @@
...
+The format is already defined and provides metric types that are known to work with at least
+one monitoring system. This format also gives us the ability to describe each metric presented,
+so the end user would know what it means. The format itself is human readable and easy to parse,
+making it easy to both get started with and write tooling around.
+
+Using this format directly also could allow us to use metric exporters already written for other co-located
+services such as PostgreSQL, RabbitMQ, and Solr.
+
+#### Why Not?
+This would introduce an inconsistency in that the current API uses JSON and this one route would not. We
```
mivok: This is a good point - if you want to be extra fancy, if the requester sends an Accept: application/json header, then you could return json too. I'm not sure of the implications of supporting multiple formats though and if it's worth the trouble.

new/server-stats-endpoint.md
```diff
@@ -0,0 +1,115 @@
...
+could just as well create a JSON specification with information such as type, description, metric name, etc.
+
+
+### Services
+At a minimum, the Chef Server frontend service would need to provide stats. For example, `bookshelf`, `oc-id`,
+`oc_bifrost`, and `erchef` should provide metrics as those are always colocated. Since these run together, these could also be grouped into a single metrics endpoint(`_stats/chef-server`) to reduce the number of calls needed to gather the metrics.
+
+If Chef Server is configured to provide services such as RabbitMQ, Solr, Elasticsearch, and PostgreSQL, an
+endpoint should be provided for each service.
```
mivok: It might be worth including manage/reporting addons here too.

new/server-stats-endpoint.md
```diff
@@ -0,0 +1,115 @@
...
+
+#### Service List
+Since different implementations or configurations of Chef Server could have different dependencies, 
+`GET /_stats` will provide a list of services.
+
+Example response:
+
+```
+{
```
mivok: This feels odd given that all child endpoints don't give out json as things are currently specified. I don't have a better thought on this though, unless the single _stats endpoint gave you all metrics at once.

new/server-stats-endpoint.md
```diff
@@ -0,0 +1,115 @@
...
+available through statsd or Folsom will be available from the stats endpoint. This means that other
+methods should be deprecated and removed in the next major version bump of Chef Server. Those who
+wish to push to Graphite may still do so by polling. Doing the same for statsd may no longer make sense
+as the stats endpoint will have already aggregated the metric data. That being said, those who wish to
+still use statsd can get the data from the request logs.
+
+## Unanswered Questions
+
+- What about chef-backend?
+- Do we need or should we have authn/authz?
```
mivok: We don't currently have authn/authz for the _status endpoint. It's also not _too_ hard to limit access in nginx if it needs locking down.

new/server-stats-endpoint.md
jaym: I believe we will get these if we do this change because we have to do the aggregation. So good to keep in mind

new/server-stats-endpoint.md
jaym: We could do this. From an implementation perspective, it makes life a bit harder as we cannot directly use the exporters others have written

new/server-stats-endpoint.md
jaym: this is what i wanted somebody to tell me :)

jaym:
@coderanger I think `_status` is different than metrics. That being said, if people don't want to blow up the number of top level endpoints, putting it under `_status` would make sense.

new/server-stats-endpoint.md
jaym: yea, the inconsistency feels bad

new/server-stats-endpoint.md
jaym: we can do both :)

new/server-stats-endpoint.md
stevendanna: A single _stats endpoint that returns everything would probably mean the client-side code could be substantially more simple.

new/server-stats-endpoint.md
stevendanna: I don't see this as a major blocker and I think the benefit of using a pre-defined format outweighs the inconsistency.

new/server-stats-endpoint.md
```diff
@@ -0,0 +1,114 @@
...
+This introduces yet another way to get metrics for the Chef Server. All metrics that were previously
+available through statsd or Folsom will be available from the stats endpoint. This means that other
+methods should be deprecated and removed in the next major version bump of Chef Server. Those who
+wish to push to Graphite may still do so by polling. Doing the same for statsd may no longer make sense
+as the stats endpoint will have already aggregated the metric data. That being said, those who wish to
+still use statsd can get the data from the request logs.
+
+## Unanswered Questions
+
+- What about chef-backend?
```
stevendanna: I don't think we need this in the RFC, but we can have a similar stats endpoint in leaderl and add it as an upstream to check in chef-server when `use_chef_backend true` is set.

new/server-stats-endpoint.md
stevendanna: My guess is we would implement this in erchef so we could do the authn/authz there if we wanted to.  I agree that we should keep this open to all for now. Similar to _status.

new/server-stats-endpoint.md
```diff
@@ -0,0 +1,114 @@
...
+- Do we need or should we have authn/authz?
+- Versioning
```
stevendanna: If this is part of the API we could re-use the API versioning headers.  However, I think part of the contract should be that clients should not expect any particular stats or services since we want to make it easy to add statistics as the system changes.

new/server-stats-endpoint.md
jaym: This might be best, and would also make returning a JSON format easier if we decided we wanted to.

new/server-stats-endpoint.md
```diff
@@ -0,0 +1,114 @@
...
+
+    As a Chef Server operator,
+    I want to understand the metrics I'm given,
+    so that I can easily monitor the health of my Chef Server.
+
+
+## Specification
+
+This RFC adds a series of `/_stats/$service` endpoints to Chef Server. Each service
+colocated on the box will be provided an endpoint, and with respond to GET requests.
```
jeremymv2: s/and with respond/and will respond/ ?

jaym: sorry, fixed

new/server-stats-endpoint.md
err0: I've seen it done where there is a catalog endpoint that lists all of the other endpoints available.

new/server-stats-endpoint.md
```diff
@@ -0,0 +1,108 @@
...
+one monitoring system. This format also gives us the ability to describe each metric presented,
+so the end user would know what it means. The format itself is human readable and easy to parse,
+making it easy to both get started with and write tooling around.
+
+Using this format directly also could allow us to use metric exporters already written for other co-located
+services such as PostgreSQL, RabbitMQ, and Solr.
+
+#### Why Not?
+This would introduce an inconsistency in that the current API uses JSON and this one route would not. We
+could just as well create a JSON specification with information such as type, description, metric name, etc.
```
ronlipke: +1 for a JSON specification.  There's already a lot of tooling around grabbing metrics as straight JSON.

jaym:
We should come together tomorrow and settle on a format, I don't think it's worth doing both JSON and prometheus. But that shouldn't take long, it sounded like most people wanted json

jaym:
Add a format parameter

new/server-stats-endpoint.md
jaymzh: @andy-dufour  and I spent a lot of time working on server metrics. When and where the right ones are used is important.

This stuff shuold be based on the folsom-stats and realtime-pool-stats work Andrew did here: https://github.com/facebook/chef-utils/tree/master/chef-server-stats

That data has been critical in us monitoring our servers

jaym: @jaymzh The specifics about which metrics was left as an implementation detail. However, to ease your concerns, one of the goals was to replace the need for that script. The idea is we will start with that script and add metrics as needed.

andy-dufour:
I'm loving the way this is shaping up. Reading over the RFC I didn't see if stats could be broken down by subsystem - an idea I was toying with would be to have /_stats be a rollup of all metrics, and /_stats/<service> be a specific services stats. 

It could also be taken care of via post-processing by whatever process is consuming the stats endpoint but allowing a consumer to only get the data they want may be a better experience.

jaym:
@andy-dufour we originally had stats broken down by service, however people wanted to one endpoint to give all of them, so I took that section out. If you still think it's valuable, we might be able to work it in: https://github.com/chef/chef-rfc/blob/c968b55b56e458f37c017ee88c8b4284e448ea0d/new/server-stats-endpoint.md

new/server-stats-endpoint.md
jaymzh: Yeah the examples were just missing a lot, but that's fine as long as we plan to get it in there and JSON is an output option I'm 👍

thommay:

adamleff:
RFC 093 assigned, thanks @jaym!

PR #272, opened 2017-08-02
Transfer openstack user on Supermarket to Chef OpenStack team

scassiba:


lamont-granquist:
should probably figure out what e-mail registered that user, might just be a chef employee or something.

scassiba:
that would be best case scenario, and something we could work with. it would also alleviate an immediate sadness in the OpenStack world.

thommay:
@scassiba is going to contact the user and try direct negotiation.

scassiba:
I have sent an initial ping to the user in an attempt to negotiate directly as of 2017-08-03. There is no evidence of a given name or even searchable handle that I can find, that does not lead directly back to me or my team. Therefore, I have little reason to reasonably expect that a human is still monitoring the inbox after three years.

To me, an ample amount of time to await reply seems to be either coinciding with the PTL elections, or the Pike release itself per the [https://releases.openstack.org/pike/schedule.html](Pike release schedule), deferring to @thommay on what actually defines ample. If a human is actually on the other end and still aware of OpenStack, these milestones matter.

For the sake of transparency, I have already reached out to the Foundation due to the matter at hand, and it is on their radar.

lamont-granquist:
So since Hosted Chef accounts are also supermarket accounts (and vice versa, its the same API) the problem is that someone may have used the 'openstack' user to manage their own openstack infrastructure via Hosted Chef (potentially without knowing at all that accounts on Hosted Chef show up as accounts on supermarket).  We'd need to get our ops folks to validate that there's no data associated with that account in Hosted Chef in order to consider this.  The fact that the user has done nothing on supermarket at all makes me somewhat suspicious that this is an account that someone used to sign up to Hosted Chef with.

scassiba:
I understand that Supermarket and Hosted Chef are tied to the same ID, being a Hosted Chef user. This request was not made flippantly, in a vacuum. My team was externally prompted to explore alternate distribution methods, particularly Supermarket.

Back to business, I have sent a follow-up as of 2017-08-08, with no response to the initial inquiry. Please see if Hosted Chef even has any data associated with the account in the meantime.

It would be downright delightful to close the bug that a Chef employee initiated on our bug tracker with optimal results.

robbkidd:
Research hath occurred.

Suggested update to the **Downstream Impact** section:

> Hosted Chef and Supermarket admins have determined the `openstack` user:
>
> + on Supermarket
>     + owns 0 cookbooks
>     + collaborates on 0 cookbooks
>     + follows 0 cookbooks
> + on Hosted Chef
>     + is a member of only one organization
>     + is the only member of the one organization
>     + is associated with 0 clients and nodes (no running infrastructure)
>     + has uploaded > 1 cookbook

Given the above, I suggest the following replace the contents of the **Deprecation Process for this RFC** section:
 
> 1. Hosted Chef and Supermarket administrators record that the `openstack` user is not associated with any running infrastructure or published cookbooks.
> 1. Hosted Chef administrators remove the `openstack` user from its existing organization.
> 1. Hosted Chef administrators change the email address on the `openstack` user to an address of the [Chef OpenStack team](https://github.com/openstack/openstack-chef-specs/blob/master/MAINTAINERS.md)'s choosing.
> 1. Supermarket administrators announce this change on [the Chef mailing list](https://discourse.chef.io/c/chef).
> 1. The Chef OpenStack team receives access to Hosted Chef `openstack` account via password reset over email.
> 1. The Chef OpenStack team creates a new key pair for the `openstack` user and logs in to Supermarket to update account information and key cached there.
> 1. The Chef OpenStack team begins publishing their OpenStack cookbooks to Supermarket.
>

coderanger:
I'm real uncomfortable with moving this forward in light of the fact that the user has actually interacted with Hosted Chef, even if there are no nodes at this time. That moves this from a name-squat to an eminent domain request, and I would prefer if we had even pro forma trademark complaint on file before we do anything, just to set a precedent.

scassiba:
I have updated the deprecation process from the feedback above. I have also reached out to the Foundation to weigh in with their position, as the actual trademark holder.

fifieldt:
Hi,

Tom from the OpenStack Foundation here (you may verify with an email to tom@openstack.org, communitymgr@openstack.org or info@openstack.org).

I confirm that Samuel Cassiba is the project leader of the OpenStack Chef project, under OpenStack Foundation governance (https://www.openstack.org/legal/).

Samuel has discussed the needs of the project, and the potential to use the Chef Supermarket to distribute the cookbooks produced within the project. Doing so seems in the best interest of both OpenStack and Chef users.

We support the request to reclaim the "openstack" name in the Chef Supermarket and for the project to use it under our community guidelines and other applicable agreements. We would would be appreciative of any assistance you could provide and are here to answer any questions needed.

Regards,

Tom

robbkidd:
Noting here for the record that the single organization with which the Hosted Chef `openstack` user is affiliated is _not_ named `openstack`.

If this transfer of the username were to be approved and the human currently behind the account were to finally respond, we would have the ability to help that user create a new account with a new username and grant them access to the existing Hosted Chef organization.

coderanger:
Action item from the meeting was that the Chef team will talk to Legal and work out any specific requirements on that side of things to help ground my random mumblings in reality :)

ricksentam:
So what's that really mean ? I still don't understand what sort of code they or we should use

thommay:
As Robb says, this has now been done. My feelings are that there's not much requirement to assign this an RFC number or merge it to master, since this is not community affecting going forwards, so I'm going to close this PR. Thanks to everyone for getting this done, and thanks for your patience and your contribution to Chef, @scassiba

PR #276, opened 2017-10-04
Provide many levels of debug logs

thommay:
In general, the logging one needs to debug a failing chef run is very different to the logging one needs to debug a problem in chef's code.
We should make it so they are two separate things.

### DIFF
--- a/new/debug_levels.md
+++ b/new/debug_levels.md
@@ -1,0 +1,53 @@
+---
+RFC: unassigned
+Title: Improved Log Levels
+Author: Thom May <thom@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Improved Log Levels
+
+Presently, the only way to debug a failing chef run is to enable debug
+logging, which is intended more for the maintainers of Chef than for a
+systems engineer. We would like to introduce several levels of enhanced
+logging, enabling more appropriate output for various roles.
+
+## Motivation
+
+    As a Chef maintainer,
+    I want to know exactly what chef is doing,
+    so that I can maintain and develop chef.
+
+    As a cookbook author,
+    I want to ensure that my cookbook is interacting with the system as
+    I expect,
+    so that I can write cookbooks efficiently.
+
+    As a systems engineer,
+    I want to see the output of running commands,
+    so that I can debug chef failures.
+
+## Specification
+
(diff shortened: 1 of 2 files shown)


lamont-granquist:
we should do either this or the numbered levels or something :+1:

coderanger:
As a possible counter-offer, a thing I've been tossing around in my head for a while is something like `--debug-cookbook foo` which would set the log level to debug only for that one cookbook (i.e. inside attrs and recipes for that cookbook or custom resources defined in that cookbook). Might be less confusing since it doesn't need any new levels (though the two are not in direct competition, we could do both).

jaymzh:
A static set of named debug levels will run you out of room quickly. I would instead do **either**:

levels 1-9 with some guidance on what those levels mean (like you did, but call those 1, 3, 5, and 7 or something)

**or**:

Named logtypes, that can be toggled independent of each other, ala:

```
--debug ohaiplugins,chefinternal
--debug cookbooks,ohaiinternal,chefinternal
--debug chefinternal
```

thommay:
We think the correct approach is actually structured logging, and then building a UX that allows you to compose output based on that structure

rmiesen:
I agree with @jaymzh's idea of adding named log types. On a similar vein, it would be nice if there were a `--no-debug` option. For example
```
--no-debug ohaiplugins,chefinternal
--no-debug cookbooks,ohaiinternal,chefinternal
--no-debug chefinternal
```
This would allow end-users to squelch all debugging output for things they selectively don't care about (example: OHAI output when you're trying to debug an SSH connection issue with chef provisioning).

jaymzh:
It would be easy to make Chef::Log handle this gracefully. Anything where only one arg is passed in is assumed to be:

```
Chef::Log.info('chefinternal', "message here")
```

danielsdeleo:
It would be great if we could use structured logging to unify the formatter approach and the logger approach for output. It would be a fair bit of work, but would be great to clear up confusion between those things and make information that's only available in one or the other available in a unified thing.

lamont-granquist:
this looks like a good direction 👍

tas50:
👍 This will be really useful to cookbook authors

jonlives:

adamleff:
RFC 96 assigned, merging now - thanks!

PR #277, opened 2017-10-04
required ohai plugins

thommay:
Enable a cookbook to specify the Ohai plugins it expects to have available and working.

### DIFF
--- a/new/required_ohai_plugins.md
+++ b/new/required_ohai_plugins.md
@@ -30,16 +30,19 @@ This RFC proposes the addition of two fields to cookbook metadata:
  * `required_ohai_plugins`: A list of ohai plugins that are critical to
    the operation of the cookbook. If any of the listed plugins are not
    available or fail to run, the chef client run is aborted.
- * `suggested_ohai_plugins`: A list of ohai plugins that enhance the
-   cookbook. Absence or failure of these plugins do not cause the chef
-   client run to fail.
 
-These fields are additive to any site wide configuration of critical
+This field is additive to any site wide configuration of critical
 ohai plugins.
 
-The chef client run would be altered to use a minimal set of ohai
-plugins by default, supplemented by the set of plugins requested by
-cookbooks and site wide configuration.
+Currently, the chef client runs ohai twice - once before and once after
+cookbook syncing, so that ohai plugins that are loaded from cookbooks
+get properly used. This proposal would tweak that process by ensuring
+the first run uses only the minimal ohai plugin set. Any plugins
+in the minimal set should be treated as critical.
+
+The second run would then be subject to the full set of ohai
+configuration options, which would include the full set of critical
+plugins.
 
 ## Copyright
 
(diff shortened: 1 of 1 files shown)


lamont-granquist:
i think "suggested" is likely the wrong term?  if they're declared here, then the chef-client should run them all.   they're not just suggested, they're more like dependencies.

phil already shipped his feature as "critical" plugins as well, and we should likely stick to that name (or change it real quick).

jaymzh:
I'm wary of using the same name here, lest people think that defining `critical_ohai_plugins` would add that config to their Ohai config, which it won't...

coderanger:
How would this interact with custom plugins added from recipe code (or the as-yet unrealized `ohai/` segment)? Overall very :+1:, even though it would probably be a ~2019 goal to switch to a minimal plugin set :)

thommay:
@jaymzh i think it _would_ add that to the configuration for ohai in chef…

@coderanger `ohai_plugin` stuff is harder, but i imagine if you're writing cookbooks and using this you'd just use the ohai segment. That would just work, since by the time we go to load all the plugins, we'd have our configuration in place.

minimal by default is harder, although what this enables is that we do a first pass with the equivalent of `--minimal-ohai` and then run a second pass when we've got our cookbooks. Right now 13 will run two passes (although not with a minimal first pass) if you're using "segments" for ohai plugins.

thommay:
Let's remove `suggested_ohai_plugins` and think about a consistent name for required/critical plugins

new/required_ohai_plugins.md
```diff
@@ -0,0 +1,52 @@
...
+    I want to only run the ohai plugins that I need,
+    so that I send less data to the chef server.
+
+    As a cookbook author,
+    I want to ensure that my cookbook has the system data it needs,
+    so that it correctly configures the system.
+
+## Specification
+
+This RFC proposes the addition of two fields to cookbook metadata:
```
tas50: two fields?

danielsdeleo:
It may be more ergonomic for cookbook developers to specify the _paths_ they want, such as `filesystem` or whatever, which is exactly why ohai has the `provides` stuff, like https://github.com/chef/ohai/blob/6c58ca9df17971e71dca6c4e4c73406cfad86bcd/lib/ohai/plugins/linux/filesystem.rb#L22

jonlives:
Feedback from the dev meeting is that daniel's comment above still needs addressing before we move this one forward.

danielsdeleo:
Also, if we were to go the route of ohai data being accessed from a different API than just the "giant set of merged attributes", then we could just load plugins on demand. The new API would have to build up a path such that we could ask ohai for the full path, like what happens when you run `ohai filesystem2/by_mountpoint` on the command line.

Usage could look like:

```
sys.get("filesystem2", "by_mountpoint")
```

That would imply a bigger set of changes, though.

thommay:
@tas50 i think this would still be a valuable thing to do, but clearly not in this repo ;)

PR #278, opened 2017-10-04
target mode

thommay:
Allow the easy management of switches and other appliances with Chef

### DIFF
--- a/new/target_mode.md
+++ b/new/target_mode.md
@@ -1,0 +1,53 @@
+---
+RFC: unassigned
+Title: Chef Target Mode
+Author: Thom May <thom@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Chef Target Mode
+
+When deploying applications or systems, we frequently need to configure
+appliances where Chef cannot run natively. To allow us to do so, we need
+to run chef on a dedicated management node and interact with the
+appliance remotely, via an API.
+
+## Motivation
+
+    As a site reliability engineer,
+    I want to configure routers, switches and load balancers,
+    so that my application functions correctly.
+
+## Specification
+
+The concept of a node in chef is a 1:1 mapping with a client, enabling a
+single node to have several identities if needed. We can leverage this
+facility to create a target mode, allowing a management node to run chef
+for many appliances.
+Target mode is enabled by several new configuration options:
+ * `target`: The details for connectivity to the target
+   system. The result of this configuration option should be a
+   class instance that cookbooks can send messages to. This
+   enables target mode, disabling ohai entirely.
+ * `platform`: The platform that we're connecting to. Since we don't
+   have ohai, this can be used to make decisions in a cookbook.
+
+Since target mode does not affect the local node, we can run many target
+mode chef client runs in parallel. Thus, a chef client run should only
+block other chef client runs that affect the same target, rather than
+being system wide.
+
+For cookbook authors, target mode implies having a connection to the
+node that enables RPC, via an API or using ssh or similar. Once
+configured, this connection would be made available using a DSL method,
+enabling cookbooks to build resources that interact with the target.
+
+Target nodes would be assigned policies or `run_lists` as normal.
+
+## Copyright
+
+This work is in the public domain. In jurisdictions that do not allow for this,
+this work is available under CC0. To the extent possible under law, the person
+who associated CC0 with this work has waived all copyright and related or
+neighboring rights to this work.


new/target_mode.md
```diff
@@ -0,0 +1,53 @@
...
+   system. The result of this configuration option should be a
+   class instance that cookbooks can send messages to. This
+   enables target mode, disabling ohai entirely.
+ * `platform`: The platform that we're connecting to. Since we don't
+   have ohai, this can be used to make decisions in a cookbook.
+
+Since target mode does not affect the local node, we can run many target
+mode chef client runs in parallel. Thus, a chef client run should only
+block other chef client runs that affect the same target, rather than
+being system wide.
```
lamont-granquist: This should probably just call out directly that we need to modify the run lock around the chef-client to have per-target filenames so that they do not conflict with each other.  That is an implementation detail, but its the kind of obvious work that we'd need to do in order to make this more pleasant to work with.

new/target_mode.md
```diff
@@ -0,0 +1,53 @@
...
+
+The concept of a node in chef is a 1:1 mapping with a client, enabling a
+single node to have several identities if needed. We can leverage this
+facility to create a target mode, allowing a management node to run chef
+for many appliances.
+Target mode is enabled by several new configuration options:
+ * `target`: The details for connectivity to the target
+   system. The result of this configuration option should be a
+   class instance that cookbooks can send messages to. This
+   enables target mode, disabling ohai entirely.
```
lamont-granquist: It seems like a cookbook could already do most of this, and could expose its own API for setting a target.  The only thing that it really needs here is the disabling of default ohai plugins by default.  We could just build a boolean to do that easily.

new/target_mode.md
```diff
@@ -0,0 +1,53 @@
...
+ * `platform`: The platform that we're connecting to. Since we don't
+   have ohai, this can be used to make decisions in a cookbook.
```
lamont-granquist: Can we do this simply by shipping an ohai plugin in a cookbook which runs first and populates the platform + friends?  I don't know what happens if we don't have a platform in between when ohai runs and early in the cookbook_loader, but if we just debug that so that it doesn't explode, then we could let this be the responsibility of the cookbook author.  All the ProviderResolver resolution is lazy, we don't need a platform at class loading time with resources or providers in order to setup the NodeMap, so I don't think there's anything preventing us from letting the platform get set later.  It might not work without some debugging but could probably be treated as bugs and just fixed.

lamont-granquist:
I guess I could sum up all three comments I just made as wanting to keep this RFC as thin as is possible so that we're not architecting new concepts that we don't actually need.  And really it feels like we should write a POC cookbook first and see what works and what doesn't and then fix the things which fail.  There's a few details which I'm sure we need to fix (the lockfiles are probably not delightful to mess with), but that feels like some fairly lightweight UX tweaking could fix that, and I'm very worried that we assume that this is totally novel and requires bigger architecture in an RFC than what we need and results in design-by-committee / bikeshedding, when a POC with some iteration cycles could result in a better product.

thommay:
Yeah I agree in general - I think this RFC is very much "we should do _something_ in this arena", please have at it.

coderanger:
I agree with Lamont here. The use case is absolutely valid, but until there are enough implementations of this to not be a rounding error I don't know what machinery we really need in core. We already have a minimal-ohai flag, having a disable-ohai seems fine too (though you can already do that with an explicit ohai config I think?).

There is probably a kernel of this use case with fewer moving pieces in core that we can extract, leaving most of the data handling decisions up to future cookbooks.

new/target_mode.md
jaymzh: At Facebook we do this with a wrapper that takes a target (in our case phone IDs), and uses separate lockfiles, and uses that to pass in a separate set of attributes.json files to Chef to bootstrap itself.

new/target_mode.md
zfjagann: I think a slightly better way to do this would be to define a lower-level transport which includes ways to mutate and inspect some remote target system which can be used for both ohai plugins and implementing resources/providers.

I'm thinking of something similar to [chef/train](https://github.com/chef/chef-rfc/pull/278/), which implements POSIX file operations, and allows you to run shell commands on the target system, and wholy defines how these calls get implemented on the target, be they simply to locally run the operations, or to forward them over ssh, or adb, or whatever.

I think that'd allow you to reuse a lot of existing providers with fairly minimal modifications, and I don't think it'd be too difficult to extend the Resource/Provider architecture to allow writing providers for specific target types, like Android phones over ADB, or a switch over ssh for instance.

new/target_mode.md
zfjagann: I think in practice this is difficult to enforce. If you do this as a cookbook with a resource/provider API, it'll be difficult to prevent users from doing other operations that affect the local system. On linux you can enforce this with namespaces, but that's very platform (and version) specific.

zfjagann: I don't really think the lock file is the issue here. That's also really easy to work-around by changing the `lockfile` in `client.rb`.

I think it's pretty difficult to enforce that one chef run will only touch the remote system and not the local system.
In the case of android phones, it makes a lot of sense to control the installed version of the android toolchain with the code that interacts with the phones, since they usually need to change in lock-step. But once you start managing files on the local system, you have to have some way to deconflict the operating-system level chef run and the chef run managing the phones.

new/target_mode.md
coderanger: @zfjagann I think you might be reading a bit too far in to this. We are very much not planning for this to do transparent transport abstraction stuff. This is purely around running the client, what's actually in the recipe code is entirely up to those cookbooks.

thommay:
there's an inflight InSpec RFC that is related: https://github.com/chef/inspec/issues/1661

PR #279, opened 2017-10-06
Add draft RFC to support a global option to disable ignore_failure

jonlives:
This PR submits an RFC to support a config option to globally (in client config) disable ignore_failure.

Signed-off-by: Jon Cowie <jonlives@gmail.com>

### DIFF
--- a/new/ignore-failure-disable.md
+++ b/new/ignore-failure-disable.md
@@ -1,0 +1,36 @@
+---
+RFC: unassigned
+Title: Option to Disable Ignore Failure
+Author: Jon Cowie <jcowie@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Option to disable Ignore Failure
+
+This RFC proposes the addition of a configuration setting to ```client.rb``` which allows the administrator of a Chef organisation to disable usage of the ```allow_failure``` resource parameter on all cookbooks executed by that node.
+
+## Motivation
+
+    As a Chef  administrator,
+    I want the option to disable ignore_failure
+    so that users cannot write recipes which fail invisibly when they should instead be fixed
+
+## Specification
+
+This RFC proposes the addition of two new configuration parameters to be supported in client.rb / solo.rb.
+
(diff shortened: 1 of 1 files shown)


coderanger:
This seems like a niche enough thing to handle via existing tools:

```ruby
Chef.event_handler do
  on :resource_action_start do |resource, *args|
    if resource.ignore_failure
      Chef::Log.info("whatever")
      resource.ignore_failure(false)
    end
  end
end
```

lamont-granquist:
No?

It can be used in community cookbooks to suppress, e.g. 404's in some cases where they're actually expected:

https://github.com/lamont-cookbooks/sk_ruby/blob/3f6085d310edf330f5a3be003b744dd24d197e2a/providers/default.rb#L61

That means that as a community cookbook author you can't ever use ignore_failure deliberately, and you'll wind up getting into arguments with users who have disabled ignore_failure by policy, which will then wind up making this approach unusable for community cookbooks, which will result in community cookbook authors arguing with core chef developers around adding additional use cases (clearly that use case there we could work around that with some kind of a `ignored_failures :404` flag added to remote_file, but that opens up a whack-a-mole problem.

The right approach here is to have a foodcritic rule which fails on ignore_failure being used.  And if users are constantly overriding that with exceptions where its not appropriate, then there's a training/"human resources" issue going on, and adding a flag to disable ignore_failure globally isn't going to fix that.

And before someone gets super fixated on the fact that growing more properties off of remote_file fixes the example I just gave "better" and is therefore more "correct", that misses the point.   The `ignore_failure` workaround has worked fine there in the past, and I haven't needed to fix it, and if you allow some users to globally disable and change the fundamental behavior of ignore_failure, you effectively eliminate it from ever being used in any community cookbook for any circumstance.  That means we can never use it to work around core chef bugs or backcompat issues, etc.  And we absolutely need the flexibility to do terrible things on occasion.

jonlives:
I'm going to close this PR for now. Noah's fix should work well enough for the specific case this was proposed for, and your comments are indeed valid Lamont.

PR #280, opened 2017-10-06
RFC-075 proposed replacement

jonlives:
This PR proposes an RFC to replace the existing RFC-075 (Multiple Policy Files and Teams)

### DIFF
--- a/new/multiple-policy-revision.md
+++ b/new/multiple-policy-revision.md
@@ -214,7 +214,7 @@ Finally, after we've run ```chef update Myapp.rb```, here is the resulting merge
   "revision_id": "xyz12345xyz12345xyz12345xyz12345xyz12345xyz12345",
   "name": "myapp",
   "run_list": [
-    "recipe[base::default,cookbook::default]"
+    "recipe[base::default]", "recipe[mycookbook::default]"
   ],
   "cookbook_locks": {
     "base": {
--- a/new/multiple-policy-revision.md
+++ b/new/multiple-policy-revision.md
@@ -119,8 +119,7 @@ In the event of any conflicts occuring, this RFC makes it explicitly clear that 
 There are several foreseeable potential conflicts I will highlight here explicitly where we will not attempt to resolve the conflict, but will rather return an error (please note, this list is illustrative and not exclusive):
 
 * Conflicting dependant cookbook versions (ie one Policyfile depends on version 1.2.4 and another on 1.2.5)
-* Conflicting values for Policyfile attributes
-* In the case of Policyfile attributes which have an array as their value, the arrays will not be merged but rather one array will overwrite the other (whichever was included last will win, essentially)
+* Conflicting values for Policyfile attributes - if the same attribute is set in two places, no merge of their values will be attempted.
 * An include loop where a policy includes a second policy which in turn includes the first policy.
 
 Essentially, we will only merge elements from Policyfiles where we can be sure that we are not overriding something specified in another Policyfile (ie we can safely combine two sets of cookbook locks if the dependencies do not clash). My approach to this RFC is that you should never have to be surprised by the effect of including another Policy, and it should not be able to change the behavior of a Policyfile which includes it. Please see the "Example" section following this one for an example of a policy including another, and the resulting merged policy.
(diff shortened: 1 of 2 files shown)


new/multiple-policy-revision.md
```diff
@@ -0,0 +1,110 @@
...
+```ruby
+include_policy "base", git: "github.com/myorg/policies.git", path: "foo/bar/baz.lock.json"
+```
+
+The ```include_policy``` directive will support three sources for policies: 
+
+* ```git```, with the following parameter being the URL to a Git repository and a path parameter specifying the location of the file within the repository
+* ```server```, with the following paramater being set to a Chef server URL, a policy_name paramter and a policy_revision_id parameter.
+* ```local```, with the following parameter being a path to a file on disk.
+
```
Afterglow: http support for people building policy files to and storing them in artifactory or s3 (for example) would be good.

new/multiple-policy-revision.md
```diff
@@ -0,0 +1,110 @@
...
+* Data contained in the parent policyfile's .rb file
+* Computed cookbook locks from the lock files for all policies specified with an ```include_policy``` directive.
+
+The single fused lockfile produced by the above would then be uploaded to the Chef server as normal, and would function as a single Policy.
+
+
+## Problems
+
+The principal potential issue with this approach is that because we are pulling cookbook locks from a number of included cookbooks, it is necessary for all included policy Lockfiles to be re-scanned upon regeneration of the parent Policyfile, so that the full set of cookbook locks and constraints can be examined and merged. Although using the ```git``` source to include Policies will mean that the same commit SHA is used every time to rescan the included Lockfile and the ```server``` source will always ensure the same revision ID is used, when the ```local``` source is used we have to use whatever the current Lockfile present on disk is. This means that we cannot guarantee it has not changed since the last time it was scanned.
+
```
Afterglow: I think there are two major issues with trying to squash multiple policy files into a single ```parent```

1) Attribute overrides - If two included policy files set the same attribute how is this handled? It almost has to be an error otherwise policy files can override each others attributes and the results would be almost chaos. Debugging this situation as a developer of an included policy file which worked everywhere except where it was included would be difficult at best. 

2) In the context where all the included Policyfiles are squashed and then compiled as a single Policyfile does this mean that the computed locks for all common dependencies have to be identical? Otherwise the behaviour would be the contrary to the intention of Policyfiles where dependencies could change depending on where the Policyfile was included.

coderanger:
This is a solid start for sure. I don't think it really addresses the same use case as RFC75 but regardless it's a :100: feature. The include sources looks good, but I think we need to further specify the semantics of an include. Like how does the run list get merged, which attributes win (presumably the local policy), how array attributes get merged (because this will never not bite us >_<), and how we handle conflicts.

jonlives:
Thanks for the feedback so far folks - I'll be incorporating it asap :)

jonlives:
@coderanger @Afterglow I've added the above commit to clarify behaviour in the event of merges / conflicts - let me know what you think!

new/multiple-policy-revision.md
jonlives: I think that's an entirely reasonable thing - I'm probably not going to target it for the initial RFC, but it's absolutely something we could look at adding in.

Afterglow:
The clarifications looks good to me and address my questions over merging policies.

jonlives:
Per developer meeting on 26th October, more words are needed around what happens in the event of duplicate runlist entries, array attribute merges, and a real world example of a merge would be good too.

jonlives:
Added a few more comments to address the above - example policy & merge will come tomorrow or early next week.

jonlives:
For consistency and making sure that what you specify in your policyfile is what you get, after internal discussion I've changed the RFC to stipulate that two attributes with the same name will never be merged and always produce an error.

coderanger:
I think you mean to say that leaf values won't merge/overwrite but non-leaf hashes will? I'm not sure I love that plan but it's easier to start with that and add in role-style merging later if there is strong demand :)

jonlives:
Right @coderanger that is indeed the intent (wording clarified to reflect this)...our feeling internally was the same, we can add in role-style merging further down the line.

jonlives:
Approved by a maintainer vote, take it away @chef/rfc-editors

adamleff:
Assigning 97 to Policyfile Includes. Merging now, thanks!

PR #282, opened 2017-10-10
First pass on resource deprecation RFC

coderanger:
I think we should remove `deploy`, `http_request`, and `erl_call` from Chef. Discuss.

### DIFF
--- a/new/deprecate-deploy-http-erlcall.md
+++ b/new/deprecate-deploy-http-erlcall.md
@@ -1,0 +1,59 @@
+---
+RFC: unassigned
+Title: Deprecate deploy, http_request, and erl_call
+Author: Noah Kantrowitz <noah@coderanger.net>
+Status: Draft
+Type: Standards Track
+---
+
+# Deprecate `deploy`, `http_request`, and `erl_call`
+
+Three resources in Chef core that cause a lot of user confusion and issues are
+the `deploy`, `http_request`, and `erl_call` resources.
+
+With the `deploy` resource, the confusion is mostly inherent in the very large
+number of moving pieces in the resource. The Capistrano-style deployment logic
+makes sense when a deployment tool is expecting to co-exist with manual management,
+but this is not usually the case with Chef recipes. Most users would be better
+served by a simpler `git` resource for deployment.
+
+`http_resource` is a common stumbling block as people see it and expect they
+can use it to get the value of the HTTP response in their recipe code. Given the
+use case for making an HTTP request without caring about the response is incredibly
+niche, this is probably no longer a good fit for Chef core.
+
+`erl_call` is more debatable, it doesn't tend to cause a lot of user confusion
+and the maintenance burden isn't really a problem, but if we're already doing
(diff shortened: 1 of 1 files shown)


tas50:
Http_request is pretty heavily used on supermarket

coderanger:
@tas50 What's the use case?

lamont-granquist:
yeah mostly people want feature requests and extensions to http_request, not removing it...

stevendanna:
👍 on `erl_call`.  It is kinda a silly resource.

I think that if we are going to move these out into cookbooks, it would probably good to have a plan for the life of the cookbook.  Moving them to a cookbook means we still maintain them, so it seems to me that the main benefit is preventing users from seeing confusing/ill-advised resources in core chef.  That goal can probably also be solved in documentation without any deprecation.

That said, I *am* in favor of deprecating the deploy resource on some timeline.  My worry there is doing that before we have other primitives in core chef to facilitate replacing it.

iennae:
I thought step one was to write up the model of deprecation. Then based on that we could align which resources to deprecate. 

The erl_call resource based on the definition we discussed at Chef Summit sounds like the svn resource we discussed as an example. If there is not user confusion or maintenance burden and it helps even a small number of people what is the reason to deprecate?

From the notes that Seth wrote on the board I have:

- primitive versus compound resources
- broken for N days
- disconnect from user expectation
- % of supported platforms
- footgun-ness for new user

coderanger:
@iennae That is certainly an option, but I don't think it would be a productive one given that all three of these are IMO candidates for removal from core for different reasons that have little to do with each other.

iennae:
That's kind of my point. I'm not saying every resource has to have the same reason, but it'd be nice to be clear and specific about the reasoning with something to map back to for the rest of the resources that might be brought up.

jaymzh:
I'm with @iennae - We should start with some basic guidelines for when, why and how we remove things from core. It does not need to be an exhaustive list, but when I mentioned in the meeting an RFC, it was for getting some documentation around "valid" reasons we might want to do that as well as documenting the process for doing so...

coderanger:
@iennae I don't think there is any such clear reason here though, all three are IMO candidates for totally unrelated reasons. That said, if someone wants to make a process I'm happy to go that route too, I just don't really want to wait on it.

thommay:
I'm happy with the removal of `erl_call` since it's bonkers and probably no-one has used it since couch died. I'm happy with the removal of `deploy` to a cookbook, since it's complex, hugely tied to very specific rails/cap workflows and more or less unmaintained.
I think we can't remove `http_request`, but we can make it clearer how it's used, what it's good for, and how to make http requests and use the data from them in Chef.

thommay:
@iennae please can you write up an RFC that captures that deprecation model?

iennae:
@thommay I could take a stab at a first draft of it :)

thommay:

adamleff:
RFC 98 assigned, merged to master in f45eb1d6

PR #283, opened 2017-10-19
Sleep to deprecation notices

jjasghar:
Inspired from a blog post about using sleep in compiling C++ programs
this RFC will add sleep to our deprecation notices. This is a way for us
to gently nudge our users to pay attention to the notices so they aren't
effected in major releases.

Signed-off-by: JJ Asghar <jj@chef.io>

### DIFF
--- a/new/rfc-sleep-deprecation.md
+++ b/new/rfc-sleep-deprecation.md
@@ -1,0 +1,56 @@
+---
+RFC: unassigned
+Title: Adding sleep to deprecation notices
+Author: JJ Asghar <jj@chef.io>
+Status: Draft
+Type: Process
+---
+
+# Adding sleep to deprecation notices
+
+This RFC is to add a progressively longer sleep time per `Deprecated` notice.
+Most users of software don't pay attention to deprecation notices, and can
+get bit by the removing resources in later versions of software. If we
+something that can get progressively more and more noticeable we can gently
+nudge our downstream users to move away from these resources. This has been
+inspired from a [blog post][blogpost] about how a C++ library was
+deprecated with a more and more progressive sleep for compile time.
+
+## Motivation
+
+    As a developer of Chef or downstream user of Chef,
+    I want to make sure I am using the supported resources,
+    so that I have the best tool for my job.
+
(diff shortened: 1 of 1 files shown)


lamont-granquist:
different problem space though and on production servers where you want chef-client runs to be finishing in 60 seconds or less this can impact your ability to deploy changes.

i'd also find that behavior of a compiler completely infuriating and the first thing i'd do would be to add whatever environment variable to my .zshrc that would make that go away.

coderanger:
Yeah, I’m -1 on this as well, even limited to just Test Kitchen. Because of the complex web of community cookbooks, many warnings are not actionable by the person seeing them. This is the same reason that warnings as errors isn’t on by default in Kitchen.

danielsdeleo:
granular warnings as errors would be cool, if you could extend policyfile/berks to let you know which cookbooks are likely owned by the user.

coderanger:
Part of the structlog context work on my list is the ability to do per cookbook debug logging, hopefully we could do a similar thing for warnings too.

adamleff:
I think this is a nice approach for things like the ChefDK and InSpec CLI where there's interaction with a human, but I'm :-1: on this being used with CCRs.

jjasghar:
Turns out it wasn't an article I found on hacker news but this tweet:

https://twitter.com/johnregehr/status/920691341738123264

It get's the point across too though :)

lamont-granquist:
After seeing the tweet, I'm a very solid 'no' 👎  on this in general.  Either with CCRs or with the CLI.  In most cases it is just going to be an abusive UX experience to the user.

There's a very specific case where the problem is something like that deprecated library where all the user needs to do is delete linking against it where I can see it.

But this was not being added to `-Wall` notices from the compiler or anything like that.

thommay:
Yeah, I think this not something we want to do as part of a CCR. Building better tooling in Automate and so on to surface deprecations is much less user intrusive.

Thanks for the RFC, @jjasghar

jonlives:
👎 on this one too

tas50:
👎 as well

PR #285, opened 2017-11-14
add a credentials config file

thommay:
Based heavily on http://docs.aws.amazon.com/cli/latest/userguide/cli-multiple-profiles.html .

Would appreciate feedback from @spheromak @fujin @coderanger etc.

### DIFF
--- a/new/authentication_config_file.md
+++ b/new/authentication_config_file.md
@@ -40,7 +40,7 @@ client_key = "barney_rubble.pem"
 chef_server_url = "https://api.chef.io/organisations/bedrock"
 
 [dev]
-node_name = "admin"
+client_name = "admin"
 client_key = "admin.pem"
 validator_key = "test-validator.pem"
 chef_server_url = "https://api.chef-server.dev/organizations/test"
@@ -49,12 +49,32 @@ chef_server_url = "https://api.chef-server.dev/organizations/test"
 File paths, such as `client_key` or `validator_key`, will be relative to
 `~/.chef` unless absolute.
 
+Clients can be identified with either `node_name` or `client_name`, with
+`client_name` being preferred.
+
+Keys can be specified inline using TOML multiline strings.
+
 The profile is selected using the `CHEF_PROFILE` environment variable, which
-client libraries MUST support. Optionally, tools can also provide a
-`--profile` option, which would override the environment variable.
+client libraries which implement credential handling MUST support.
+Optionally, tools can also provide a `--profile` option, which would
+override the environment variable.
+
+To allow the user to specify statefully which profile they wish to
+use, we will introduce another new file, `~/.chef/context`, which would
+contain only the name of the profile to be used.
 
 It is expected that the credentials file will be parsed first, allowing
 the user's `knife.rb`/`config.rb` to continue to function as previously.
+
+To select the profile to use, libraries should follow the following
+order:
+
+ 1. If an argument is passed with a command line flag, use that.
+ 2. Otherwise, check for the `CHEF_PROFILE` environment variable, and
+    use the value of that.
+ 3. Otherwise, check for the context file, and read the profile name
+    from that.
+ 4. Otherwise, select the default profile.
 
 ## Downstream Impact
 
--- a/new/authentication_config_file.md
+++ b/new/authentication_config_file.md
@@ -1,0 +1,73 @@
+---
+RFC: unassigned
+Title: Authentication Configuration
+Author: Thom May <thom@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Authentication Configuration
+
+Currently, the Chef community maintains many tools to aid in the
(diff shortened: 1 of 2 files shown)


thommay:
also @benvaljean

new/authentication_config_file.md
```diff
@@ -0,0 +1,73 @@
...
+client_key = "admin.pem"
+validator_key = "test-validator.pem"
+chef_server_url = "https://api.chef-server.dev/organizations/test"
+```
+
+File paths, such as `client_key` or `validator_key`, will be relative to
+`~/.chef` unless absolute.
+
+The profile is selected using the `CHEF_PROFILE` environment variable, which
+client libraries MUST support. Optionally, tools can also provide a
```
coderanger: Most Chef client libraries don't currently support loading Chef configs at all (because need to either parse or run Ruby) so it will be a while before this `MUST` will really be correct. Not saying I disagree since TOML is definitely easier to parse, just setting up for reality.

new/authentication_config_file.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Specification
+
+A new file, `~/.chef/credentials`, will be supported by the chef client
+libraries, and other implementations. It is a TOML file, containing one
+or more sections, each associated with a profile.
+
+```
+[default]
+node_name = "barney"
```
coderanger: If we're making something new, can we support `client_name` as an alias for this and use that as primary in the docs?

new/authentication_config_file.md
```diff
@@ -0,0 +1,73 @@
...
+client_key = "barney_rubble.pem"
+chef_server_url = "https://api.chef.io/organisations/bedrock"
+
+[dev]
+node_name = "admin"
+client_key = "admin.pem"
+validator_key = "test-validator.pem"
```
coderanger: Can we list fully which keys will be allowed here? Or say "anything goes but not all Chef options will be expressible in TOML", though even in a latter case we could then document which options will always be expressible in here.

new/authentication_config_file.md
```diff
@@ -0,0 +1,73 @@
...
+It is expected that the credentials file will be parsed first, allowing
+the user's `knife.rb`/`config.rb` to continue to function as previously.
+
+## Downstream Impact
+
+Any chef API client library (such as `pychef`, `go-chef`,
+`rs-chef-api`), as well as `chef-config`, should be updated
+to support this file. 
+
+Manage should be updated to produce a credentials file when generating a
```
coderanger: If you wanted to make my new-user-confusion heart even happier, could we allow for in-line keys a la:

```toml
client_key = """
-----BEGIN RSA PRIVATE KEY-----
MIIE...
-----END RSA PRIVATE KEY-----
"""
```

Would make new user set up a lot easier if we can hand them a single file instead of config+user key.

new/authentication_config_file.md
thommay: I think anything goes is probably _correct_, but there may well be value in just documenting the above to make it easier, and then say that any complex config should go in `config.rb`.

new/authentication_config_file.md
thommay: 👍

new/authentication_config_file.md
thommay: "client libraries that implement this"?

thommay:
I'll take a peak at `kubectl config set-context staging `

new/authentication_config_file.md
coderanger: :+1: that sounds better. Make it clear we prefer this feature be all or nothing :)

thommay:
Looking further at `kubectl config set-context`, I think this would be achievable by:

Adding a new `chef` sub command, `chef credentials set dev`, which would create `~/.chef/context`, containing only `dev`.
Libraries would then read that file if it exists, then parse `CHEF_PROFILE`, then handle direct user config from a command line option or whatever.

coderanger:
:+1: That sounds about right (though I would flip the order, CLI arg, env var, ambient context) :)

thommay:
per slack we're on the same page on this one.

new/authentication_config_file.md
thommay: added!

benvaljean:
This seems like a great idea +1

thommay:
We should ensure that it's easy to set a profile in a knife.rb so that users can continue to use site or directory specific chef config.

thommay:
@svanharmelen @spheromak some eyes on this from a golang client perspective would be awesome.

thommay:
https://github.com/chef/rs-chef-api/blob/tm/builder_pattern/chef_api/src/credentials.rs is an implementation of this...

svanharmelen:
@thommay I see no issues with this RFC. Looks just fine to me 👍

thommay:
Approved at the 2017-12-07 meeting, take it away @chef/rfc-editors

PR #286, opened 2017-11-15
Adding myself as an advocate to the mailing list

eeyun:
Since we are creating a Habitat topic in the chef mailing list (discourse) I would like to function as an advocate in the mailing list!

Signed-off-by: eeyun <ihenry@chef.io>


nathenharvey:
👍

PR #287, opened 2017-11-16
Official deprecation freeze policy

coderanger:


coderanger:
Merging under my powers as RFC-y person, since this is just documenting existing, informal plans.

PR #289, opened 2017-11-27
disable session and passwd

thommay:
We get so many reports of failed chef run reports, or huge use of disk space as a result of these two plugins. Let's turn them off by default.

An alternative approach might be to add them to the `automatic_attribute_blacklist`, but I think the UX of understanding that and how users would go about managing the blacklist would be much worse.

### DIFF
--- a/new/disable_passwd_and_sessions_plugins.md
+++ b/new/disable_passwd_and_sessions_plugins.md
@@ -1,0 +1,37 @@
+---
+RFC: unassigned
+Title: Disable Passwd and Sessions Ohai plugins
+Author: Thom May <thom@chef.io>
+Status: Draft
+Type: Informational
+---
+
+# Disable the Passwd and Sessions Ohai plugins
+
+We get a large number of user reports that these two plugins can cause a
+huge amount of unwanted storage use, especially in large enterprises
+with nodes hooked up to an AD tree. We should disable them by default
+in Chef 14, and provide appropriate notifications.
+
+## Motivation
+
+    As an Operations Engineer,
+    I want to run my Chef Server efficiently,
+    so that I can concentrate on providing value to my employers.
+
+## Specification
+
+By default, make the `passwd` and `sessions` plugins disabled by
+default.
+
+## Downstream Impact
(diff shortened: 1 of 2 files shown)


tas50:
Just a note we'll need to update the systemd_unit and service/systemd providers to not use node["etc"] to determine the user's UID then.

thommay:
We should ship both `passwd` and `passwd_min`, disable them both, and let our users come up with a plan. We should also hook in to the attributes code https://github.com/chef/chef/blob/7c5f4dee428b0d369e67fc31b4753ff8039d2c10/lib/chef/node.rb#L194 to give reasonable warnings

lamont-granquist:
we should also have a foodcritic rule that looks for `node['etc']` and pushes people towards etc.getpwnam or whatever

lamont-granquist:
we'll need a way to enable_plugins to go along with disable_plugins (i think since we're enabled-by-default we never made a switch for that), and along with that api it might be useful to have some way of specifying `:all` so that the behavior is backwards-compatible.  it seems like it'd be cheap to implement and the major impact would just be on the site in questions own node size (i.e. it doesn't have any weird knock-on effect to cookbook authoring)

new/disable_passwd_and_sessions_plugins.md
```diff
@@ -0,0 +1,37 @@
...
+## Motivation
+
+    As an Operations Engineer,
+    I want to run my Chef Server efficiently,
+    so that I can concentrate on providing value to my employers.
+
+## Specification
+
+By default, make the `passwd` and `sessions` plugins disabled by
+default.
```
jeremymv2: nit: use of the word `default` twice in the sentence is redundant.

sean-horn:
If you go the password_min direction, this was tested and working earlier for Chef 12 and its embedded Ohai 6+ https://github.com/stevendanna/ohai-plugins/pull/2

No problem philosophically with automatically disabling both plugins. We've had a rash of session plugin tickets lately.

thommay:

PR #291, opened 2017-11-29
Add Deprecation Warnings Within Custom Resources RFC

tas50:

### DIFF
--- a/new/deprecations-in-resources.md
+++ b/new/deprecations-in-resources.md
@@ -36,7 +36,7 @@ We will introduce 3 new methods in custom resources in order to implement this R
 
 ### deprecated method for resources
 
-This new method will let authors communicate to consumers that a resource is going away in the future. Right now we rely on readme or changelog entries, which are not a very effective way to communicate to consumers. This method will accept true, false (default), or a string. If set to true any consumer of the resource will see a canned deprecation warning. If a string, then that string becomes the warning message.
+This new method will let authors communicate to consumers that a resource is going away in the future. Right now we rely on readme or changelog entries, which are not a very effective way to communicate to consumers. This method will accept a string which becomes the warning message.
 
 #### Example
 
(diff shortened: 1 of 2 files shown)


new/deprecations-in-resources.md
```diff
@@ -0,0 +1,89 @@
...
+RFC: unassigned
+Title: Deprecation Warnings Within Custom Resources
+Author: Tim Smith <tsmith@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Deprecation Warnings Within Custom Resources
+
+If Chef 12 we introduced deprecation warnings within the chef-client. This allowed us to start communicating future breaking changes to users. The warnings and integration within Test Kitchen have become a powerful tool allowing users to future proof their cookbooks and ease chef-client migration projects.
```
nathenharvey: In Chef 12...

tas50: I looks like it was 12.16 when we improved the way they worked based on our deprecations RFC

majormoses:
Very nice, will be super helpful!

new/deprecations-in-resources.md
nathenharvey: Sorry, I meant to replace "If" with "In"

new/deprecations-in-resources.md
```diff
@@ -0,0 +1,89 @@
...
+RFC: unassigned
+Title: Deprecation Warnings Within Custom Resources
+Author: Tim Smith <tsmith@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Deprecation Warnings Within Custom Resources
+
+In Chef 12 we introduced deprecation warnings within the chef-client. This allowed us to start communicating future breaking changes to users. The warnings and integration within Test Kitchen have become a powerful tool allowing users to future proof their cookbooks and ease chef-client migration projects.
```
iennae: %s/start communicating/communicate/

new/deprecations-in-resources.md
```diff
@@ -0,0 +1,89 @@
...
+         deprecated: 'Turns out destroying everything was a bad idea. This property will be removed in the 3.0 release of this cookbook in April 2018 and will throw an error if set at that time.'
+```
+
+### deprecated_property_alias
+
+Currently if a resource author decides to change the name of a property they have two options: 
+	- Use alias_method which silently routes old properties to the new names
+	- Define both properties in the resource and include functionality to set the new value using the old value while warning the user.
+
+alias_method doesn't alert cookbook consumers to the change and writing your own code to perform deprecation warnings is cumbersome and rarely done. A new deprecated_property_alias would behave similar to a alias_method, but throw deprecation warnings while providing backwards compatibility. It would accept and optional String value that would be used in place of a generic deprecation message.
```
thommay: "an optional String"

thommay:
both `deprecated` methods should just take String, per the 2017-12-07 meeting.

thommay:

PR #293, opened 2017-11-29
Resource Lifecycle RFC

iennae:


new/resource_lifecycle.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Process for adopting resources
+
+* Identify resources for adoption.
+* Create an RFC following the [Chef RFC](https://github.com/chef/chef-rfc) process.
+* Announce on Chef mailing list and Chef Community Slack.
+* Any proposed changes to the interfaces must be implemented in the cookbook prior to adoption.
+* Any bugs discovered must be repaired in the cookbook and released prior to adoption.
+* Resource is added to core chef and documentation is updated.
+* Add deprecated cookbook warnings for conflicting cookbooks.
```
lamont-granquist: If the API of the resource needs to change in core compared to the cookbook then the resource should be renamed in core so as not to conflict with the cookbook.

an example is the `chocolatey_package` resource, which needed to be renamed anyway to comply with the naming standard that we have in core, and it needed to adopt the API shape of the rest of the package resources.  the cookbook version of the resource was designed around chocolatey's CLI API, while the core version of the resource was designed around core chef's package resource API, which was fundamentally different design concerns.  since they were named differently, they did not conflict.

iennae: At what point would we change the resource name? You mean upon adoption or post adoption?

tas50: If the naming structure doesn't match in the cookbook we'd change it as we bring it into core. That wouldn't impact the cookbook.

iennae:
Updated language. @lamont-granquist would appreciate your thoughts.

new/resource_lifecycle.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Specification
+### Criteria for adopting resources
+
+Factors that influence and inform the decision to adopt a resource include:
+
+* widespread use of a resource based on analysis of publicly available supermarket cookbooks,
+* well tested and supported cookbook resources,
+* percentage of supported platforms,
+* impact of implementing as a primative that can be built upon.
```
thommay: primitive

nellshamrell:
👍

thommay:

PR #295, opened 2017-12-20
Add Self Documenting Resources RFC

tas50:


coderanger:
If we aren't going to go with YARD-style markup, we should say why :)

coderanger:
Also would like to see some API examples for how to use this data either inside Chef itself or other tools, and possibly a proposed JSON schema for how we might store it.

thommay:
We approved this @chef/rfc-editors

thommay:
thanks @adamleff

PR #296, opened 2018-01-24
expose automatic chef_environment attr

thommay:
We should've done this ages ago, but let's get it done.
This is also important should we wish to integrate hoist
directly in to chef.

### DIFF
--- a/new/chef_env_attribute.md
+++ b/new/chef_env_attribute.md
@@ -1,16 +1,16 @@
 ---
 RFC: unassigned
-Title: `chef_environment` as an attribute
+Title: Expose more settings as attributes
 Author: Thom May <thom@chef.io>
 Status: Draft
 Type: Standards Track
 ---
 
-# Exposing the chef\_environment attribute
+# Exposing some core settings as attributes
 
-Historically we've not exposed the Chef environment as an automatic
-attribute, although tools such as poise-hoist have changed that. Policy
-groups make the whole situation a little more murky.
+There are a number of settings, such as the node name and the environment, that we've historically not exposed as automatic attributes.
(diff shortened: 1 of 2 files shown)


coderanger:
Very :+1: and we should probably make `node['name']` Do The Right Thing too. The only slight complication is people miiiiight misunderstand what `node.set['chef_enviroment'] = 'prod'` does (i.e. it does nothing) but I think that is 100% worth it given how many people trip up on this.

lamont-granquist:
normal level though right?  run_list and tags are normal.

coderanger:
I would imagine automatic, thats how we set things like `node['recipes']` and `node['policy_revision']` currently.

new/chef_env_attribute.md
```diff
@@ -0,0 +1,41 @@
...
+Author: Thom May <thom@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Exposing some core settings as attributes
+
+There are a number of settings, such as the node name and the environment, that we've historically not exposed as automatic attributes.
+The environment attribute is a bit more complicated in the world of
+policy groups, and this RFC clarifies that situation too
```
tas50: Can we give an example of how you'd reference this as an attribute vs. the "old" way.

thommay:
I was definitely thinking automatic.

lamont-granquist:
okay so `cookbooks`, `recipes`,  `expanded_run_list` and `roles` are all automatic...

new/chef_env_attribute.md
thommay: Not sure that the audience should need that level of explanation?

thommay:
@chef/rfc-editors we done approved this!

PR #297, opened 2018-02-01
Hoist policyfile attributes

thommay:
When using policyfiles, it is convenient to be able to 
access attributes based on the policy_group of the node.

### DIFF
--- a/new/hoist.md
+++ b/new/hoist.md
@@ -1,0 +1,58 @@
+---
+RFC: unassigned
+Title: Integrate hoisting of policy\_group attributes
+Author: Thom May <thom@chef.io>
+Status: Draft
+Type: Standards Track
+---
+
+# Integrate Hoisting in core Chef
+
+Most users of Policies rely on "hoisting" to provide group specific
+attributes. This approach was formalised in the `poise-hoist` extension,
+which we'd like to move in to the Chef Client and formalise.
+
+To hoist an attribute, the user would provide a default attribute
+structure in their Policyfile similar to:
+```
+default['staging']['myapp']['title'] = "My Staging App"
+default['production']['myapp']['title'] = "My App"
+```
+and would access the node attribute in their cookbook as:
+```
+node['myapp']['title']
+```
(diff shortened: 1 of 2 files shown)


new/hoist.md
```diff
@@ -0,0 +1,58 @@
...
+    I want to leverage policy groups to describe my estate,
+    so that I can isolate environments correctly.
+
+## Specification
+
+Hoisting will be enabled by default for all Policyfile using chef-client
+runs. We will not inherit the data bag hoisting that poise-hoist
+supports.
+Attributes of the form `policy_group`.`<path>` will be made available as 
+`<path>`, at the level of `role_default` or `role_override`, which is where
```
coderanger: I think I actually use env_default and env_override? That seems more like how I would expect it to work :)

new/hoist.md
```diff
@@ -0,0 +1,58 @@
...
+Policyfile `default` or `override` attributes are levelled currently.
+
+## Downstream Impact
+
+Poise-hoist will not be required in chef client 14.
```
coderanger: I'll probably keep it around for the data bag attributes feature but should be used very rarely :)

new/hoist.md
thommay: https://github.com/poise/poise-hoist/blob/345ffccea8ddee98f2e43ea08cbfeea4a0a070c6/lib/poise_hoist.rb#L41
is definitely `role_default`.

coderanger: Horray for not knowing my own code! I have vague memories of maybe issues with attribute merging?

jonlives:
Do we need to cover handling the chef_environment shim that poist-hoist includes at the moment, or is that not going to be kept?

coderanger:
@jonlives That was handled in https://github.com/chef/chef-rfc/pull/296 so that we didn't have to do it here :)

jonlives:
Oh perfect, I'd even read that RFC too :D

thommay:
Approved @chef/rfc-editors

PR #299, opened 2018-02-24
Fix accidental italics rendering caused by failure to escape underscores

juliandunn:


coderanger:
trolol markdown. Thanks!

PR #301, opened 2018-04-12
ship resource descriptions in cookbook metadata

thommay:


new/resource_metadata.md
```diff
@@ -0,0 +1,122 @@
...
+      }
+      ]
+    }
+  }
+}
+```
+## Downstream Impact
+
+* Stove will be updated to run `chef-resource-inspector` as part of the upload process.
+* Supermarket will add a new tab displaying resources in a cookbook.
```
robbkidd: Should there be a mention of `knife cookbook site/supermarket` having this behavior added, too?

new/resource_metadata.md
```diff
@@ -0,0 +1,122 @@
...
+
+It's now possible to extract detailed API documentation of individual
+resources, but running other people's ruby code on your system to do so
+is unsafe. Therefore, a cookbook's metadata should contain the details of
+the resources contained in the cookbook.
+
+## Motivation
+
+    As a user of Supermarket,
+    I want to view details of resources,
```
damacus: Would this be displayed similarly to how the Terraform registry does it? 👍

robbkidd: The Inputs and Outputs sections there? Yea, that's along the lines we are picturing. With the resource documentation in the metadata.json, Supermarket could be updated to include transforming that into web content associated with a particular version of a cookbook during the upload processing.

damacus: That's exactly what I was thinking. It's really useful 👍

new/resource_metadata.md
```diff
@@ -0,0 +1,122 @@
...
+Resources are a set of resource descriptions generated by the
+`chef-resource-inspector`. Each resource has the following keys:
+
+* Description: The description of the resource
+* Actions: The list of actions available on the resource
+* Examples: Optional code examples
+* Introduced: The cookbook version that the resource was introduced in.
+* Properties: A list of properties that the resource supports.
+
+Each property has the following keys:
```
tas50: We probably want to know if it's a name_property too

PR #304, opened 2018-04-26
Resources can only be added at new major releases

thommay:
We hear repeatedly from our users that adding resources is a breaking change for them. Therefore, we should only do so when appropriate, which is at the time of a yearly major release.
Let's unpack why adding resources is a breaking change:

 * If a user has not updated their copy of the resource cookbook
 * If a user has changed their copy of the resource cookbook

In both cases, the version of the resource in the chef client is the one that will be selected, rather than the one in the user's cookbook. This means that a routine upgrade has the potential to change behaviour that the user relies upon, forcing them to either perform unscheduled work, or, more likely, simply not upgrade their chef client (or worse have to do an emergency downgrade).

This is the most pressing reason to only add new resources once a year.

On top of that, this RFC already specifies that resources must be available and finalised in cookbooks before they're adopted in to core chef; thus this does not impede adoption if a user needs to use the resource.

Additionally, it is much easier to reason about the availability of a resource if one says "This resource is in Chef 14" rather than "this is in 14.1.103".

### DIFF
--- a/rfc100-resource-lifecycle.md
+++ b/rfc100-resource-lifecycle.md
@@ -86,7 +86,7 @@ Factors that influence and inform the decision to deprecate a resource include:
 * Any proposed changes to the interfaces must be implemented in the cookbook prior to adoption.
   * If there are fundamental issues with the resource whether due to naming standards or interface implementation, a new cookbook should be created implementing the resource as intended.
 * Any bugs discovered must be repaired in the cookbook and released prior to adoption.
-* Resource is added to core chef and documentation is updated.
+* During the next major release cycle, the resource is added to core chef and documentation is updated.
 * Add deprecated cookbook warnings for conflicting cookbooks.
 
 ### Process for deprecating resources


coderanger:
This implies that resources pulled into core differ from the version in the cookbook, which we already established was a bad idea and are no longer doing. If we are being rigorous about freezes for resources being added to core (or if not a freeze, make sure it is changed identically on both sides) does that address the problem? This is the kind of thing that should be detected by Test Kitchen testing, though we do have an unfortunate number of users that have no yet joined the Cult of Testing so that's not a complete answer.

Also re: version numbers, we shouldn't be adding them in point releases so the comparison would be "Chef 14" vs. "Chef 14.1".

jaymzh:
This massively slows down development on Chef. If you want to start doing major releases monthly this could perhaps be fesible, but (1) waiting a *year* to get any new feature and (2) waiting a *year* to have your feature available to the world are both terrible experiences.

The only justification here is "it can conflict with a cookbook resource" so why not instead make it so you can mark an internal resource as "secondary" to a cookbook, so it only takes effect when someone stops using the cookbook. You could then mark it as "secondary" while until the new major release?

What about all of the resources that *aren't* in cookbooks first?

jaymzh:
I'm a very strong :-1: on this.

thommay:
> This implies that resources pulled into core differ from the version in the cookbook, which we already established was a bad idea and are no longer doing.

It doesn't, and I specifically covered the cases where it doesn't above.

thommay:
@jaymzh as this RFC states we don't accept any resources that aren't in cookbooks first.

jaymzh:
@thommay where? The closest I see is "  * Any proposed changes to the interfaces must be implemented in the cookbook prior to adoption. " which is saying if it's already *in* a cookbook, you must change any interfaces desired before it's imported into core.

Lots of resources get added to core that don't make sense in a cookbook.

thommay:
> widespread use of a resource based on analysis of publicly available supermarket cookbooks
> well tested and supported cookbook resources

jaymzh:
No, that's under " Factors that influence and inform the decision to adopt a resource include:"

That *can inform* whether something is adopted. It is **not** a requirement. Chef Inc. adds resources without them being in cookbooks. It would have been absurd, for example, to have to make a cookbook to contribute `whyrun_safe_ruby_block` which solved a problem with whyrun mode. There was no risk involved in an orthogonal new resource.

thommay:
Closing in favour of #305

PR #305, opened 2018-05-02
Proposed improvements for the resource adoption/migration process.

coderanger:
The goal is to avoid situations where feature releases can, on their own, destabilize a Chef environment. This is based on conversations from last week's developer meeting.

### DIFF
--- a/new/resource-map-locking.md
+++ b/new/resource-map-locking.md
@@ -47,7 +47,7 @@ name in the map. This can be used by advanced users for things like pointing all
 `package` resources at some new code or other very edge case things.
 
 Any resource which has `allow_cookbook_override: true` set should be marked as
-"experimental" in the documentation. This will help communicate that between
+"preview" in the documentation. This will help communicate that between
 the minor version where a resource is added and the following major release when
 it takes over from cookbooks forcibly is a trial run and we may continue to
 improve the resource (which should be coordinated with the cookbook).
--- a/new/resource-map-locking.md
+++ b/new/resource-map-locking.md
@@ -46,6 +46,12 @@ used in Chef core as part of the RFC100 adoption process. The second is
 name in the map. This can be used by advanced users for things like pointing all
 `package` resources at some new code or other very edge case things.
 
+Cookbooks adding new providers to core resources is less problematic, but for
+consistency will work the same way so it will generally require the use of the
+`__core_override__` option. We may want to revisit this in the future for core
+resources which are most commonly extended like `package` and `service`, but
+this is still rare enough to be called out as a special case.
+
 In order to not disrupt the Chef 14 cycle, we can either change the behavior so
 that during Chef 14 when you `provides` over an existing name, it logs the
 deprecation warning but still adds the name, or we could add `allow_cookbook_override`
--- a/new/resource-map-locking.md
+++ b/new/resource-map-locking.md
@@ -1,0 +1,67 @@
+---
+RFC: unassigned
+Title: Resource Map Locking
+Author: Noah Kantrowitz <noah@coderanger.net>
+Status: Draft
+Type: Standards Track
+---
+
+# Resource Map Locking
+
+To improve the migration process of resources from cookbooks to core, as well
+as prevent user confusion, resources in cookbooks will no longer be allowed to
+override resources in Chef core by default.
+
+The overall goal is enable the RFC100 resource adoption process to run more
+smoothly by allowing us to enforce that the version of a resource in Chef core
+will win priority even with an old version of the cookbook still active, but
+only when we want to (on Chef major version boundaries).
+
+## Motivation
+
+    As a Chef developer,
+    I want to migrate resources to core,
+    so that users can use them.
+
+## Current State
+
+As of this writing, the way resource and provider mappings work is that every
+class providing a given name is ranked based on how "tightly" they bind to the
+name, and then within equal "tightness", the last class to load wins. In
+practical terms, this means that if both Chef core and a cookbook declare a
+resource with identical `provides` lines, the cookbook will win priority.
+
+## Specification
+
+The main new concept would be introducing a "mode" to the resource and provider
+`NodeMap` instances, locked and unlocked. They would start in unlocked mode,
+allowing Chef core to register all its resources and providers as normal. It would
+then be switched into locked mode, where attempting to `provides` on a name
+that is already in use will result in a deprecation warning and the resource or
+provider *not* being mapped. This process is modulated in two ways, both via
+`provides` option flags. The first is `allow_cookbook_override: true`, which
+makes a resource or provider name always behave as if it is unlocked. This is
+used in Chef core as part of the RFC100 adoption process. The second is
+`__core_override__: true`, which allows intentionally remapping over a locked
+name in the map. This can be used by advanced users for things like pointing all
+`package` resources at some new code or other very edge case things.
+
+In order to not disrupt the Chef 14 cycle, we can either change the behavior so
+that during Chef 14 when you `provides` over an existing name, it logs the
+deprecation warning but still adds the name, or we could add `allow_cookbook_override`
+to every core resource.
+
+## Downstream Impact
+
+It is possible some users are already mapping over core resources on purpose.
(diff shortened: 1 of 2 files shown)


coderanger:
As a note, the `defer_until` version doesn't need to be the _next_ major release of the cookbook if we want to allow for more time there. We could also flip the annotation around so the cookbook is in control of when it loses priority, though then we would be more locked in on the next Chef major rev, but that's probably fine.

rfc100-resource-lifecycle.md
```diff
@@ -89,6 +89,64 @@ Factors that influence and inform the decision to deprecate a resource include:
...
+3. The resource is added to core with an annotation of `defer_to_cookbook_until '>= x.0'`,
+   where `x.0` is the next major version for the cookbook in which the resource
+   will be removed.
+4. Chef releases a minor (features-only) version with the new resource. It will
+   remain inert if the original cookbook is active until the given version.
+5. At some point in the future, the cookbook does the major version release. Users
+   can upgrade to it when they feel ready, otherwise they will see no change in
+   behavior even with the new Chef release.
+6. The following April, as Chef prepares for the yearly major release, all pending
+   `defer_to_cookbook_until` annotations will be removed.
```
lamont-granquist: the default behavior of core resources though is to lose to cookbooks.

which means that cookbooks can override the file resource, which is probably dangerous nonsense, but we've been favoring the case of cookbooks iterating faster and people potentially bugfixing their old chef client versions with new cookbooks.

jaymzh: I don't think we'd import a resource into core that overwrote the file resource :)

majormoses: If there was something insane like that I would imagine you would re-write it to more sensible.

rfc100-resource-lifecycle.md
```diff
@@ -78,7 +78,7 @@ Factors that influence and inform the decision to deprecate a resource include:
 * complex implementation that lacks clarity of consequences of use.
 * impact of using the resource incorrectly.
 
-### Process for adopting resources
+### Process for adopting resources from a community cookbook
```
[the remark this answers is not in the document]
coderanger: It was to emphasize the difference between this section and the following one.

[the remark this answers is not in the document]
jaymzh: There's a different process for stuff already cookbook vs other stuff as outlined below.

rfc100-resource-lifecycle.md
coderanger: @lamont-granquist Fair, so we would switch to a new annotation of some kind?

coderanger: Hmm, this might be the reason to switch the annotations around to be driven from the cookbooks side. Need to think through the process >_<

coderanger:
Aside from the details of how exactly the annotations work, are we cool with that table as The Way Things Should Work given each of the 9 options?

coderanger:
Okay, so new words that I think I know how to actually implement. I decided the new `provides` flags and behavior was complex enough to warrant its own RFC so now this is a matinee presentation :) I think the new RFC should allow implementing the behavior of that table.

rfc100-resource-lifecycle.md
```diff
@@ -89,6 +89,62 @@ Factors that influence and inform the decision to deprecate a resource include:
...
+1. A resource is added in a cookbook.
+2. That resource is nominated for core inclusion (see above section).
+3. The resource is added to core with an annotation of `allow_cookbook_override: true`.
+4. Chef releases a minor (features-only) version with the new resource. It will
+   remain inert if the original cookbook is active until the given version.
+5. At some point in the future, the cookbook does the major version release. Users
+   can upgrade to it when they feel ready, otherwise they will see no change in
+   behavior even with the new Chef release.
+6. The following April, as Chef prepares for the yearly major release, all pending
+   `allow_cookbook_override` annotations will be removed.
```
coderanger: An alternative here would be to use a Chef version matcher as the value instead of `true` so it can turn off automatically. But this should always (famous last words) be the next major version following when the resource is added to core, which is already documented in the `introduced` method/annotation. How "always" do we think that will be?

new/resource-map-locking.md
```diff
@@ -0,0 +1,67 @@
...
+Title: Resource Map Locking
+Author: Noah Kantrowitz <noah@coderanger.net>
+Status: Draft
+Type: Standards Track
+---
+
+# Resource Map Locking
+
+To improve the migration process of resources from cookbooks to core, as well
+as prevent user confusion, resources in cookbooks will no longer be allowed to
```
jaymzh: This is already true today - core wins and that's what's caused the problems, I thought?

Perhaps this should say "resources migrated from cookbooks will not override the cookbook until we want it to (on Chef major version boundaries)" ?

new/resource-map-locking.md
```diff
@@ -0,0 +1,67 @@
...
+allowing Chef core to register all its resources and providers as normal. It would
+then be switched into locked mode, where attempting to `provides` on a name
+that is already in use will result in a deprecation warning and the resource or
+provider *not* being mapped. This process is modulated in two ways, both via
+`provides` option flags. The first is `allow_cookbook_override: true`, which
+makes a resource or provider name always behave as if it is unlocked. This is
+used in Chef core as part of the RFC100 adoption process. The second is
+`__core_override__: true`, which allows intentionally remapping over a locked
+name in the map. This can be used by advanced users for things like pointing all
+`package` resources at some new code or other very edge case things.
```
jaymzh: I like this idea of explicitness.

new/resource-map-locking.md
tas50: Core wins because we put lines like this into any resource in order to try to make core more specific:

```ruby
provides(:dmg_package) { true }
```

If someone had the same thing in a cookbook it would win.

rfc100-resource-lifecycle.md
jaymzh: I like the idea of the version matcher... 

Though shouldn't those get cleaned up eventually anyway?

rfc100-resource-lifecycle.md
```diff
@@ -89,6 +89,62 @@ Factors that influence and inform the decision to deprecate a resource include:
...
+
+```
++-------------+--------+--------+-----------+
+|             | old cb | new cb | future cb |
+|             | (3.4)  | (3.5)  | (4.0)     |
++-------------+--------+--------+-----------+
+| old chef    | none   | cb     | none      |
+| (15.2)      |        |        |           |
++-------------+--------+--------+-----------+
+| new chef    | core   | cb     | core      |
```
jaymzh: shouldn't new chef old cookbook be 'cb' ?

jaymzh:
As mentioned in the Slack meeting today, I think we should probably be explicit that the `__override_core__` functionality would enable a "subprovider" (ala "I need to implement `service` for me weird custom thing"), but other than that I am :+1: on this.

lamont-granquist:
So, the alternative way of doing things would be to still continue to synch cookbooks from core into the cookbooks for some time afterwards, but to gate on the cookbook side which version of chef-client the cookbook would override.   This is attractive because then either people on the latest cookbook or people on the latest chef version would get up-to-date resources.  It is more the model that the compat_resource cookbook took towards the end -- and the back-porting of resources was the more successful part of compat_resource (the tirefire of monkeypatching core chef features over old chef-client versions was where it failed).  We can also be more careful here since we we can use the ProviderResolver to avoid using monkeypatching at all and not be creating frakenclasses, but just have the cookbook version of the resource and the core version of the resource and switch between them based on chef-client version.  That has the advantage in that if you're old `old chef + new cookbook` or on `new chef + old cookbook` you get the up to date resource behavior.

thommay:
https://github.com/chef/chef/issues/7256 is an ideal example of the problem our users face when we add resources.

coderanger:
But also an example of a case that the proposed new process/API would have detected, warned about, and ensured no changes would happen until a major release :)

thommay:
> But also an example of a case that the proposed new process/API would have detected, warned about, and ensured no changes would happen until a major release :)

Totally - but it's good to have examples of *why* we're doing it.

PR #306, opened 2018-05-12
Ohai new plugin policy

jaymzh:


new/ohai-new-plugins-policy.md
```diff
@@ -0,0 +1,45 @@
...
+    I want to be able to provide new functionality to the community
+    so that they can use it.
+
+    As a Chef user,
+    I want to be able to leverage new functionality as quickly as possible
+    with minimal work while not breaking existing features.
+
+## Specification
+
+As of Ohai 14, most plugins are disabled by default. New plugins have no effect
```
coderanger: Only a few plugins should be disabled by default unless something major changed?

jaymzh: I can go look at the code. I was under the impression we switched a whitelist by default.

jaymzh: That said, even if they are run, again as long as they don't effect existing namespaces, I see no way in which this breaks compat.

jaymzh: Ah, OK, so I misunderstood the `optional` feature, it's off by default not on. So **ideally** I'd want the policy to be:

* As long as you're not touching an existing namespace, it's fine to add

But I'd see at as reasonable to say:

* As long as you're not touching an existing nsmaespace, and it's marked optional, it's fine to add

Perhaps what the best compromise here is:

* In current, as long as you're not touching an existing namespace, it's fine to add
* In stable (current - 1), as long as your'e not touching an existing namespace and it's marked as optional, it's find to add

new/ohai-new-plugins-policy.md
```diff
@@ -0,0 +1,45 @@
...
+policy for clarity.
+
+## Motivation
+
+    As a Chef developer,
+    I want to be able to provide new functionality to the community
+    so that they can use it.
+
+    As a Chef user,
+    I want to be able to leverage new functionality as quickly as possible
```
thommay: I don't think this reflects anything but the smallest minority of our user base at this point.

thommay:
The majority of the feedback that we at Chef receive is that our users want Chef to provide stable releases that don't break anything. Most users these days are doing the work of upgrading to a new major version of the chef client once a year, and between those times they do the work that they absolutely have to do - upgrading for security releases, or when deploying new software.

This week has reiterated the fact that Ohai plugins have a much larger blast radius than we might expect - the changes to the shard plugin, even though they were reviewed and thought correct, completely broke 14.1 on any FIPS platform. We also see ongoing reports of users having their chef runs break unexpectedly because systemd doesn't clean up sessions and one day their node object goes over 10mb (chef/ohai#1089).

On a personal level, I'm totally against shipping new plugins in Ohai that are turned off out of the gate. These plugins will never benefit the majority of our users and just increase the cognitive load of the maintainers, and are highly likely to bit rot (and be broken on less used platforms, like filesystem). If we think a plugin is important enough to ship with Ohai, it should be on by default. And we should be testing that importance by shipping the plugin in a cookbook first and getting user feedback.

jaymzh:
Much like our recent policy of having a 'preview' of a new resource in a Chef version which is masked by a cookbook, and then having it 'win' in the next version, there's no reason we can't do that here (optional first, then non-optional).

The idea of changing things once a year is antithetical to what Chef stands for. Continuous Delivery. We don't have Developer Meetings once a year. We don't write code once a year. And Chef has been yelling at their customers for years now about how bad it is to store up new code for a year and then release it all at once with tons of changes it's hard to track down. This is *literally* you pushing for the opposite of what Chef Inc. has been asking its customers to do since Day 1: push more often, push more often, push more often.

I understand API changes, complicated refactors, etc. - I'm all for providing for a reasonable level of stability... but 0 changes to Chef/Ohai for a whole year, every year, is exactly the opposite of what we're all trying to build.

coderanger:
@thommay I think we also need to keep a strong non-response bias in mind re: "The majority of the feedback that we at Chef receive ...". I think there is a strong case that the "status quo" expectation for most open-source projects is that they add features in new releases, so there probably isn't going to be a lot of feedback in support of that, whereas the ones urging more careful movement are encouraged to be loud about it as they are the ones having problems.

The worry with adding new Ohai plugins is a bit different from new resources, it's not usually a compat with cookbooks problem, more often the risk is stomping on top of cookbook attributes because `automatic` outranks them inherently. We've talked in the past about splitting Ohai data into its own API rather than muxing it into the other node attributes, which would help with that. There is still some risk of colliding with Ohai plugins from cookbooks, but those are both rare and I think the cookbook version currently wins in all cases (and since it's less common than migrating resources to core, I'm less worried about that as a long-term problem).

That said, I think the examples of me screwing up with the Shard plugin (which was 100% my bad, not really a structural problem except insofar as we don't actually test FIPS mode) or how the Passwd plugin causes issues are super related to the issue of when/how to add new plugins in the same way as the discussion about resources was shaped. That was about ensuring existing cookbooks continue to work. The counter-argument here seems to be more along the lines of "we might do it wrong" (a la me and Shard), but that's just the reality of software. There is no way to make changes without a risk of breakages, always. Making changes more slowly doesn't inherently make things more stable, for that we would need to double-down on test coverage and documentation :)

jaymzh:
Updated the text of the RFC to reflect changes in discussion.

thommay:
It literally isn't the opposite of what Chef wants to do. We can, should, and do make as many changes to the *current* version of Chef and Ohai.
We should not, and have not, made releases to the n-1 version of Chef and Ohai, and given that the desire to do that seems to be the only reason for this RFC I was responding to that.

jaymzh:
@thommay - you were pushing back against making changes *in current*. You specifically proposed an RFC to say no new providers **in current**. So, no, this current-1 not "the only reason for this RFC" - with you and Tim pushing hard against any changes **in current**, I'm trying to clarify policy for **both** current *and* current-1, which btw, is better called "stable".

What I'm worried about is you cut the 15 branch in a week and don't want to merge anything to 14 anymore. I'm also vaguely worried about getting stuff into 13 over the next few months until people reasonably have a chance to make the changes requires to get to 14... though I'm **more** worried about the tendency to not want changes in **current**.

thommay:
It seems that my intent is unclear. I'm trying to ensure that we have a development environment that doesn't lead to burn out in the three months or so leading up to the next major, where we can safely do large changes, but at the same time continue releasing features in a way that is safe for our users to consume.

jaymzh:
I think the disconnect is what we define as breaking changes. You (I think) wanted to define "all new resources" as "breaking changes" - this seemed untenable to many of us, but we we found another definition we could all agree on.

It seems like you find "new ohai plugins" to be "potentially breaking changes" - that also seems untenable to me, so I'm here trying to define what is a breaking and non-breaking change, so all know what we can expect to merge and get merged to both current and stable.

thommay:
I am against *adding features* to the *n-1/stable* release of Chef or Ohai.

jaymzh:
OK, so let's focus on **current** then. You felt that new resources were a possibly "breaking change" (because of the `windows_task` fallout). We found a way to make that not a breaking change. I don't feel like new Ohai plugins are a breaking change. Do we agree on that? That's like half this RFC by itself. :)

jaymzh:
It's not. Here's @tas50  in https://github.com/chef/ohai/pull/1171 blocking a PR for the **current** release (13 at the time):
```
I'm -1 on this one. It's going into 14 as a disabled plugin, but we can't do that in 13. That makes this a breaking change at the end of a stable cycle, which doesn't seem like we're doing users a service.
```

This was in March, before Chef 14, while Chef 13 was **current**.

So again, I'd like to clarify what the policy is. Is the policy "we don't add anything to **current** in the last month of it being current? Or the last 3 months? Or the last 6 months?

thommay:
You're splitting incredibly fine hairs. That was **one week** before the expected release of 14, and clearly at that point we knew that the user experience would be different, we'd be - as Tim says - shipping something in 13 and immediately turning it off in 14.

jaymzh:
You clearly have a policy in your head somewhere. Can we write it down, so that I as a maintainer know what to merge, and I as a contributor know what to expect? I don't think this is unreasonable.

thommay:
As I said in the developer meeting, I'm happy to write an RFC that defines what goes in to n and n-1 at all levels.

jaymzh:
@thommay I updated n-1 to be only bugfixes and reworked this to clarify what can and can't go into current. Seem better for you?

new/ohai-new-plugins-policy.md
```diff
@@ -0,0 +1,60 @@
...
+    with minimal work while not breaking existing features.
+    
+    As a Chef user,
+    I also want to be able to discover available plugins.
```
tas50: I think this is an important part of this RFC, but we're not mentioned it below. If we continue to add more plugins we really need to build the functionality to list installed plugins with descriptions. That should be included in this RFC

PR #308, opened 2018-06-05
The Future of Chef Workstation

coderanger:
I want to start off by congratulating the Workstation team for launching an awesome new tool. It's an absolutely critical step towards improving our new user experience and I look forward to seeing it develop in the coming months.

In the week or two since ChefConf, there has been much discussion about where the new `chef-run` tool should live in the greater Chef ecosystem. This RFC addresses this question with two main parts:
* the `chef-run` tool will be renamed (currently to `chef-target` but we don't love that name so suggest better ones below)
* ChefDK as a whole will be renamed to "Chef Workstation"

Speaking personally: I'm not 100% sold on this rename. I think it's going to take a good bit of development time and resources to make the upgrade path smooth that could be spent elsewhere to greater effect. I think we as a community may over focus "silver bullet" solutions when the underlying problem with both our new user experience and CLI experience is many years of organic tool growth and minimal tech debt paydown. That said, the name does seem better so even if it's not the best ROI, it's still moving in a good direction, if only a small step.

This RFC will affect almost all of our community, so even if you're never participated in the Chef RFC process before, I encourage you to drop a comment on here with your thoughts or feelings. We value the input of every community member, no matter how active or well-known.

### DIFF
--- a/new/chef-workstation.md
+++ b/new/chef-workstation.md
@@ -1,7 +1,9 @@
 ---
 RFC: unassigned
 Title: The Future of Chef Workstation
-Author: Noah Kantrowitz <noah@coderanger.net>
+Author:
+- Jon Morrow <jmorrow@chef.io>
+- Noah Kantrowitz <noah@coderanger.net>
 Status: Draft
 Type: Informational
 ---
@@ -27,66 +29,76 @@ lack of certainty about which should be used.
 
 ## Specification
 
-Addressing the issues in reverse order:
+1. The `chef-run` command will be renamed to `chef-target` [ed: not attached to this name, suggest a better on in comments].
+2. The `chef-target` gem will be split into its own repository for future development.
+3. The installer currently called "ChefDK" will be renamed "Chef Workstation".
+   This will require setting up documentation to explain the rename, as well as
+   redirects and supporting pages in the download system to ensure users don't
+   get lost.
+4. The `chef-dk` gem will not be renamed at this time, as few users see the gem
+   name there and we're already changing a lot of things. This may be revisited
+   in the future when the dust settles.
+5. The installer for `chef-dk` (now named "Chef Workstation") will be reconfigured
+   to install into `/opt/chef-workstation` and use `~/.chef-workstation` as the
+   primary configuration folder. A symlink for `/opt/chef-dk` (and related windows-y paths)
+   should be added to ensure any scripts that hardcode paths like `/opt/chefdk/bin/chef`
+   continue to function (until such time as we decide to remove them as compatbility layers).
+   Gems in `~/.chefdk/gems` should be added to the gems path, and config files in
+   `~/.chefdk` should work if no file in `~/.chef-workstation` takes priority.
+6. Add the `chef-target` gem to the Chef Workstation installer.
+7. A command stub will be added to the `chef-dk` command processor called `target` [ed: or whatever we call it],
+   which will dispatch to the `chef-target` command. The Chef Workstation installer
+   will map `chef-target` into `embedded/bin/` rather than `bin/` so for most users,
+   this will be the UX.
 
-While the phrasing of "Chef Workstation" is probably better than "ChefDK", the
-difference is not sufficient to warrant such a massive community pivot given
-how long the ChefDK name has been in use and the fact that it doesn't seem to
-be a major source of confusion (at least not in a way that would be solved by
-"Chef Workstation"). As such the Chef Workstation installer will be discontinued
-in favor of ChefDK.
-
-In order to both reduce the naming collision and better integrate the tool with
-the existing ChefDK CLI experience, the `chef-run` command will be moved to be
-a subcommand of the `chef` CLI (a part of the `chef-dk` gem) named `chef target`.
-
-The actual implementation of the `chef target` command could still be in its own
-gem (name undetermined) if that is useful to the development team to improve the
-agility of releases.
+Additionally, while it is out of scope for this RFC, the authors strongly encourage
+future RFCs and discussions about adding additional tools to Chef Workstation/DK
+that are of value to the Chef community.
 
 ## Alternatives
 
-### Keeping the "Chef Workstation" Name
+### Keeping the "ChefDK" Name
 
-If the improved clarity of purpose of the "Chef Workstation" package name is
-something we want to keep, an alternate path would be to do the same `chef-run`
-to `chef target` switch but then rename the ChefDK installers to Chef Workstation.
-This presents significant switching costs to the community in the form of "what
-happened to ChefDK?" and general confusion about the rename, given that ChefDK
-has been used as a name for quite a long time now. But this could be overcome
-with sufficient documentation and redirects of existing ChefDK-related pages.
+While "Chef Workstation" is a clearer name, there is broad community usage and
+understanding of the ChefDK name and brand. We could do a similar process to the
+above, but leave the installed called ChefDK in the end. This would vastly
+simplify the process as we could skip all the compatibility gunk. This would
+just be adding a new tool to ChefDK, as we have done many times before.
 
-In either case, the actual software in the installer is the same, the question
-is is the improved branding of "Chef Workstation" worth the hit to users as we
-switch over and people learn the new term.
+The downside is that there is some pushback against the "Developer Kit" part of
+ChefDK from people that feel the "developer" label is exclusionary or at least
+presents a barrier to the new user experience.
 
-If we go this route, we'll need to work out an upgrade strategy for existing
-ChefDK users who may have `/opt/chefdk` paths in scripts and `~/.chefdk`
-for installed secondary gems. This would likely involve a symlink from `/opt/chefdk`
-to the new `/opt` folder for a while (though eventually it would have to get cleaned
-up so this only delays migration pain for some). Gem installs would have to
-migrated from a user context, possibly with a provided script or `chef` CLI
-command.
+### Making a Dedicated Chef-Target Installer
 
-### Use "Chef Workstation" as a Different Brand
+Rather than including `chef-target` in the DK/Workstation installer, we could
+move it to a focused installer just for the one tool, similar to the InSpec
+installer. This would give the team more agility as they would have much more
+freedom in shaping the UI and UX of this new workflow.
 
-My understanding (which is very incomplete) is part of the initial reason to
-split `chef-run` to its own command was, in part, concern about the growing sprawl of
-commands in the `chef` CLI which could be a speedbump for new users.
+The downside is that this would split the new workflow off from the rest of the
+Chef community, possibly creating friction for users experimenting with new workflows
+for the first time or switching from an existing workflow.
 
-Another option is to leave both "ChefDK" and "Chef Workstation" as active brands,
-with ChefDK continuing as it is today while Chef Workstation becomes a focused,
-single-workflow installer (i.e. remove Berkshelf and other tools not part of the
-`chef-run` workflow) from it. This could mirror the focused installers built by
-the InSpec team.
+### Using "Chef Workstation" as Its Own Brand
 
-I feel like the very near overlap of ChefDK and Chef Workstation in this universe
-would result in long-term user confusion though.
+In this path, we would still add `chef-target` to ChefDK, but we wouldn't
+rename the installer. Instead, "Chef Workstation" would become a new thing,
+explicitly aimed at being a cross-product-line workstation experience, while
+ChefDK stays focused on the Chef (the project) experience.
+
+The downside here is the limitations of human language (more or less). Because
+Chef Software and Chef (the project) have such overlapping names, it would be
+very difficult to explain how these are two different things in the long term,
+likely resulting in user confusion and frustration.
 
 ## Downstream Impact
 
-All current users of the beta Chef Workstation installer would have to eventually
-remove it and install ChefDK to get future updates.
+All current users of ChefDK would need to upgrade to Chef Workstation, though
+this upgrade should be transparent.
+
+Any users of the current beta Chef Workstation installer would also need to
+switch over, though that would not be a transparent upgrade process.
 
 ## Copyright
 
--- a/new/chef-workstation.md
+++ b/new/chef-workstation.md
@@ -1,0 +1,96 @@
+---
+RFC: unassigned
+Title: The Future of Chef Workstation
+Author: Noah Kantrowitz <noah@coderanger.net>
+Status: Draft
+Type: Informational
+---
+
+# The Future of Chef Workstation
+
+ChefConf 2018 saw the unveiling of both the `chef-run` command and a Chef Workstation
+installer package containing it. This has lead to user confusion on two fronts.
+First, the term "Chef run" (or "Chef client run") is already in very common use
+in our community, so trying to communicate about `chef-run` has been tricky.
+Second, having two installers (Chef Workstation and ChefDK) has resulted in a
+lack of certainty about which should be used.
+
+## Motivation
+
+    As a Chef user,
+    I want to use the new Chef functionality,
+    so that I can manage machines/resources without Chef Server.
+
+    As a Chef community member,
+    I want to communicate with other Chef community members,
+    so that I can share knowledge and answer questions.
+
+## Specification
+
+Addressing the issues in reverse order:
+
+While the phrasing of "Chef Workstation" is probably better than "ChefDK", the
+difference is not sufficient to warrant such a massive community pivot given
+how long the ChefDK name has been in use and the fact that it doesn't seem to
+be a major source of confusion (at least not in a way that would be solved by
+"Chef Workstation"). As such the Chef Workstation installer will be discontinued
+in favor of ChefDK.
+
+In order to both reduce the naming collision and better integrate the tool with
+the existing ChefDK CLI experience, the `chef-run` command will be moved to be
+a subcommand of the `chef` CLI (a part of the `chef-dk` gem) named `chef target`.
+
+The actual implementation of the `chef target` command could still be in its own
+gem (name undetermined) if that is useful to the development team to improve the
+agility of releases.
+
+## Alternatives
+
+### Keeping the "Chef Workstation" Name
+
+If the improved clarity of purpose of the "Chef Workstation" package name is
+something we want to keep, an alternate path would be to do the same `chef-run`
+to `chef target` switch but then rename the ChefDK installers to Chef Workstation.
+This presents significant switching costs to the community in the form of "what
+happened to ChefDK?" and general confusion about the rename, given that ChefDK
+has been used as a name for quite a long time now. But this could be overcome
+with sufficient documentation and redirects of existing ChefDK-related pages.
+
+In either case, the actual software in the installer is the same, the question
+is is the improved branding of "Chef Workstation" worth the hit to users as we
+switch over and people learn the new term.
+
+If we go this route, we'll need to work out an upgrade strategy for existing
+ChefDK users who may have `/opt/chefdk` paths in scripts and `~/.chefdk`
+for installed secondary gems. This would likely involve a symlink from `/opt/chefdk`
+to the new `/opt` folder for a while (though eventually it would have to get cleaned
+up so this only delays migration pain for some). Gem installs would have to
+migrated from a user context, possibly with a provided script or `chef` CLI
+command.
+
+### Use "Chef Workstation" as a Different Brand
+
+My understanding (which is very incomplete) is part of the initial reason to
+split `chef-run` to its own command was, in part, concern about the growing sprawl of
+commands in the `chef` CLI which could be a speedbump for new users.
+
+Another option is to leave both "ChefDK" and "Chef Workstation" as active brands,
+with ChefDK continuing as it is today while Chef Workstation becomes a focused,
+single-workflow installer (i.e. remove Berkshelf and other tools not part of the
+`chef-run` workflow) from it. This could mirror the focused installers built by
+the InSpec team.
+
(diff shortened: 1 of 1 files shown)


troyready:
Very well put.

I generally agree with the main proposal. I also think the first alternative (keeping the ChefDK name) would be fine as well, but don't have any strong feelings either way.

btm:
The mechanics of the rename from DK to Workstation herein are all fine. I don't want to overload one RFC, but this looks mostly like the 'rename RFC' to me. It comes off a bit as "here is a better name" but the promise here is that ChefDK was everything you needed for developing cookbooks, and Chef Workstation is the evolution of that into a package to guide you on your automation journey. It's more about the automation user experience than being a package of blessed tools.

1) ChefDK 3 (recently released) is the last major version of ChefDK, and per [RFC086](https://github.com/chef/chef-rfc/blob/master/rfc086-chef-oss-project-policies.md) and traditional expectation we'd support it until April 2020 as the n-1 release.
2) Workstation is effectively ChefDK 4, and we'll have to figure out what it's major release cadence will look like, but outside of acknowledging it's painful to have lots of breaking changes in a short time period, that's premature.

The real value of Workstation beyond the name:
3) Someone propose separately/next what the minimal/expected user experience is for sub-commands in Workstation. Workstation has had [a lot of work put into a user experience](https://blog.chef.io/2018/05/23/introducing-chef-workstation/) like meaningful error messages, I think we'd all LOVE to see that across the board.
4) the Chef Workstation team would then look at workflows that we use knife/chefdk for and build that experience into workstation one at a time. For example [The Great Workflow RFC019](https://chef.github.io/chef-rfc/rfc019-chef-workflows.html) suggested `chef test unit && chef test integration && chef upload` but we never got there also we did get the `delivery local` [tooling in delivery-cli](https://docs.chef.io/delivery_cli.html#delivery-local), in Chef DK.

Then:
5) When we have a migration between DK3 and Workstation figured out, and there's a couple great journeys built in for users, we'd see Workstation 1.0 or Workstation 4.0 or something.

Today is just about how we take the great work the Chef Workstation team has started and reconcile it with what we have today so we can all contribute to achieving that reality together. So we should probably take a stab at figuring out 1 & 2 here, and 3-5 are just things we want to get over the next year for the trouble of renaming.

coderanger:
@btm I think you were mostly using the term "Chef Workstation" there to refer to the `chef-run` tool? We should take care here to be specific as it will be easy to miscommunicate if we use the same term for both the tool and the installer :)

btm:
@coderanger Actually I mean Chef-Workstation-DK4. I haven't been part of the Chef Software team working on Chef Workstation, but my understanding is they started with "How do we make Chef easy to start with," went down the road which lead to `chef-run` / `chef-target`, and then discovered what we really needed was a truckload of a better experience than ChefDK.

As a community, DK has built up into this big package of tools that's on maintenance life support. A lot of our vision of a better Chef workflow (lower-case workflow intended) around the `chef` binary early on, in [RFC019](https://chef.github.io/chef-rfc/rfc019-chef-workflows.html), in `delivery local`, ended up without a team to continue to implement the vision.

So when I caught wind of the Chef Software team's efforts my reaction was "awesome, that sounds just like the vision of DK but for today." So while `chef-run` / `chef-target` is awesome, I'm much more interested in there being a team again focused on cleaning up the DK/Workstation experience.

I'm notably biased by my own vision of what I think they can do, but for me the Workstation name is about a real feature-ful and better DK4.

coderanger:
Gotcha. I don't think anyone is objecting to having a team to clean up the desktop experience and that will continue regardless of the decisions (or lack thereof) reached here :) The question is very specifically how to move forward with maximum benefit to the community and minimum disruption as trying to do both in parallel is probably not the best option (though maybe it is).

btm:
@coderanger :+1: mechanically speaking, I prefer thinking of Workstation as DK4 rather than just a rename, but also brings up the mechanics of renaming timing, is that a major release, how does that affect our April release cadence, etc.

jakauppila:
Is there any concern around the `Workstation` name being attributed strictly to an `end-user` machine? That was my initial impression when I saw the name.

With `chef-run`/`chef-target` I see it as potentially being an important component to home-grown deployment pipelines. For ours today (utilizing Jenkins), we simply reach out to the target servers via WinRM to kick off a Chef run via scheduled task, I'll be exploring if that changes with this new tooling available.

Thinking about it initially (with the above statement in mind), I would prefer to see `chef-target` to be a self-contained installer as from a pipe-lining perspective, that's likely the only component I would care about deploying on those servers whereas the rest of the `ChefDK` is for development.

From a release standpoint, as `chef-target` gains momentum both from a development and capabilities standpoint, I would expect there to be much more iteration than what may be present on the `ChefDK` side of things. So does continuously releasing new `development` tool-set versions when it's really just a component make sense?

jjasghar:
I realize making this statement can force a bike shed very quickly; but I was thinking about the name and wanted to offer my suggestion. 

In keeping with the cooking naming conventions I suggest “mise” so it would be chef-mise or chef mise as the command. 

Mise is short hand for [mise en place](https://en.m.wikipedia.org/wiki/Mise_en_place) which is a prep area and workspace for chefs. 

I feel like it captures the core of what this package does and aims for.

Thoughts?

thommay:
> In keeping with the cooking naming conventions I suggest “mise” so it would be chef-mise or chef mise as the command.

The cooking name conventions get us in a lot of trouble because they're totally opaque to our users; rather than being descriptive of why one might use the command. We'd like to move away from them, rather than add more (boring as that is).

tyler-ball:
> Is there any concern around the Workstation name being attributed strictly to an end-user machine? 
@jakauppila TL;DR - I think we do want Workstation focused on an end-user machine.

One thing we talk about is the personas we are targeting. I think there a "end-user" persona who needs a good ad-hoc UX for managing remote nodes, and thats exactly what we developed `chef-run` to target. A lot of the UX we designed (and potential improvements we talked about) is moving towards _more_ interactivity in the interest of making the tool easy to use for less advanced users.

Then I think there is a "CI focused" persona (I'm using that term really loosely here, because both those personas could be embodied in one person, and what they really want to do is manage their Chef infrastructure) that wants to run `chef-client` on a remote machine. I think both of these are good problems we want to tackle but I'm not sure if we want to tackle them with the exact same tool.

In my experience, interactive tools do not embed well into automated processes like CI pipelines.

The Chef workstation team has theorized that we could pull some of the current ChefDK components into more API focused libraries. Those libraries could be leveraged both by a UX focused layer in the Workstation package as well as other ways, like a CI pipeline or the Automate server. Thats a larger change we would need to work through in another RFR, but it is an idea we have talked about.

In the short term I think it would be great to see `chef-run` used in a CI pipeline to trigger chef client runs. I would love to hear your feedback about that because its definitely a problem users have so we want to solve it well for them. But in the long term I hypothesize that we want different end-user focused products/experiences from CI focused products/experiences.

new/chef-workstation.md
```diff
@@ -0,0 +1,108 @@
...
+    so that I can manage machines/resources without Chef Server.
+
+    As a Chef community member,
+    I want to communicate with other Chef community members,
+    so that I can share knowledge and answer questions.
+
+## Specification
+
+1. The `chef-run` command will be renamed to `chef-target` [ed: not attached to this name, suggest a better on in comments].
+2. The `chef-target` gem will be split into its own repository for future development.
```
tyler-ball: What is the motivation for moving this into its own repository?

new/chef-workstation.md
```diff
@@ -0,0 +1,108 @@
...
+
+Additionally, while it is out of scope for this RFC, the authors strongly encourage
+future RFCs and discussions about adding additional tools to Chef Workstation/DK
+that are of value to the Chef community.
+
+## Alternatives
+
+### Keeping the "ChefDK" Name
+
+While "Chef Workstation" is a clearer name, there is broad community usage and
```
tyler-ball: I think we could potentially call out what "clearer" means - if we're on board with my previous comment about `Chef Workstation` being focused on end-user desktops and people who need to manage their Chef infrastructure from their laptop, we should be explicit about that.

new/chef-workstation.md
```diff
@@ -0,0 +1,108 @@
...
+Type: Informational
+---
+
+# The Future of Chef Workstation
+
+ChefConf 2018 saw the unveiling of both the `chef-run` command and a Chef Workstation
+installer package containing it. This has lead to user confusion on two fronts.
+First, the term "Chef run" (or "Chef client run") is already in very common use
+in our community, so trying to communicate about `chef-run` has been tricky.
+Second, having two installers (Chef Workstation and ChefDK) has resulted in a
```
tyler-ball: 👍 to not having 2 installers for the "workstation" experience (whatever that installer ends up called). I think thats one of my biggest motivations for this RFR - I want to converge on 1 product ASAP.

new/chef-workstation.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Specification
+
+1. The `chef-run` command will be renamed to `chef-target` [ed: not attached to this name, suggest a better on in comments].
+2. The `chef-target` gem will be split into its own repository for future development.
+3. The installer currently called "ChefDK" will be renamed "Chef Workstation".
+   This will require setting up documentation to explain the rename, as well as
+   redirects and supporting pages in the download system to ensure users don't
+   get lost.
+4. The `chef-dk` gem will not be renamed at this time, as few users see the gem
```
tyler-ball: I'm sensing a common theme that we may want to address as part of this RFR (or another RFR). Basically: Is Chef Workstation / ChefDK an actual tool with its own functionality plus a bunch of packaged components? Or is it just the installer that bundles a bunch of components, one of those components being a `chef` command?

I think I prefer the latter. Thoughts?

new/chef-workstation.md
```diff
@@ -0,0 +1,108 @@
...
+    I want to use the new Chef functionality,
+    so that I can manage machines/resources without Chef Server.
+
+    As a Chef community member,
+    I want to communicate with other Chef community members,
+    so that I can share knowledge and answer questions.
+
+## Specification
+
+1. The `chef-run` command will be renamed to `chef-target` [ed: not attached to this name, suggest a better on in comments].
```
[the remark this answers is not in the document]
jakauppila: `chef-execute` ?

If inventory support is added, will it be clear that you're targeting a single node vs multiple?

new/chef-workstation.md
jonsmorrow: I tend to agree with the idea of an installer that bundles components as a longer term vision. One caveat might be components such as a tray app and/or management cli for the package itself.

new/chef-workstation.md
coderanger: Because we probably shouldn't have repos called both `chef-dk` and `chef-workstation` if we decided to rename things in one way or the other, and putting it in its own repo will allow for a faster release cycle (if it ends up in the `chef-dk` gem/repo or a renamed echo of that then it would be more bound by the monthly release cycles). Also just in general, doing releases from a monorepo is super awkward for users, you end up with problems like multiple sets of release tags in one repo which can be very non-obvious for users not used to working with git monorepos.

new/chef-workstation.md
coderanger: This is why I tried to draw a sharp distinction between "the tool" and "the installer", but that said, we should still have one primary entry point for new Chef (the project) users, and if that isn't going to be the `chef` CLI, that probably needs some discussion and a lot of justification (since `chef` was already our "do over" from `knife`).

new/chef-workstation.md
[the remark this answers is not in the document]
jonsmorrow: I suspect once we do inventory we will want the ability to target those host with system native commands and chef-execute feels like a good candidate for that.

marcparadise:
Reading this RFC and the discussion, it seems that there are really three different things being proposed: 

1. Combine ChefDK and Chef Workstation  into a single Chef Workstation repository
2. Rename chef-run 
3. Inclusion of chef-run into ChefDK 

There are different discussions to have around each, and the work would occur on different timelines.  While there is some interdependency, I feel  keeping the discussion for all three under a single RFC will make it more difficult to come to clear consensus since it's a package deal. 

In order to keep the discussions more tightly focused, I propose that we split this into three RFCs.

coderanger:
@marcparadise I wouldn't phrase number 1 quite like that, but something on that axis (renaming the output(s) moreso than being about the repositories). Unfortunately I think time is tight for all three of those needs so we should be deciding them all here, now. Every day that passes makes the status quo more of a factor.

jakauppila:
I feel like it would make sense to have two distinct repos `chef-target` and `chef-dk` so releases of the individual pieces can be disjointed.

The user installs `workstation` which contains the `chef-target` and `chefdk` components.

coderanger:
From some quick Slack chats, I wanted to throw in what I think should be the overarching goal (but this is just my opinion, not imbued truth):

> There should be one-- and preferably only one --obvious way to do it.

This is almost certainly my Python heritage showing again, but I think this should be the primary guiding principle for resolving this UX concern. There should be one obvious entry point in terms of what to download and install, and then after that there should be one obvious entry point in terms of commands to run. The first part I think we can resolve here (and with some improvements to docs and the downloads site), the second is a bit more long-term (sigh at `knife`) but we can at least make sure we're moving in that direction.

marcparadise:
> . Unfortunately I think time is tight for all three of those needs so we should be deciding them all here, now. Every day that passes makes the status quo more of a factor.
I do agree - the suggestion above was to help us get to the decisions more quickly than I suspect we will when  the questions are tied together.

jonsmorrow:
> I feel like it would make sense to have two distinct repos chef-target and chef-dk so releases of the individual pieces can be disjointed.
>
> The user installs workstation which contains the chef-target and chefdk components.

I think this would imply 3 repos. One for the Chef Workstation packaging, one for chefdk code, and one for chef-run or whatever we call it. This doesn't feel too bad as long as we commit to saying the entry point is Chef Workstation. In this model it doesn't really matter if the dk installer hangs around or not because we all direct folks to the Workstation one. There's also some nicety here in that we start separating the packaging concern from functional parts of the code. If we went this way it seems to imply follow on work in the dk code base to pull out it's current functional bits `chef`.

mchiang0610:
Chef’s goal and vision of Chef Workstation is to build a workstation (end-user) experience and product for users and customers of Chef; this may very well include features expanding beyond what’s currently included in ChefDK and outside of pure infrastructure automation play - such as experiences around compliance or application automation. 

We need to be sure that by merging the two, we don’t limit Chef Workstation to purely enhance and build out the features currently in ChefDK, but allow us to explore and iterate on more use cases.

jonsmorrow:
> Reading this RFC and the discussion, it seems that there are three different but loosely-related things being proposed:

I think the key here is sussing out the order of operations. If we can get to agreement on everything cool if not I'm all for pulling pieces out so we can get to work on those specific pieces. For me I think the order and timing are something like:

1. Decide if we rename chef-run
2. Do the work to rename and move it to dk
3. Do merge whatever that ends up being technically

coderanger:
@mchiang0610 As stated in the RFC (though this is already the case and will remain no matter what), ChefDK (or whatever it is called in the future) will always be open to new projects. Anyone can open an RFC or PR to add a new tool/lib/whatever to it, just need to explain the benefits to users and get the maintainer team on board.

jonsmorrow:
I think @mchiang0610 is asking for commitment to the broader vision of moving DK/Workstation's stated focus/goal to be CHEF the company as opposed to Chef the project. There is a bit of precedent here with Inspec and compliance automation, but I think that mostly came along for the ride with kitchen-inspec stuff and there probably wasn't a real conversation of expanding functional scope. (I'm guessing here a bit as i wasn;t involved). Inspec and Chef are arguably pretty close and two sides of a coin, whereas habitat is a bit of a different animal but we currently think there's going to be some really interesting experiences we can build if you are using all our products together. There's a lot of things to figure out on that path and likely a lot more rfcs but general agreement that we want it to evolve in that direction is important.

jonsmorrow:
I want to provide a bit of context around how we came up with `chef-run` for the name. We originally had `chef target converge` and got a lot of user feedback that target was new language and converge required a lot of explaining up front. Also so few of us say converge unless describing the underlying mechanics. We all say ccr or chef client run or i'm going to run the chef client. We felt run matched this as it's what the users intent was. "I want to 'run' chef on that machine[s] with this resource/recipe/cookbook." Internally we've been talking about enabling localhost as a target since train supports that and if we do I could see a world where when a human wants to run chef the type `chef-run` or `chef run` and when running on servers chef-client is a service. I guess this is to say I am for leaving it as chef-run unless we come up with a better name since our user testing early on showed chef target was confusing.

coderanger:
I think that's a compelling argument, but I also think the pre-existing usage of "Chef run" as a term is too strong to overload any further. How about `chef cast` as something in the same general vicinity?

jakauppila:
Just to throw a few more suggestions out there:

`chef invoke` (this is my favorite)

`chef conduct`

`chef administer`

`chef perform`

`chef direct`

`chef control`

`chef work`

`chef act`

`chef use`

new/chef-workstation.md
```diff
@@ -0,0 +1,108 @@
...
+3. The installer currently called "ChefDK" will be renamed "Chef Workstation".
+   This will require setting up documentation to explain the rename, as well as
+   redirects and supporting pages in the download system to ensure users don't
+   get lost.
+4. The `chef-dk` gem will not be renamed at this time, as few users see the gem
+   name there and we're already changing a lot of things. This may be revisited
+   in the future when the dust settles.
+5. The installer for `chef-dk` (now named "Chef Workstation") will be reconfigured
+   to install into `/opt/chef-workstation` and use `~/.chef-workstation` as the
+   primary configuration folder. A symlink for `/opt/chef-dk` (and related windows-y paths)
```
thommay: `/opt/chefdk`

thommay:
Leaving aside the naming issues, I think there are some things we can do to make the developer experience better.
Currently, the `chef-dk` repository conflates several things:
 * The `chef` cli 
 * The packaging of the tools we think make up the workflow
 * The policyfiles code
 * The `chef generate` code

This RFC already touches on the first two things, but not the latter two. I'd like to propose moving (at least) the policyfiles code in to a separate repository (in line with `cookbook-omnifetch`), which would also (I think) eventually facilitate converging berks and policies into a single code base.

thommay:
I also think we should start the discussion that Noah alludes to in
> There should be one-- and preferably only one --obvious way to do it.
as a separate RFC, urgently.

coderanger:
Summary of discussion from community meeting:

1. People suggested and seemed on board with changing the primary CLI entrypoint from `chef-run` to `chef apply`.
2. To make (conceptual) room for this, we will deprecate `chef-apply` as a CLI and migrate the debugging-centric functionality to `chef-shell -a recipe.rb`
3. We'll work towards a local execution mode in `chef apply` (nee `chef-run`) as an eventual goal.
4. People seemed broadly in favor of having just one installer package and naming it Chef Workstation.

new/chef-workstation.md
robbkidd: I agree with this sharp distinction. Here's my understanding of the "just the installer" option for what Workstation/DK becomes:

* the "product"/project under which end-user tooling is available
* the release artifact (installer) for that tooling
* directly responsible for the source code of "entry point" utilities (e.g. a tray app or CLI command suite root like `chef`)

Actual functionality would be implemented in component tools that the `chef` root command dispatches to, as the run/target/ad-hoc capability is described in this RFC. Some of our current muddy waters come from that the `chef` command—and all of its sub-command logic for policy files and such—is currently implemented in the `chef-dk` gem.

coderanger:
Summary for this week:

1. No major objections to the summary from last week.
2. General agreement to do the rename sooner rather than later.
3. Floated the idea of syncing the versions of DK and Workstation, so the first post-rename release would be "Chef Workstation 3.1" or "Chef Workstation 4.0" (depending on final opinions).
4. Undoing the mega-gemfile-lock is a route some people want to explore, but won't be part of the initial rename swapover.
5. For the 3.1 release, we'll migrate the mega-gemfile to the chef-workstation repo and do a final sync of the chef-dk/omnibus/ content such that going forward, the chef-workstation repo "owns" packaging for all of this.

new/chef-workstation.md
```diff
@@ -0,0 +1,131 @@
...
+    so that I can share knowledge and answer questions.
+
+## Specification
+
+1. The current `chef-apply` command will be deprecated. The use cases around
+   debugging single recipes will be folded into the `chef-shell` tool as `chef-shell --apply`.
+   The simplified local execution use cases for `chef-apply` will be folded into
+   `chef-run`.
+2. The `chef-run` command will be left as-is for now. At some point after Chef 15
+   is released, it will likely be renamed to `chef-apply` once the soon-to-be-deprecated
```
jaymzh: you mean `chef apply`. The goal is to house it as a chef sub command, afaik

jaymzh: Oh, nevermind you cover this below.

new/chef-workstation.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Downsides
+
+This does introduce some potential confusion around the "chef apply"/`chef-apply`
+conceptual space. The existing `chef-apply` command line tool has been useful
+for the very early stages of teaching Chef and for some types of debugging, but
+after those early days, it's mostly ignored. As such, the authors of the RFC feel
+this is an acceptable trade off to make.
+
+## Alternatives
```
robbkidd: For the purposes of the eventual merge of this RFC, should this section be renamed to "Alternatives Considered, but not Selected"? Or the contents moved to the pull request as a record of what was considered? I think that would make this RFC more declarative about what is and will be.

PR #311, opened 2018-07-05
Chef Workstation Tray Application

mchiang0610:
The goal of the Chef Workstation tray application is to provide a greater user experience on the desktop for Chef users.

Initially, we want to start with automatic updates as the first feature to address a problem we’ve heard for a long time from our community members and customers - the process of checking and updating versions of ChefDK and Chef Workstation isn’t as a smooth of an experience as the team wanted. Often users are left in the dark when new features or bug fixes that addressed a particular concern are made available. With automatic updates, the tray application will notify the user about new versions with a clear changelog, and facilitate the update if accepted by the user.

This will set a great foundation for the team to focus on features that will provide improved cohesion of the current Chef tooling and what’s to come.


jaymzh:
I'd like to see two things:
1. More words about how the update mechanism will work (ala using native packages for the distro, etc.)
2. A mechanism for administratively turning off these notifications.

mchiang0610:
@jaymzh 
For this RFC, I originally wanted to keep it high-level to get agreement that this is something that we want to pursue, and then layout the discussion for implementation as we proceed - subject to our collective knowledge. 

I want to unravel your questions a little bit: 
1. Is there a specific update mechanism you would like to see? Also, are there any hard requirements that would prevent usage? 

2. Do you/your team currently administratively install ChefDK?

coderanger:
@mchiang0610 I think you'll find a lot of us would like to see at least a basic proposal for the implementation before we move forward, given that it's not going to be very straightforward and there's a lot of details that matter as to how disruptive this will be :) Not that any of us are against the idea, but we need to walk through a lot of the possible scenarios to make sure we don't cause problems in the name of "ship early and ship often" as happened with chef-run.

coderanger:
That is not the correct course, the goal of an RFC is it includes both design and implementation discussion in one document so that we can discuss it as a whole. I'm going to close out the other PR so you can move the content into this document.

majormoses:
One thing that I have personally always found harder than it should be is to find which versions of which gems are included in `chefdk` before updating. After watching the video I feel like the About screen (and possibly in the update screen as a diff) should contain a list of gems (or to be) installed. While this might seem trivial at a glance updating can pull in dependencies such as `rubocop`, `cookstyle`, or `foodcritic` that force you to have to update your code to comply with new/different rule sets.

new/tray-app.md
```diff
@@ -0,0 +1,106 @@
...
+
+At this time, updates will be installed via the operating system’s default package manager.  
+
+#### Disabling Update Checks
+
+Users and administrators will have multiple methods to disable updates if necessary. 
+
+From the tray application preferences
+From the Chef Workstation command line
+From the Chef Workstation configuration file
```
coderanger: How do you propose to actually implement this? On macOS it might be harder than you think :-/

new/tray-app.md
```diff
@@ -0,0 +1,106 @@
...
+
+The system will not notify the operator again for the same version of Chef Workstation after the operator declines to install an available upgrade, though the option to install will remain available. 
+
+Subsequent releases of Chef Workstation to the configured update channel will trigger a new prompt. 
+
+#### How Updates are Installed
+
+At this time, updates will be installed via the operating system’s default package manager.  
+
+#### Disabling Update Checks
```
coderanger: This section needs to be fleshed out a bunch.

new/tray-app.md
```diff
@@ -0,0 +1,106 @@
...
+
+##### Repeat Prompts
+
+The system will not notify the operator again for the same version of Chef Workstation after the operator declines to install an available upgrade, though the option to install will remain available. 
+
+Subsequent releases of Chef Workstation to the configured update channel will trigger a new prompt. 
+
+#### How Updates are Installed
+
+At this time, updates will be installed via the operating system’s default package manager.
```
coderanger: How does this handle updates to the tray icon tool itself? Also Windows and macOS do not have a default package manager so this needs to be massively expanded.

new/tray-app.md
```diff
@@ -0,0 +1,106 @@
...
+
+From the tray application preferences
+From the Chef Workstation command line
+From the Chef Workstation configuration file
+
+The user will still be able to manually check for updates. 
+
+## Alternatives
+
+Users using package managers,such as homebrew, to check for and perform updates. We will not limit users who want to continue using their package managers.
```
coderanger: How will the two interact?

new/tray-app.md
```diff
@@ -0,0 +1,106 @@
...
+
+#### How the Need for Update is Determined
+
+The updater will be configured to one of Chef’s product channels. At this time that is the ‘current’ channel and the ‘stable’ channel. The updater will periodically make calls to the omnitruck.chef.io service to check for a newer version within the configured channel.
+
+#### When Updates Prompt
+
+The user will be prompted for all updates if update checks are not disabled
+ 
+For new major releases the user can choose to skip the major release and instead update from the stable channel of their current release.
```
coderanger: I don't understand this sentence.

new/tray-app.md
tas50: Either or

coderanger: That's not really an option if we're saying this will be part of Chef Workstation and start automatically.

new/tray-app.md
tas50: Package manager == packaging system.

new/tray-app.md
tas50: I can already homebrew packages that auto update themselves. You just have to chose which one you want to rely on.

new/tray-app.md
tas50: If you're on Worksation 1.x and 2.x comes out. You can skip 2.X and continue to get release prompts for say 1.10 or 1.11 since we support N-1.

new/tray-app.md
coderanger: Retracting this point as we agreed to not chase Mac App Store, so it goes back to being very easy.

coderanger:
To capture from the meeting:

```
robbkidd [9:43 AM]
Rev:
+ there will be per-user config and a single, administrator-owned host-wide config
+ User and Host configs will merge with Host winning
+ the user will have feedback on config settings that have been set by Host config
```

coderanger:
I would like to call again for getting more implementation details in this. I realize not everyone shares my views on the RFC process, but I don't think a design doc should be enough. Given this is not something we have ever done before, I would like to see a basic prototype and some implementation details based on that. They can't be Forever Binding™ because that's not how software works, but I would like to at least have some specifics to ground the discussion.

coderanger:
(I think we've gotten past the point of the community being at least in-general on board with the idea of a tray icon service)

tyler-ball:
@majormoses I think thats a great feature I would love to see, but not in the first pass of the tray app. I would see that being a great feature request against the project once it is out though.

new/tray-app.md
```diff
@@ -0,0 +1,108 @@
...
+### Update Checking Specification
+
+#### Update Checks
+The update checks will occur in a daemon that runs in the background. By default this will be configured to start on user login, and may be started by chef-workstation tooling if it’s not running when chef-workstation commands are executed.
+
+This daemon may provide additional functionality beyond update checking in the future. The prompt for installation will occur from the Tray Application that Chef Workstation ships with.
+
+#### When Update Checks Run 
+
+Unless disabled, update checks will run automatically in the background of any workstation with Chef Workstation installed, no more frequently than once per (hour? day? week?)
```
tyler-ball: I like hourly 😀

PR #313, opened 2018-07-12
Chef Workstation Update Notifications

mchiang0610:
The goal of this RFC is to discuss how the update notification process will work inside Chef Workstation through its included tray application.

This is for us to work through the implementation before we move forward.


coderanger:
Please move this content into your existing RFC PR #311.

PR #314, opened 2018-07-12
Refine RFC 045.

markan:
This is a proposed set of refinements to the rfc 045 node state separation API. 

While avoiding particular implementation details, this has been designed to be both reasonable to implement, and maximize backwards compatibility.


lamont-granquist:
:+1:

PR #315, opened 2018-07-16
New RFC to lay out a process for working on guide content in the chef/chef repo

coderanger:
This is covered in the RFC, but the tl;dr:

It's been really hard to keep the docs in sync with Chef as features get added/changed. There's already long-term improvements in the works for the documentation system, but as a short-term mitigation we're going to add a `guides-drafts/` folder to the `chef/chef` repo so we can start writing some long-form content in the Chef repo itself. This will allow updating both the code and guide related to the code in a single PR (which we would like to eventually require as policy).

These draft guides will not be advertised as ready for user consumption, this is about making sure there is a place in `chef/chef` to work on writing as we go.

### DIFF
--- a/new/draft-guides.md
+++ b/new/draft-guides.md
@@ -1,0 +1,56 @@
+---
+RFC: unassigned
+Title: In-Repo Draft Guides for Chef
+Author:
+- Noah Kantrowitz <noah@coderanger.net>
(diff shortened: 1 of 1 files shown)


tas50:
Approved in our meeting.

PR #318, opened 2018-07-26
Update RFC 100: Resource Lifecycle

tas50:
Further explanation of what an appropriate resource is to port to core.

### DIFF
--- a/rfc100-resource-lifecycle.md
+++ b/rfc100-resource-lifecycle.md
@@ -16,15 +16,21 @@ This RFC describes the support lifecycle of [core resources](https://docs.chef.i
 As a Chef user,
 I want core resources to be usable, reasonable and supported,
 so that using Chef is a quality experience.
+```
 
+```
 As a Chef user,
 I want Chef to ship with resources out of the box,
 so that I can easily automate my systems without additional dependencies.
+```
 
+```
 As a Chef user,
 I want to understand the current state of documented resources,
 so that I implement resources correctly.
+```
 
+```
 As a Chef developer,
 I want a process for adding new resources to Chef,
 so that I improve the usability of Chef.
@@ -34,7 +40,7 @@ so that I improve the usability of Chef.
 
 ### Resource Scope
 
-Core resources should exist to automate all components of the underlying operating system and common subsystem services. This includes components such as authentication, raid, disk partitions, firewalls, or remote management transports. This does not include specific applications that users may want to automate such as database or webservers, as these are better suited for resources within cookbooks that can be rapidly iterated upon new software releases.
+Ideally core resources should exist to automate all components of the underlying operating system and common subsystem services. This includes components such as authentication, raid, disk partitions, firewalls, containers, or virtualization systems. This does not include specific applications that users may want to automate such as database, application servers, or webservers. Application resources are better suited for resources within cookbooks that can be rapidly iterated upon new application releases.
 
 ### Design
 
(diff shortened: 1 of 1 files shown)


majormoses (approved):

rfc100-resource-lifecycle.md
```diff
@@ -12,28 +12,47 @@ This RFC describes the support lifecycle of [core resources](https://docs.chef.i
...
+
+Ideally core resources should exist to automate all components of the underlying operating system and common subsystem services. This includes components such as authentication, raid, disk partitions, firewalls, containers, or virtualization systems. This does not include specific applications that users may want to automate such as database, application servers, or webservers. Application resources are better suited for resources within cookbooks that can be rapidly iterated upon new application releases.
+
+### Design
+
+Core resources should be well-designed and self contained. Resources should work as end-users would expect them to work and care should be taken to reduce surprises.
+
+### Maturity
+
+Users should have an expectation of stability within core resources and this requires resources to have be proven out before being shipped in Chef. In order to accomplish this resources should first be shipped in cookbooks where the resource design can be proven and refined and support added for all currently supported platforms](<https://github.com/chef/chef-rfc/blob/master/rfc021-platform-support-policy.md>).
```
coderanger: You've got some wonky Markdown formatting in this link.

rfc100-resource-lifecycle.md
```diff
@@ -12,48 +12,69 @@ This RFC describes the support lifecycle of [core resources](https://docs.chef.i
...
+I want Chef to ship with resources out of the box,
+so that I can easily automate my systems without additional dependencies.
+```
 
-    As a Chef developer,
-    I want to develop useful resources and reduce complexity by eliminating unhelpful resources,
-    so that I improve the usability of Chef.
+```
+As a Chef user,
+I want to understand the current state of documented resources,
```
kagarmoe: We need a section declaring the documentation & docs maintenance for resources.

coderanger:
We should add a "requirements for merge" section which should include at least:

1. Description for the resource
2. Description fields for all properties
3. Unit tests (once chefspec-in-core branch is merged this should be easy)
4. Integration tests for relevant platforms

tas50:
Pinging @coderanger @majormoses @happycoil @stuartpreston @tomdoherty @jasonwbarnett @martinisoft @nathwill @NaomiReeves @eoldegrove @gsreynolds who have touched resources in the last year.

martinisoft:
Overall very comprehensive and I like what is going on here with the outlined processes. This (to my knowledge) has more or less been an unofficial policy for getting stuff into core (speaking from the FreeBSD side of things) so I am glad this is formal. Also much better wording and clarification in general to define everything. 👍 from me

tomdoherty:
Reads well. I like it

Happycoil:
Should this outline how a community member should pick up a proposed core resource and start work? So we can avoid multiple people starting work on the same resource.

This is what https://github.com/chef/chef/projects/15 is for, right?

tas50:
@Happycoil That's probably a good idea. Any ideas for how we could do that?

Happycoil:
@tas50 Assigning the issue maybe? That could be done by a community engineering member when it's announced in #chef-dev.

stuartpreston:
@Happycoil Yes, self-assigning the issue when it's announced would be a good start.  Sometimes a contributor will get pulled onto other tasks and we understand that, so again #chef-dev can be used to let people know when you're not working on an item any more.

rfc100-resource-lifecycle.md
```diff
@@ -63,56 +88,54 @@ Resources in retired state have been in deprecated state and emitting warnings a
...
-* Add deprecated cookbook warnings for conflicting cookbooks.
+- Identify resources for adoption.
+- Create an issue at [https://github.com/chef/chef] outlining the reason for shipping the resource in core.
+- Announce your proposal in the #chef-dev channel on Chef Community Slack.
+- Chef-client maintainers will work with you to fully document the needs and the proposed design of the resource.
+- Once two chef-client maintainers have approved the proposal the `Status: Approved for Development` label will be applied and development may begin.
+
+#### Development Process
+
+- All development must take place initially in a cookbook.
```
jaymzh: I strongly object to this. While this is **often** the best case, it's not **always** the case. Adding a new `package` resource, say, **dnf_package** should **not** have happened in a cookbook first, it's nonsensicle. I'm fine ENCOURAGING this, but it absolutely is NOT a "must" - I fought against it being in the original RFC, and i will fight against it being in this one too.

jasonwbarnett: @jaymzh I think I appreciate where you're coming from, but I prefer the idea of having dev take place in a cookbook first for the fact that you aren't locked into the chef-client release cycles which are slow (for a good reason). This especially bites you when you require changing behavior in a non-backwards compatible way because of the way behavior MUST be slowly changed once introduced in the client core.

jaymzh: I disagree. Adding DNF support to `package` should *not* happen in a cookbook. I upgrade my OS, and now you tell me I have to get a cookbook for Chef to work right? No, that's absurd, I should be able to grab the latest chef and move on. There are a lot of resources that would be nice and core and it's unclear if they'll make it into core so they need incubation inside of a cookbook, but there are many that *must* be in core. Supporting the new package format that Fedora moves to, or updating `service` to support systemd, for example, are non-optional. If we didn't support systemd for service, Chef would have ceased to be able to be used on any modern Linux distro. Putting that in a cookbook doesn't make sense, it should be developed and on and release with core Chef because these are core parts of Chef.

It's one thing to say the default path is through cookbooks, but it's a **whole separate** thing to codify a rule that no Chef developer - employee or otherwise - can submit a PR to add systemd support to the `service` provide without publishing a cookbook first. That's not reasonable.

PR #320, opened 2018-08-08
Update RFC-21: Chef Platform Support Policy

tas50:
This document was really important at the time and allowed us to write some nice words on the docs site, but now the docs site is actually better setup and we should update this RFC with some of those words.

- Remove the list of specific versions that we weren't keeping up to date
- Clarify that Ohai needs to work on any new platform
- Clarify the types of support (commercial vs. community)
- Document how we align with vendor support cycles in deciding what platform versions we support

### DIFF
--- a/rfc021-platform-support-policy.md
+++ b/rfc021-platform-support-policy.md
@@ -12,6 +12,13 @@ The purpose of this RFC is to clarify which platforms the Chef community chooses
 
 The [chef docs site](https://docs.chef.io/) has [details of the platforms Chef Software, Inc validates for release](https://docs.chef.io/platforms.html).
 
+# Types of Support
+
+This RFC outlines the process for community support of a platforms on Chef, but it's worth noting the distinction between community supported and Chef Software Inc. supported:
+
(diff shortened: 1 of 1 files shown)


stuartpreston:
This looks great, from the Microsoft perspective I think we're covered.  Do we need words to make sure we disambiguate this RFD (which mostly seems to represent Chef Client) from other "Chef" projects such as Chef DK and Chef Workstation?

rfc021-platform-support-policy.md
```diff
@@ -12,13 +12,20 @@ The purpose of this RFC is to clarify which platforms the Chef community chooses
...
+
 # Adding a new platform
 
 Community maintenance of a platform requires that core chef resources,
 including `package`, `file`, and `service`, MUST have working providers
 for the platform. For example, on Debian the package provider supports both
-`dpkg` and `apt`. Ohai SHOULD also be able to gather relevant
-information on the platform.
+`dpkg` and `apt`. Ohai MUST also be able to gather relevant information on
+the platform such as platform information, cpu type/count, and available memory.
```
jaymzh: :+1:

tas50:
Feedback:

- Clarify that this is for community support
- The definitions muddy the picture of oss/commercial

PR #321, opened 2018-08-30
A less bikeshed-y version of RFC021 (hopefully)

coderanger:
RFC021 makes me sad. I don't see any way to rewrite that document to make it actually actionable in some way. There is no formal definition of "community support" that is useful, it's always just best effort. Maybe this will be more actionable to end users if the goal is to answer "Does Chef work on X?" or "Where can I use Chef?".

### DIFF
--- a/new/platform-functionality.md
+++ b/new/platform-functionality.md
@@ -1,0 +1,49 @@
+---
+RFC: unassigned
+Title: Platform Functionality Index
+Author: Noah Kantrowitz <noah@coderanger.net>
+Status: Draft
+Type: Informational
+Replaces: RFC021
+---
+
+# Platform Functionality Index
+
+This document records the platforms that are known to "work", for various
+definitions of that word. This document is correct as of 2018-08-29, though it
+is possible that future development has altered the list.
+
+It does not define what is "supported" or "maintained". For what platforms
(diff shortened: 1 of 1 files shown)


tyler-ball:
Rather than saying "Chef works anywhere the embedded Ruby runs" could we instead track a list of places it _doesn't_ work? As it is now this seems like documentation of something that is implied

lamont-granquist:
I think the point of RFC021 should really be around answering the questions that we get on slack like "Is Ubuntu 18.10 supported?" (right after it comes out).

We have one answer to that which is defined around a narrow set of Enterprise customers that pay for "support" to mean that we have CI cranking on every commit and tested on that O/S.

I would like to have a more common sense approach to the definition of "support" which hits what most users care about which is "you can use the binary from the prior O/S rev just fine, that is a design goal of omnibus so it should just work, there are no known bugs, feel free to use it, if you report bugs we will get them fixed".  And what they would want to know about is if there were any huge known bugs (e.g. when SIP was introduced in Mac and we had to move everything out of /usr/bin bEcAuSe SeCuRiTy).

stuartpreston:
This RFC appears to be a reflection of the current state rather than a proposal for a major change to Chef.  I think we have clear "what do Chef Software (commercial)" support and "what is supported by the community" statements in https://docs.chef.io/platforms.html, so is this RFC really necessary?

PR #322, opened 2018-09-06
Close the CBGB

nathenharvey:
The CBGB has not officially met since 2016 so is all but closed.  This makes it official.


tas50:
5 - 0 vote in slack to approve.

PR #324, opened 2018-10-04
YATM: Target mode proposal

btm:
A proposal for a target mode at a lower level in Chef.

### DIFF
--- a/new/target-mode.md
+++ b/new/target-mode.md
@@ -1,0 +1,76 @@
+---
+RFC: unassigned
+Title: Chef Target Mode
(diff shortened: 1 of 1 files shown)


jaymzh:
:+1:

btm:
fixme: credential storage

lamont-granquist:
From slack:  `Chef::Provider.target_mode?` should probably get pushed into the node_map as a boolean that is false by default and changed to true for resources+providers that support target_mode.

`provides :execute, target_mode: true`

`provides :cisco_stuff, platform: "cisco", target_mode: true`

PR #325, opened 2019-01-28
Update rfc020-community-guidelines.md

jonlives:
I don't use the IRC channels anymore, so don't think it makes sense for me to be an advocate in them.


btm:
IRC was removed entirely in https://github.com/chef/chef-rfc/pull/326.

PR #326, opened 2019-01-28
Remove IRC advocates and add Slack

tas50:
As a company we're not maintaining IRC anymore. Swap IRC for Slack and also add myself for GitHub.


tas50:
We should probably kill off one of those

robbkidd:
I didn't kill either off, but added a commit to update the `ADVOCATES.{toml|md}` for comparison.

robbkidd:
Removed the list of advocates from RFC020 in favor of a link to ADVOCATES.md.

robbkidd:
@tas50 I think this is ready for a final review.

PR #329, opened 2019-02-07
RFC000: Add a removed state and tas50 as an editor

tas50:
We're using the removed state already for things we accepted and then later removed. We should add this formally. I'm open to a better name if anyone has one.

I also replaced leff with myself as an editor. He signed off on the change.


tas50:
Approved in the 2/14 meeting

PR #333, opened 2019-02-27
RFC097 update: Remote sources in include_policy & policy_revision_id validation

mattray:
This patch adds support for "remote" sources for include_policy and adding policy_revision_id as a validation field for "path" and "remote" sources.


danielsdeleo:
It would be worth discussing the pros/cons of having users put a policy archive tarball on their server instead of just the JSON file. Then `path` based cookbooks from the upstream policy would work just fine since they would be bundled in. A possible downside of this is that it could be confusing to users in that it would behave a little differently than the git one. One way to work around that would be to make a new DSL method name so that the behavior difference is reflected in the DSL language.

mattray:
@danielsdeleo like new syntax for `include_artifact`? That sounds like a larger change, probably a separate RFC. For now I just need `remote` include policies.

tas50:
Confirmed in Slack with the caveat that @mattray looks into @danielsdeleo's more involved fix.

